PR #21, opened 2021-11-16
Add hostname e2e tests

jinghanm:
## Description

Add hostname E2E tests.

## Checklist


src/tests/e2etest/HostNameTests.cs
```diff
@@ -0,0 +1,115 @@
...
+        }
+
+        [Test]
+        public void HostNameTest_Set_Get()
+        {
+            int extendedTimeoutSeconds = 70;
+            SetTwinTimeoutSeconds(extendedTimeoutSeconds);
+            var desiredHostName = new DesiredHostName
+            {
+                DesiredName = "hostmachine",
```
MariusNi, line 66: >hostmachine [](http://example.com/codeflow?start=31&length=11)

"TestHost" or"TEST-HOST"  or similar?

MariusNi (approved):
:shipit:

src/tests/e2etest/HostNameTests.cs
```diff
@@ -0,0 +1,115 @@
...
+
+            Assert.True(IsRegexMatch(expectedHostNamePattern.NamePattern, deserializedReportedObject.Name));
+            Assert.True(IsRegexMatch(expectedHostNamePattern.HostsPattern, deserializedReportedObject.Hosts));
+        }
+
+        [Test]
+        public void HostNameTest_Set_Get()
+        {
+            int extendedTimeoutSeconds = 70;
+            SetTwinTimeoutSeconds(extendedTimeoutSeconds);
```
AhmedBM, line 63: This will affect all tests and will override the default value that is specified as a parameter in the pipeline. The timeout should only be changed for this test.
I would recommend looking at [SetUp](https://docs.nunit.org/articles/nunit/writing-tests/attributes/setup.html)/TearDown so we can set/reset the value back to the original default for the rest of the tests. You will have to create your own _TestFixture_ to do this.

AhmedBM (changes requested):

src/tests/e2etest/HostNameTests.cs
AhmedBM: You might want to consider not hardcoding the timeout but maybe using a multiple of the default value that was originally specified if the intent is to have the test timeout longer. But why exactly are we bumping the timeout for this test? Does setting the hostname a long-running operation?

src/tests/e2etest/HostNameTests.cs
```diff
@@ -0,0 +1,154 @@
...
+namespace e2etesting
+{
+    [TestFixture]
+    public class HostNameTests : E2eTest
+    {   
+        string ComponentName = "HostName";
+        public partial class ExpectedDesiredProperty
+        {
+            public string value { get; set; }
+            public int ac { get; set; }
```
AhmedBM, line 22: why not `JsonIgnore` this property? This would make the `ExpectedDesiredProperty` much easier/cleaner to read when defining the property.

src/tests/e2etest/HostNameTests.cs
```diff
@@ -0,0 +1,154 @@
...
+            public string ad { get; set; }
```
AhmedBM, line 23: same as above

src/tests/e2etest/HostNameTests.cs
```diff
@@ -0,0 +1,154 @@
...
+            {
+                DesiredName = "TestHost",
+                DesiredHosts = "127.0.0.1 localhost;127.0.1.1 TestHost;::1 ip6-localhost ip6-loopback;fe00::0 ip6-localnet;ff00::0 ip6-mcastprefix;ff02::1 ip6-allnodes;ff02::2 ip6-allrouters"
+            };
+
+             var expectedDesiredName = new ExpectedDesiredProperty
+            {
+                value = desiredHostName.DesiredName,
+                ac = 200,
+                ad = "-"
```
AhmedBM, line 95: unnecessary, see above comments on `JSONIgnore`

src/tests/e2etest/HostNameTests.cs
```diff
@@ -0,0 +1,154 @@
...
+            };
+
+            var expectedDesiredHosts = new ExpectedDesiredProperty
+            {
+                value = desiredHostName.DesiredHosts,
+                ac = 200,
+                ad = "-"
```
AhmedBM, line 102: unnecessary, see above comments on `JSONIgnore`

AhmedBM (commented):
Looks good, but please run the `E2E Cloud Tests` pipeline to make sure this works properly on your branch.

src/tests/e2etest/HostNameTests.cs
AhmedBM: ok, i can see the value of checking for equality on `ac`, but i don't think it should be a property exposed for the test to define.
I would suggest adding/checking these properties in the `E2eTest::AreEqualByJson` instead of hard coding them here. Its confusing to read for a module developer who is using this as a reference.

src/tests/e2etest/HostNameTests.cs
```diff
@@ -0,0 +1,153 @@
...
+{
+    [TestFixture]
+    public class HostNameTests : E2eTest
+    {   
+        string ComponentName = "HostName";
+        const int reposonseCodeSuccess = 200;
+        public partial class ExpectedDesiredProperty
+        {
+            public string value { get; set; }
+            public int ac { internal get; set; }
```
AhmedBM, line 23: Can we not just not include these properties and simply have them validated in the `E2eTest::AreEqualByJson`? This would make the desired property look just like the PnP object without these added ac/ad/av properties?

src/tests/e2etest/HostNameTests.cs
```diff
@@ -0,0 +1,136 @@
...
+
+            var expectedHostName = new HostName
+            {
+                Name = desiredHostName.DesiredName,
+                Hosts = desiredHostName.DesiredHosts
+            };
+
+            if ((GetNewTwin().ConnectionState == DeviceConnectionState.Disconnected) || (GetTwin().Status == DeviceStatus.Disabled))
+            {
+                Assert.Fail("Module is disconnected or is disabled- HostNameTest_Set_Get");
```
AhmedBM, line 92: No need to log test name in Assert>fail. That is already captured in the stack trace info on failure.

src/tests/e2etest/HostNameTests.cs
```diff
@@ -0,0 +1,136 @@
...
+            };
+
+            if ((GetTwin().ConnectionState == DeviceConnectionState.Disconnected) || (GetTwin().Status == DeviceStatus.Disabled))
+            {
+                Assert.Fail("Module is disconnected or is disabled - HostNameTest_Get");
```
AhmedBM, line 66: No need to log `HostNameTest_Get` in Assert.Fail - test name is already captured by Assertion

src/tests/e2etest/HostNameTests.cs
jinghanm: I have removed ac/ad/av properties from the HostName class

src/tests/e2etest/HostNameTests.cs
```diff
@@ -0,0 +1,136 @@
...
+            if ((GetNewTwin().ConnectionState == DeviceConnectionState.Disconnected) || (GetTwin().Status == DeviceStatus.Disabled))
+            {
+                Assert.Fail("Module is disconnected or is disabled- HostNameTest_Set_Get");
+            }
+
+            Twin twinPatch = CreateTwinPatch(ComponentName, desiredHostName);
+            int count = 2;
+            while (count-- > 0)
+            {
+                if (count == 0)
```
AhmedBM, line 99: Is this for setting back the hostname back to its original value?

jinghanm: Yes.

AhmedBM:
Make sure this works on the E2E Cloud Test pipelines before merging

src/tests/e2etest/HostNameTests.cs
AhmedBM: It seems `HostNameTest_Set_Get` depends on the value obtained by `HostNameTest_Get`. Tests should be atomic and not depend on other tests. The original value should be captured by this function.

AhmedBM:
Looks great! I still don't see the test on the E2E pipeline though. Make sure you test before merging!

AhmedBM (approved):

PR #32, opened 2021-11-24
Add ztsi e2e tests

jinghanm:
## Checklist


src/tests/e2etest/ZtsiTests.cs
```diff
@@ -0,0 +1,112 @@
...
+
+using NUnit.Framework;
+using Microsoft.Azure.Devices.Shared;
+using Microsoft.Azure.Devices;
+using System.Text.Json;
+using System;
+using System.Threading.Tasks;
+using System.Text.RegularExpressions;
+
+namespace e2etesting
```
MariusNi, line 12: >e2etesting [](http://example.com/codeflow?start=10&length=10)

Please use for namespace same name convention. Camel casing, etc. In this case: "E2eTesting" for example

src/tests/e2etest/ZtsiTests.cs
```diff
@@ -0,0 +1,112 @@
...
+            Twin twinPatch = CreateTwinPatch(ComponentName, desiredZtsi);
+
+            if (UpdateTwinBlockUntilUpdate(twinPatch))
+            {
+                Ztsi reportedObject = JsonSerializer.Deserialize<Ztsi>(GetNewTwin().Properties.Reported[ComponentName].ToString());
+                // Wait until the reported properties are updated
+                DateTime startTime = DateTime.Now;
+                while(((reportedObject.Enabled != expectedZtsi.Enabled) || (reportedObject.ServiceUrl != expectedZtsi.ServiceUrl)) && ((DateTime.Now - startTime).TotalSeconds < twinTimeoutSeconds))
+                {
+                    Console.WriteLine("[ZtsiTests] waiting for twin to be updated...");
```
MariusNi, line 92: > for twin [](http://example.com/codeflow?start=58&length=9)

In all logs where mentioning "twin" like this alone, add "module" to it ("module twin") to match the lines that say "module is disconnected" - to link the two terms together. Thoughts?

src/tests/e2etest/ZtsiTests.cs
```diff
@@ -0,0 +1,112 @@
...
+            if ((GetNewTwin().ConnectionState == DeviceConnectionState.Disconnected) || (GetTwin().Status == DeviceStatus.Disabled))
+            {
+                Assert.Fail("Module is disconnected or is disabled");
+            }
+
+            Ztsi deserializedReportedObject = JsonSerializer.Deserialize<Ztsi>(GetTwin().Properties.Reported[ComponentName].ToString());
+
+            Twin twinPatch = CreateTwinPatch(ComponentName, desiredZtsi);
+
+            if (UpdateTwinBlockUntilUpdate(twinPatch))
```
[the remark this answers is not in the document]
robertschaedler3: For example, `Enabled = true` and `ServiceUrl = ""` is an invalid configuration. If the update to `Enabled` state is sent to OSConfig first it will fail and the second update to `ServiceUrl` will succeed, which explains why the `ServiceUrl` is updated but not `Enabled`.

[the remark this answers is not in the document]
robertschaedler3: Suggest to break this into 2 separate updates:

1. First to `DesiredServiceUrl` (wait for successful update)
2. Then to `DesiredEnabled`

This should ensure the correct ordering and guarantee valid configurations

src/tests/e2etest/ZtsiTests.cs
```diff
@@ -0,0 +1,138 @@
...
+        [TearDown]
+        public void TestTearDown()
+        {
+            base.twinTimeoutSeconds = m_twinTimeoutSeconds;
+        }
+
+        [Test]      
+        public void ZtsiTest_Get()
+        {
+            if ((GetTwin().ConnectionState == DeviceConnectionState.Disconnected) || (GetTwin().Status == DeviceStatus.Disabled))
```
AhmedBM, line 51: Recommend using `GetNewTwin` to ensure most up-to-date twin in the first argument of the conditional

src/tests/e2etest/ZtsiTests.cs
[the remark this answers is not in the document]
jinghanm: Thanks Robert!

src/tests/e2etest/ZtsiTests.cs
```diff
@@ -0,0 +1,138 @@
...
+            };
+            var expectedZtsi = new Ztsi
+            {
+                ServiceUrl = desiredZtsi.DesiredServiceUrl
+            };
+            Twin twinPatch = CreateTwinPatch(ComponentName, desiredZtsi);
+
+            if (UpdateTwinBlockUntilUpdate(twinPatch))
+            {
+                Ztsi reportedObject = JsonSerializer.Deserialize<Ztsi>(GetNewTwin().Properties.Reported[ComponentName].ToString());
```
AhmedBM, line 82: Given we are testing reported properties only here, i would recommend not using `GetNewTwin()` as it performs an iot hub call for every call. Everything should be available in the `GetTwin()`. It will make your test faster and less-costly in the end

AhmedBM (changes requested):
For tests where only reported properties are present i would just use `GetTwin` instead of `GetNewTwin` as the latter performs is slower/more costly

src/tests/e2etest/ZtsiTests.cs
jinghanm: Thanks for pointing this out. I have fixed this issue.

src/tests/e2etest/ZtsiTests.cs
```diff
@@ -0,0 +1,138 @@
...
+        {
+            public int ac { get; set; }
+        }
+        private int m_twinTimeoutSeconds;
+
+        [SetUp]
+        public void TestSetUp()
+        {
+            m_twinTimeoutSeconds = base.twinTimeoutSeconds;
+            base.twinTimeoutSeconds = base.twinTimeoutSeconds * 2;
```
AhmedBM, line 39: why does this test require a longer timeout?

AhmedBM (commented):
Looks good, just curious as to why this test needs a longer timeout? Are the ZTSI module operations asynchronous?

src/tests/e2etest/ZtsiTests.cs
jinghanm: In practice, the extended time out seems to be needed for ztsi tests

AhmedBM (dismissed):

MariusNi (dismissed):
:shipit:

MariusNi (dismissed):
:shipit:

robertschaedler3 (approved):
👍🏻

AhmedBM (approved):

MariusNi (approved):
:shipit:

robertschaedler3 (dismissed):

MariusNi (changes requested):
:clock1: This cannot be merged until all test validation passes. Will remove the clock then.

MariusNi (dismissed):
:shipit:

robertschaedler3 (dismissed):
👍🏻

src/tests/e2etest/E2ETest.cs
```diff
@@ -46,6 +46,14 @@ public void SetTwinTimeoutSeconds(int seconds)
             twinTimeoutSeconds = seconds;
         }
 
+        public void CheckModuleConnection()
```
AhmedBM, line 49: This method will break the test on assertion failure as such, i would recommend renaming the method to `AssertModuleConnected` or something similar.

src/tests/e2etest/ZtsiTests.cs
```diff
@@ -0,0 +1,226 @@
...
+            }
+            if (reportedObject.Enabled != expectedEnabled)
+            {
+                Assert.Fail("Timeout for updating reported Enabled in module twin");
+            }
+            var responseObject = JsonSerializer.Deserialize<ResponseCode>(GetTwin().Properties.Reported[ComponentName][desiredEnabled].ToString());
+            Assert.True(responseObject.ac == enabledResponseCode);
+        }
+
+        public void Set_Both(string serviceUrl, bool enabled)
```
AhmedBM, line 112: Can we make the method more descriptive? `Set_ServiceUrl_Enabled`?

src/tests/e2etest/ZtsiTests.cs
```diff
@@ -0,0 +1,226 @@
...
+                DesiredServiceUrl = serviceUrl
+            };
+            Twin twinPatch = CreateTwinPatch(ComponentName, desiredZtsi);
+            if (!UpdateTwinBlockUntilUpdate(twinPatch))
+            {
+                Assert.Fail("Timeout for updating module twin");
+            }
+        }
+
+        public void Get_Both(string expectedServiceUrl,  int serviceUrlResponseCode, int expectedEnabled, int enabledResponseCode)
```
AhmedBM, line 128: Can we make the method more descriptive? `Get_ServiceUrl_Enabled`?

AhmedBM (approved):
Looks good, just some minor comments on naming

AhmedBM (approved):

robertschaedler3 (approved):
👍🏻

MariusNi (approved):
:shipit:

PR #34, opened 2021-11-29
Jinmen/settings

jinghanm:
## Description
Add settings e2e test

## Checklist

### DIFF
--- a/src/tests/e2etest/E2ETest.cs
+++ b/src/tests/e2etest/E2ETest.cs
@@ -262,13 +262,6 @@ namespace e2etesting
             Assert.AreEqual(expectedJson, actualJson);
         }
 
-        public static bool IsSameByJson(object expected, object actual)
-        {
-            var expectedJson = JsonSerializer.Serialize(expected);
-            var actualJson = JsonSerializer.Serialize(actual);
-            return expectedJson == actualJson;
-        }
-
         public static bool IsRegexMatch(Regex expected, object actual)
         {
             string actualJson = JsonSerializer.Serialize(actual);
--- a/src/tests/e2etest/E2ETest.cs
+++ b/src/tests/e2etest/E2ETest.cs
@@ -262,6 +262,13 @@ namespace e2etesting
             Assert.AreEqual(expectedJson, actualJson);
         }
 
+        public static bool IsSameByJson(object expected, object actual)
+        {
+            var expectedJson = JsonSerializer.Serialize(expected);
+            var actualJson = JsonSerializer.Serialize(actual);
+            return expectedJson == actualJson;
+        }
+
         public static bool IsRegexMatch(Regex expected, object actual)
         {
             string actualJson = JsonSerializer.Serialize(actual);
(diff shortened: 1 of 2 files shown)


src/tests/e2etest/SettingsTests.cs
```diff
@@ -0,0 +1,123 @@
...
+
+            Twin twinPatch = CreateTwinPatch(ComponentName, desiredSettings);
+
+            var reportedObject = new Settings();
+            if (UpdateTwinBlockUntilUpdate(twinPatch))
+            {
+                reportedObject = GetSettingsReportObject();
+                // Wait until the reported properties are updated
+                DateTime startTime = DateTime.Now;
+                while(!IsSameByJson(desiredSettings, reportedObject) && (DateTime.Now - startTime).TotalSeconds < twinTimeoutSeconds)
```
AhmedBM, line 110: not understanding why we have to keep doing `IsSameByJson`? Once we set the setting, the next reported interval should contain the value. So all needed here is to wait, until the next update after the patch was applied, none of this json comparing is needed

AhmedBM (changes requested):

src/tests/e2etest/SettingsTests.cs
jinghanm: Thanks Ahmed. I have made all necessary changes as you requested.

src/tests/e2etest/E2ETest.cs
```diff
@@ -262,6 +262,13 @@ public static void AreEqualByJson(object expected, object actual)
             Assert.AreEqual(expectedJson, actualJson);
         }
 
+        public static bool IsSameByJson(object expected, object actual)
```
AhmedBM, line 265: Unnecessary reference, not seeing any usage.

src/tests/e2etest/SettingsTests.cs
```diff
@@ -0,0 +1,115 @@
...
+            {
+                Assert.Fail("Module is disconnected or is disabled");
+            }
+
+            Twin twinPatch = CreateTwinPatch(ComponentName, desiredSettings);
+
+            var reportedObject = new Settings();
+            if (UpdateTwinBlockUntilUpdate(twinPatch))
+            {
+                reportedObject = GetSettingsReportObject();
```
AhmedBM, line 107: but were not testing against any expected values here? How do we know this is the correct result?

AhmedBM (changes requested):
This test just ensures the update is successful, but doesn't validate the expected state, we should make sure to also validate our expected state

src/tests/e2etest/SettingsTests.cs
jinghanm: I used to have IsSameByJson(desiredSettings, reportedObject) in the while loop as testing against the expected values.

jinghanm: Add testing again the expected values.

src/tests/e2etest/E2ETest.cs
jinghanm: Thanks. Removed.

AhmedBM (approved):

PR #39, opened 2021-12-02
Add file locking to ztsi module agent configuration

robertschaedler3:
## Description

Add file locking to ZTSI agent configuration file (`/etc/ztsi/config.json`). Before read/write, the ZTSI module places an exclusive lock `LOCK_EX` on the configuration file. 

If a lock cannot be placed (i.e. the file is locked by another process) during a **write** operation the module will wait and retry to acquire the lock. If the lock is not acquired after the maximum retries, the write operation will fail

If a lock cannot be placed during a **read** operation and there is a cached configuration from a previous successful read/write, the cached value is returned. Otherwise, the defaults are returned.

[flock(2) - Linux man page](https://linux.die.net/man/2/flock)

## Checklist


src/modules/ztsi/src/lib/Ztsi.h
```diff
@@ -61,11 +61,18 @@ class Ztsi
 
 private:
     static bool IsValidConfig(const AgentConfig& config);
-    virtual int ConfigFileExists();
+    virtual std::FILE* LockFile(const char* mode);
+    virtual std::FILE* LockFile(const char* mode, unsigned int millis, int count);
```
MariusNi, line 65: >millis [](http://example.com/codeflow?start=63&length=6)

Use full words please for variable names (milliseconds in this case)

MariusNi: Also in Ztsi.cpp for several comments and variable names that all use "Millis"

src/modules/ztsi/src/lib/Ztsi.h
```diff
@@ -61,11 +61,18 @@ class Ztsi
...
+    virtual void UnlockFile(std::FILE* fp);
+    static bool FileExists(const std::string& filePath);
     virtual int ReadAgentConfig(AgentConfig& config);
     virtual int WriteAgentConfig(const AgentConfig& config);
+    virtual int CreateConfigFile(const AgentConfig& config);
+    virtual int ParseAgentConfig(const std::string& configJson, AgentConfig& config);
```
MariusNi, line 71: >ParseAgentConfig [](http://example.com/codeflow?start=16&length=16)

ParseAgentConfiguration? Here and elsewhere

src/modules/ztsi/src/lib/Ztsi.cpp
```diff
@@ -205,97 +202,180 @@ bool Ztsi::IsValidConfig(const Ztsi::AgentConfig& config)
     return isValid;
 }
 
-int Ztsi::ReadAgentConfig(Ztsi::AgentConfig& config)
+bool Ztsi::FileExists(const std::string& filePath)
```
MariusNi, line 205: Consider making this part of CommonUtils lib (this is identical from FileExists used in Settings, CommandRunner, etc). That change would replace all these separate instances with the common lib one.

We can do this as part of a separate PR later, your call.

MariusNi (approved):
:shipit:

src/modules/ztsi/src/lib/Ztsi.cpp
robertschaedler3: Thanks, I will do in a later PR and move `IsValidClientName()` at the same time.

src/modules/ztsi/src/lib/Ztsi.cpp
```diff
@@ -205,97 +202,180 @@ bool Ztsi::IsValidConfig(const Ztsi::AgentConfig& config)
...
 
-    rapidjson::IStreamWrapper isw(file);
-    rapidjson::Document document;
-    if (document.ParseStream(isw).HasParseError())
-    {
-        OsConfigLogError(ZtsiLog::Get(), "Failed to parse config file: %s", m_agentConfigFile.c_str());
-        status = EINVAL;
-        return status;
-    }
+std::FILE* Ztsi::LockFile(const char* mode)
```
MariusNi, line 211: >Ztsi::LockFile [](http://example.com/codeflow?start=11&length=14)

OpenAndLockFile? Since this function does more than just locks. Same for UnlockFile

robertschaedler3: Thanks, fixed 👍🏻

MariusNi (approved):
:shipit:

src/modules/ztsi/src/lib/Ztsi.cpp
MariusNi: Thanks!

src/modules/ztsi/src/lib/Ztsi.cpp
```diff
@@ -158,201 +160,329 @@ int Ztsi::SetEnabled(bool enabled)
...
-        status = EINVAL;
-        return status;
-    }
+std::FILE* Ztsi::OpenAndLockFile(const char* mode)
+{
+    int fd = -1;
+    std::FILE* fp = nullptr;
 
-    if (document.HasMember(g_configPropertyEnabled))
+    if (FileExists(m_agentConfigurationFile) && (nullptr != (fp = fopen(m_agentConfigurationFile.c_str(), mode))))
```
sprt, line 216: Isn't this subject to race conditions? What if the file gets deleted right after the `FileExists()` but before the `fopen()` call? Isn't the `FileExists()` call simply redundant anyway?

robertschaedler3: `FileExists()` is there to ensure we only open the file if it has been created. Now looking at it I see that `FileExists()` can be removed there since we only ever open the file using a mode that will not also create a file if it does not exist (`r` `r+`).

robertschaedler3: Correct, I can remove this `FileExists()` in `OpenAndLock()`

PR #41, opened 2021-12-07
Update return values for cached enabled state

robertschaedler3:
## Description

Update return values of `SetEnabled` to return success even when a value is cached and not immediately applied to the configuration file.

## Checklist

### DIFF
--- a/src/modules/ztsi/src/lib/Ztsi.cpp
+++ b/src/modules/ztsi/src/lib/Ztsi.cpp
@@ -400,7 +400,7 @@ int Ztsi::WriteAgentConfiguration(const Ztsi::AgentConfiguration& configuration)
         std::string configurationJson = BuildConfigurationJson(configuration);
 
         int rc = std::fwrite(configurationJson.c_str(), 1, configurationJson.length(), fp);
-        if ((MMI_OK > rc) || (EOF == rc))
+        if ((0 > rc) || (EOF == rc))
         {
             OsConfigLogError(ZtsiLog::Get(), "Failed to write to file %s", m_agentConfigurationFile.c_str());
             status = errno ? errno : EINVAL;
--- a/src/modules/ztsi/src/lib/Ztsi.cpp
+++ b/src/modules/ztsi/src/lib/Ztsi.cpp
@@ -145,7 +145,7 @@ int Ztsi::SetEnabled(bool enabled)
         if (enabled != configuration.enabled)
         {
             configuration.enabled = enabled;
-            status = WriteAgentConfiguration(configuration);
+            status = IsValidConfiguration(configuration) ? WriteAgentConfiguration(configuration) : 0;
         }
     }
     else if (ENOENT == status)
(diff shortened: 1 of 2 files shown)


src/modules/ztsi/tests/ZtsiTests.cpp
```diff
@@ -87,7 +87,7 @@ namespace OSConfig::Platform::Tests
     {
         // Enabled can only be set to true when no configuration file exists since serviceUrl is empty string by default
         // No file is created for invalid configurations
-        ASSERT_EQ(EINVAL, ZtsiTests::ztsi->SetEnabled(true));
+        ASSERT_EQ(0, ZtsiTests::ztsi->SetEnabled(true));
```
MariusNi, line 90: >0 [](http://example.com/codeflow?start=18&length=1)

MMI_OK, not 0

MariusNi: Same in the other instances

src/modules/ztsi/src/lib/Ztsi.cpp
```diff
@@ -145,15 +145,16 @@ int Ztsi::SetEnabled(bool enabled)
         if (enabled != configuration.enabled)
         {
             configuration.enabled = enabled;
-            status = WriteAgentConfiguration(configuration);
+            status = IsValidConfiguration(configuration) ? WriteAgentConfiguration(configuration) : 0;
```
MariusNi, line 148: >0 [](http://example.com/codeflow?start=100&length=1)

MMI_OK?

src/modules/ztsi/src/lib/Ztsi.cpp
```diff
@@ -383,86 +392,72 @@ int Ztsi::ParseAgentConfiguration(const std::string& configurationJson, Ztsi::Ag
...
-            }
-            else
-            {
-                ftruncate(fileno(fp), rc);
-                m_lastAvailableConfiguration = configuration;
-            }
-
-            CloseAndUnlockFile(fp);
+        int rc = std::fwrite(configurationJson.c_str(), 1, configurationJson.length(), fp);
+        if ((MMI_OK > rc) || (EOF == rc))
```
MariusNi, line 403: >MMI_OK > rc [](http://example.com/codeflow?start=13&length=11)

There can be negative errors as well

robertschaedler3:
> Does m_lastEnabledState store the last enabled value we try to set?

@jinghanm yes, `m_lastEnabledState` is updated with the latest value set in `SetEnabled()`.

jinghanm:
> > Does m_lastEnabledState store the last enabled value we try to set?
> 
> @jinghanm yes, `m_lastEnabledState` is updated with the latest value set in `SetEnabled()`.

What if this value is not set successfully, do we still store this value in `m_lastEnabledState`

MariusNi (approved):
:shipit:

robertschaedler3:
> > > Does m_lastEnabledState store the last enabled value we try to set?
> > 
> > 
> > @jinghanm yes, `m_lastEnabledState` is updated with the latest value set in `SetEnabled()`.
> 
> What if this value is not set successfully, do we still store this value in `m_lastEnabledState`

Since the value is a boolean there is no additional validation that needs to be done before caching. So the value is always cached and stored for subsequent calls to set the service URL. Where it will be applied.

jinghanm:
> > > > Does m_lastEnabledState store the last enabled value we try to set?
> > > 
> > > 
> > > @jinghanm yes, `m_lastEnabledState` is updated with the latest value set in `SetEnabled()`.
> > 
> > 
> > What if this value is not set successfully, do we still store this value in `m_lastEnabledState`
> 
> Since the value is a boolean there is no additional validation that needs to be done before caching. So the value is always cached for subsequent calls to set the service URL. Where it will be applied.

Thank you very much for the explanation

PR #42, opened 2021-12-08
Update Networking to capture global DnsServers

hachotze:
## Description

- Updated Networking.DnsServers functions to capture global DNS servers in reported data.
- Added a new enumeration 'NetworkManagementService' to enable Networking module objects to remember the correct command to run based on enabled system services.
- Fixed a bug in the DnsServers data collection flow caused by unexpected trailing whitespace in command output.

## Checklist


src/modules/networking/src/lib/Networking.h
```diff
@@ -47,6 +47,13 @@ struct NetworkingSettings
     std::string connected;
 };
 
+enum class NetworkManagementService
```
MariusNi, line 50: Instead of a scoped to class enumeration declared outside of the class scope, better to define such enum as a class member. Same for NetworkingSettingType below.

src/modules/networking/src/lib/Networking.cpp
```diff
@@ -50,6 +50,16 @@ const char* g_templateWithDots = R"""({"InterfaceTypes":"..","MacAddresses":".."
 const char* g_emptyValue = "";
 const char* g_twoDots = "..";
 
+std::regex g_ipv4Pattern("(([0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])\\.){3}([0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])");
+std::regex g_ipv6Pattern("(([0-9a-fA-F]{1,4}:){7,7}[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,7}:|([0-9a-fA-F]{1,4}:){1,6}"
+":[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,5}(:[0-9a-fA-F]{1,4}){1,2}|([0-9a-fA-F]{1,4}:){1,4}(:[0-9a-fA-F]{1,4}){1,3}"
+"|([0-9a-fA-F]{1,4}:){1,3}(:[0-9a-fA-F]{1,4}){1,4}|([0-9a-fA-F]{1,4}:){1,2}(:[0-9a-fA-F]{1,4}){1,5}|[0-9a-fA-F]{1,4}"
+":((:[0-9a-fA-F]{1,4}){1,6})|:((:[0-9a-fA-F]{1,4}){1,7}|:)|[fF][eE]80:(:[0-9a-fA-F]{0,4}){0,4}%[0-9a-zA-Z]{1,}|::([fF][eE]{4}(:0{1,4}){0,1}:){0,1}"
+"((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])\\.){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])|([0-9a-fA-F]{1,4}:){1,4}"
+":((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])\\.){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9]))");
```
MariusNi, line 59: Lines 55 - 59 can be moved right 4 spaces to respect overall code alignmnt, without corrupting the string constant.

src/modules/networking/src/lib/Networking.cpp
```diff
@@ -391,53 +402,84 @@ void NetworkingObjectBase::GenerateInterfaceTypesMap()
...
     {
-        interfaceTypesData = RunCommand(g_getInterfaceTypesNetworkctl);
-        std::stringstream interfaceTypesDataStream(interfaceTypesData);
-        while(std::getline(interfaceTypesDataStream, interfaceData))
+        if (this->m_networkManagementService == NetworkManagementService::Unknown)
+        {
+            this->m_networkManagementService = NetworkManagementService::SystemdNetworkd;
+        }
+
+        std::regex interfaceTypesDataPatternNetworkctl("^\\s*[0-9]+\\s+.*$");
```
MariusNi, line 426: >"^\\s*[0-9]+\\s+.*$" [](http://example.com/codeflow?start=55&length=20)

Make this a constant (same as any other such incline constant strings or characters or numbers)

src/modules/networking/src/lib/Networking.cpp
```diff
@@ -391,53 +402,84 @@ void NetworkingObjectBase::GenerateInterfaceTypesMap()
...
+    {
+        GetInterfaceTypesFromSystemdNetworkd();
+    }
+    if (this->m_networkManagementService == NetworkManagementService::Unknown)
+    {
+        if (IsFullLoggingEnabled())
+        {
+            OsConfigLogError(NetworkingLog::Get(), "Network interface management service not found");
         }
     }
```
MariusNi, line 484: Should lines 474 and 478 be 'else if' rather than 'if'?

MariusNi, line 484: > } [](http://example.com/codeflow?start=3&length=2)

Also add an 'else' here once changing to 'if else' above

MariusNi:
>             "DNS Domain: mshome.net\n";

Align the constant strings properly. 

Like:

std::string foo = 
    "..."
    "...";

---
Refers to: src/modules/networking/tests/NetworkingTests.cpp:136 in bc90545. [](commit_id = bc90545c476ce20926b6e4eb4ea107c35f91edfa, deletion_comment = False)

src/modules/networking/tests/NetworkingTests.cpp
```diff
@@ -218,7 +219,18 @@ namespace OSConfig::Platform::Tests
...
+        const char* payloadNetworkManagerEnabled =
+        "{\"InterfaceTypes\":\"docker0=bridge;eth0=ethernet\","
+        "\"MacAddresses\":\"docker0=0a:25:3g:6v:2f:89;eth0=00:15:5d:26:cf:89\","
+        "\"IpAddresses\":\"docker0=172.32.233.234,::1;eth0=172.27.181.213,10.1.1.2,fe80::5e42:4bf7:dddd:9b0f\","
+        "\"SubnetMasks\":\"docker0=/8,/128;eth0=/20,/16,/64\","
+        "\"DefaultGateways\":\"docker0=172.17.128.1;eth0=172.13.145.1\","
+        "\"DnsServers\":\"docker0=8.8.8.8,172.29.64.1;eth0=172.29.64.1\","
+        "\"DhcpEnabled\":\"docker0=true;eth0=false\","
+        "\"Enabled\":\"docker0=true;eth0=false\","
+        "\"Connected\":\"docker0=true;eth0=false\"}";
```
MariusNi, line 231: Move 4 spaces to right (will stop commenting on this, please apply to all multiline string constants)

src/modules/networking/tests/NetworkingTests.cpp
```diff
@@ -244,29 +256,44 @@ namespace OSConfig::Platform::Tests
...
+        EXPECT_STREQ(resultString.c_str(), payloadInterfaceTypesDataMissing);
 
         EXPECT_NE(payload, nullptr);
         delete payload;
 
+        std::vector<std::string> returnValuesNetworkManagerEnabled = {g_testCommandOutputNames, g_testCommandOutputInterfaceTypesNmcli, g_testIpData, g_testCommandOutputDefaultGateways, g_testCommandOutputDnsServers};
         testModule.runCommandCount = 0;
-        testModule.returnValues = { g_testCommandOutputNames, "", g_testCommandOutputInterfaceTypesNetworkctl, g_testIpData, g_testCommandOutputDefaultGateways, g_testCommandOutputDnsServers };
+        testModule.returnValues = returnValuesNetworkManagerEnabled;
         result = testModule.Get(nullptr, nullptr, nullptr, &payload, &payloadSizeBytes);
```
MariusNi, line 278: >testModule.Get(nullptr [](http://example.com/codeflow?start=17&length=22)

MmiGet with a null session handle must fail. So the vase class Get method must.

src/modules/networking/tests/NetworkingTests.cpp
```diff
@@ -345,6 +502,32 @@ namespace OSConfig::Platform::Tests
...
+
+        EXPECT_NE(payload, nullptr);
+        delete payload;
+
+        testModule.runCommandCount = 0;
+        testModule.returnValues[4] = testCommandOutputOnlyGlobalDnsServers;
+
+        result = testModule.Get(nullptr, nullptr, nullptr, &payload, &payloadSizeBytes);
+
+        EXPECT_EQ(result, MMI_OK);
```
MariusNi, line 524: This is the wrong check. That testModule.Get must fail with a null session handle, not return MMI_OK.

Bug in test and also bug in module.

src/modules/networking/src/lib/Networking.cpp
hachotze: Each should be an 'if' statement. This is where Networking discovers the network management service enabled on the system (Network Manager or systemd-networkd) and caches the result in m_networkManagementService. Code contained in the first 'if' checks if Network Manager is enabled and collects InterfaceTypes data if possible. If Network Manager is not enabled, code contained in the second 'if' checks if systemd-networkd is enabled and collects InterfaceTypes data if possible. If neither Network Manager nor systemd-networkd are enabled, and full logging is enabled, code contained in the third 'if' logs a Networking error. It is important that this flow is implemented as a series of 'if' statements because in cases where one or more target network management services are not found, the module needs to continue trying new queries, or log an error if none of the known services are discovered.

src/modules/networking/tests/NetworkingTests.cpp
hachotze: NetworkingObjectBase::Get does not use any of the arguments passed here as nullptr. It may be best to simply remove the arguments from the method prototype (as is done for the Tpm::Get method)?

src/modules/networking/tests/NetworkingTests.cpp
```diff
@@ -75,79 +76,79 @@ namespace OSConfig::Platform::Tests
...
+            "Current Scopes: DNS\n"
+        "DefaultRoute setting: yes\n"
+            "LLMNR setting: yes\n"
+        "MulticastDNS setting: no\n"
+        "DNSOverTLS setting: no\n"
+            "DNSSEC setting: no\n"
+            "DNSSEC supported: no\n"
+        "Current DNS Server: 172.29.64.1\n"
+                "DNS Servers: 172.29.64.1\n"
+                "DNS Domain: mshome.net\n";
```
MariusNi, line 136: These sub-strings can be aligned at right. And should be... The left " should be all on one column, 4 spaces from left margin in this case

MariusNi (approved):
:shipit:

PR #46, opened 2021-12-13
Jinmen/command runner e2e tests

jinghanm:
## Description

Add more command runner e2e tests

## Checklist


src/tests/e2etest/CommandRunnerTests.cs
```diff
@@ -47,46 +52,208 @@ public partial class CommandStatus
...
+            AssertModuleConnected();
+            // Run an echo Hello World command
+            (CommandArguments desiredCommand, CommandStatus expectedCommandStatus) = GenerateEchoHelloWorldCommand();
+            UpdateDesiredProperties(ComponentName, desiredCommand);
+            // Verify the command status (should be succeeded)
+            AreReportedPropertiesUpdated(expectedCommandStatus, responseStatusSuccess, false);
+        }
+
+        [Test]
+        public void CommandRunnerTest_Same_CommandId()
```
robertschaedler3, line 180: Is it possible to also test this after OSConfig is restarted? This would test that the last commandIt is cached correctly to disk.

src/tests/e2etest/E2ETest.cs
```diff
@@ -270,6 +270,13 @@ public static void AreEqualByJson(object expected, object actual)
             Assert.AreEqual(expectedJson, actualJson);
         }
 
+        public static bool AreJsonObjectsEqual(object expected, object actual)
+        {
+            string expectedJson = JsonSerializer.Serialize(expected);
```
robertschaedler3, line 275: Can we ensure that the serialized string is always the same? For example, 2 JSON object may be the same but could be serialized to a string differently.

AhmedBM (approved):

src/tests/e2etest/CommandRunnerTests.cs
MariusNi: Last command result is persisted as CommandId and numeric result. It can be used for this purpose.

Can the test reboot the device and then try to reboot again using the same CommandId?

This would be a test for the infinite loop of death conditions

robertschaedler3: On second thought, this currently cannot be done since CommandRunner is only caching one command to disk specifically for reboot. We can add a test like I described (same commandId after reboot) later on in the future if/when CommandRunner gets updated.

MariusNi: All last commands are persisted. Not just reboot. This test can be done if the test can reboot the device under test and continue all within same test session.

Can we do this?

src/tests/e2etest/CommandRunnerTests.cs
```diff
@@ -47,46 +52,208 @@ public partial class CommandStatus
...
+            public int ac { get; set; }
+        }
+
+        const int responseStatusSuccess = 200;
+        const int responseStatusFailed = 400;
+
+        const string argumentsPingBing = "ping bing.com";
+        const string argumentsEchoHelloWorld = "echo Hello World";
+        const string helloWorld = "Hello World ";
+        const string argumentsReboot = "reboot";
```
MariusNi, line 66: > const string argumentsReboot = "reboot"; [](http://example.com/codeflow?start=7&length=41)

This is not used. Looks the reboot test was intended but not completed. Possible to complete it in this PR? Canm be combined with a test to validate command with same id cannot be requested after reboot.

MariusNi: This is the same in the 5th iteration...

PR #55, opened 2021-12-21
Update ZTSI module

robertschaedler3:
## Description

Refactor ZTSI module to apply patterns from sample module.

## Checklist

### DIFF
--- a/src/modules/ztsi/src/lib/Ztsi.cpp
+++ b/src/modules/ztsi/src/lib/Ztsi.cpp
@@ -30,6 +30,12 @@ static const std::string g_urlRegex = "((http|https)://)(www.)?[-A-Za-z0-9+&@#/%
 static const unsigned int g_lockWaitMillis = 20;
 static const unsigned int g_lockWaitMaxRetries = 5;
 
+const std::string Ztsi::m_componentName = "Ztsi";
+const std::string Ztsi::m_desiredServiceUrl = "DesiredServiceUrl";
+const std::string Ztsi::m_desiredEnabled = "DesiredEnabled";
+const std::string Ztsi::m_reportedServiceUrl = "ServiceUrl";
+const std::string Ztsi::m_reportedEnabled = "Enabled";
+
 // Regex for validating client name 'Azure OSConfig <model version>;<major>.<minor>.<patch>.<yyyymmdd><build>'
 static const std::string g_clientNameRegex = "^((Azure OSConfig )[1-9];(0|[1-9]\\d*)\\.(0|[1-9]\\d*)\\.(0|[1-9]\\d*)\\.([0-9]{8})).*$";
 static const std::string g_clientNamePrefix = "Azure OSConfig ";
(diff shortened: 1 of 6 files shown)


src/modules/ztsi/src/lib/Ztsi.cpp
```diff
@@ -105,6 +111,63 @@ bool IsValidClientName(const std::string& clientName)
...
+            }
+        }
+        catch (const std::exception& e)
+        {
+            OsConfigLogError(ZtsiLog::Get(), "Could not allocate payload: %s", e.what());
+            status = EINTR;
+
+            if (nullptr != *payload)
+            {
+                delete[] * payload;
```
MariusNi, line 151: >delete[] * payload [](http://example.com/codeflow?start=16&length=18)

nit: no space between '*' and 'payload' ('*payload' is the pointer being freed)

src/modules/ztsi/src/lib/Ztsi.cpp
```diff
@@ -116,6 +179,198 @@ Ztsi::Ztsi(std::string filePath, unsigned int maxPayloadSizeBytes)
...
+        "UserAccount": 0})"""";
+
+    if (nullptr == clientName)
+    {
+        OsConfigLogError(ZtsiLog::Get(), "GetInfo called with null clientName");
+        status = EINVAL;
+    }
+    else if (!IsValidClientName(clientName))
+    {
+        status = EINVAL;
```
MariusNi, line 204: >status = EINVAL; [](http://example.com/codeflow?start=8&length=16)

Add an error log for this

src/modules/ztsi/src/so/ZtsiModule.cpp
```diff
@@ -231,105 +162,11 @@ int MmiSet(
...
-        {
-            delete[] * payload;
-            *payload = nullptr;
-        }
-
-        if (nullptr != payloadSizeBytes)
-        {
-            *payloadSizeBytes = 0;
-        }
+        session = reinterpret_cast<Ztsi*>(clientSession);
```
MariusNi, line 168: >session = reinterpret_cast<Ztsi*>(clientSession); [](http://example.com/codeflow?start=8&length=49)

What if some client calls MpiSet with a non null but invalid handle? This cast would prodice an invalid module object. Add a "IsValid()" member to the module  object to validate itself? (by size, contents, etc)

MariusNi: Same for MpiGet

src/modules/ztsi/tests/ZtsiTests.cpp
```diff
@@ -75,6 +81,34 @@ namespace OSConfig::Platform::Tests
...
     std::string ZtsiTests::filename = "./ztsi/config.temp.json";
 
+    TEST_F(ZtsiTests, GetSetServiceUrl)
+    {
+        char serviceUrl[] = "\"https://localhost:8080/\"";
+        MMI_JSON_STRING payload = nullptr;
+        int payloadSizeBytes = 0;
+
+        ASSERT_EQ(MMI_OK, ztsi->Set(g_componentName, g_desiredServiceUrl, serviceUrl, sizeof(serviceUrl)));
+        ASSERT_EQ(MMI_OK, ztsi->Get(g_componentName, g_reportedServiceUrl, &payload, &payloadSizeBytes));
```
MariusNi, line 91: Add Set and Get tests that use invalid arguments? Null values, unsupported values, etc.

MariusNi: Same for next

MariusNi:
> static const std::string g_urlRegex = "((http|https)://)(www.)?[-A-Za-z0-9+&@#/%?=~_|!:,.;]+[-A-Za-z0-9+&@#/%=~_|]";

This is only used in Ztsi::IsValidConfiguration. Make it local there? Same may apply to other global constants here.

---
Refers to: src/modules/ztsi/src/lib/Ztsi.cpp:27 in ce3be78. [](commit_id = ce3be78f1b17d41fafe87f69022f87bf09975758, deletion_comment = False)

src/modules/ztsi/tests/ZtsiTests.cpp
robertschaedler3: Added here and to Sample CPP module

MariusNi (approved):
:shipit:

PR #59, opened 2022-01-06
Fix cached status after executing reboot or shutdown

robertschaedler3:
## Description

Fixes cached status after executing reboot or shutdown. When the device resumes after a reboot/shutdown command, the correct status is reported.

## Checklist


src/modules/commandrunner/src/so/CommandRunnerModule.cpp
```diff
@@ -359,17 +359,14 @@ void MmiClose(MMI_HANDLE clientSession)
...
         std::string clientName = mmiMap[clientSession].get()->GetClientName();
-        rapidjson::StringBuffer sb;
-        rapidjson::Writer<rapidjson::StringBuffer> writer(sb);
-
         mmiMap[clientSession]->CancelAll();
-        PersistCommandResults();
-        mmiMap[clientSession].reset();
 
-        if (MMI_OK != WriteToCache(sb))
+        if (MMI_OK != PersistCommandResults())
```
MariusNi, line 364: >MMI_OK != PersistCommandResults( [](http://example.com/codeflow?start=12&length=32)

Do not use MMI_OK here, PersistCommandResults is not an MMI call. Use 0 instead.

MariusNi (approved):
:shipit:

PR #65, opened 2022-01-12
Fix MPI desired configuration validation

robertschaedler3:
## Description

Fixes validation on desired configuration to align with [modules.md 3.2.3](https://github.com/Azure/azure-osconfig/blob/main/docs/modules.md#323-serialized-mim-payload-at-run-time). 

Fixes logging of JSON payloads without `IsFullLoggingEnabled()`.

## Checklist

### DIFF
--- a/src/platform/modulesmanager/src/ModulesManager.cpp
+++ b/src/platform/modulesmanager/src/ModulesManager.cpp
@@ -672,10 +672,6 @@ int ModulesManager::MpiSetDesired(const char* clientName, const char* payload, i
             {
                 OsConfigLogError(ModulesManagerLog::Get(), "MpiSetDesired invalid payload: %.*s", payloadSizeBytes, payload);
             }
-            else
-            {
-                OsConfigLogError(ModulesManagerLog::Get(), "MpiSetDesired invalid payload");
-            }
 
             status = EINVAL;
         }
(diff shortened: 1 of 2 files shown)


src/platform/modulesmanager/src/ModulesManager.cpp
```diff
@@ -663,18 +663,33 @@ int ModulesManager::MpiSetDesired(const char* clientName, const char* payload, i
...
         {
-            OsConfigLogError(ModulesManagerLog::Get(), "MpiSetDesired invalid payload: %s", payloadString.c_str());
+            if (IsFullLoggingEnabled())
+            {
+                OsConfigLogError(ModulesManagerLog::Get(), "MpiSetDesired invalid payload: %.*s", payloadSizeBytes, payload);
+            }
+            else
+            {
+                OsConfigLogError(ModulesManagerLog::Get(), "MpiSetDesired invalid payload");
+            }
```
MariusNi, line 678: MpiSetDesired and MpiGetReported are called with high frequency. We need to avoid repetitive loggings that can fill the log files. 

Let's keep only the error logging for when full logging is enabled.

src/platform/modulesmanager/tests/ModuleManagerTests/src/MpiTests.cpp
```diff
@@ -114,13 +114,11 @@ namespace Tests
...
         char payload[] = R""""(
-            [
-                {
-                    "TestComponent1": {
-                        "testParameter": "testValue"
-                    }
+            {
+                "TestComponent1": {
+                    "testParameter": "testValue"
                 }
```
MariusNi, line 120: Can we have more complex test cases here? Multiple components and objects and also settings within objects. 

Maybe do something with one instance for every possible type? E.g. arrays of strings and integers for settings, single and array objects, etc.

robertschaedler3: Agreed 👍🏻

src/platform/modulesmanager/src/ModulesManager.cpp
robertschaedler3: Changed to only log if full logging is enabled

MariusNi (approved):
:shipit:

PR #80, opened 2022-02-02
Add Command Runner E2E Tests

robertschaedler3:
## Description

Adds E2E tests for the CommandRunner module.

## Checklist

### DIFF
--- a/src/tests/e2etest/CommandRunnerTests.cs
+++ b/src/tests/e2etest/CommandRunnerTests.cs
@@ -2,15 +2,18 @@
 // Licensed under the MIT License.
 
 using NUnit.Framework;
-using Microsoft.Azure.Devices.Shared;
-using System.Text.Json;
 using System;
+using System.Text.Json;
 using System.Threading.Tasks;
 
 namespace E2eTesting
 {
     public class CommandRunnerTests : E2eTest
     {
+        private const string componentName = "CommandRunner";
+        private const string commandArguments = "CommandArguments";
+        private const string commandStatus = "CommandStatus";
+
         public enum Action
         {
             None = 0,
@@ -27,9 +30,10 @@ namespace E2eTesting
             Running,
             Succeeded,
             Failed,
-            TimeOut,
+            TimedOut,
             Canceled
         }
+
         public partial class CommandArguments
         {
             public string CommandId { get; set; }
@@ -39,6 +43,14 @@ namespace E2eTesting
             public bool SingleLineTextResult { get; set; }
         }
 
+        public partial class CommandArgumentsResponse
+        {
+            public CommandArguments value { get; set; }
+            public int ac { get; set; }
+            public string ad { get; set; }
+            public int av { get; set; }
+        }
+
         public partial class CommandStatus
         {
             public string CommandId { get; set; }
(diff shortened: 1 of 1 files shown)


src/tests/e2etest/E2ETest.cs
```diff
@@ -24,7 +24,7 @@ public class E2eTest
         private string deviceId = Environment.GetEnvironmentVariable("E2E_OSCONFIG_DEVICE_ID");
         private readonly string moduleId = "osconfig";
         protected int twinTimeoutSeconds = 0;
-        private int twinTimeoutSecondsDefault = 45;
+        private int twinTimeoutSecondsDefault = 180;
         public readonly int twinRefreshIntervalMs = 2000;
```
MariusNi, line 28: This looks strange. The timeout should be greater than the refresh interval, not the other way. Do I read this right?

src/tests/e2etest/CommandRunnerTests.cs
```diff
@@ -39,54 +43,219 @@ public partial class CommandArguments
             public bool SingleLineTextResult { get; set; }
         }
 
+        public partial class CommandArgumentsResponse
+        {
+            public CommandArguments value { get; set; }
+            public int ac { get; set; }
+            public string ad { get; set; }
+            public int av { get; set; }
```
MariusNi, line 51: What is the test doing with these?

src/tests/e2etest/E2ETest.cs
robertschaedler3: Correct, the timeout is in seconds but the refresh interval is in milliseconds.

src/tests/e2etest/CommandRunnerTests.cs
robertschaedler3: Test is checking `ac` (acknowledge code) and the others are there to "fill out" the rest of the JSON object in the twin

MariusNi: Cool. However, how are they filled?
av for example is version.
OSConfig sets that to same version that comes in the request. Same version goes in the acknowledge.
These should not be set blindly. May produce unexpected behavior with the Reported Twin.

src/tests/e2etest/E2ETest.cs
MariusNi: Let's shorten the refresh to 5000 msec and increase the timeout to 30 minutes of 1 hour.

src/tests/e2etest/E2ETest.cs
```diff
@@ -24,7 +24,7 @@ public class E2eTest
         private string deviceId = Environment.GetEnvironmentVariable("E2E_OSCONFIG_DEVICE_ID");
         private readonly string moduleId = "osconfig";
         protected int twinTimeoutSeconds = 0;
-        private int twinTimeoutSecondsDefault = 45;
+        private int twinTimeoutSecondsDefault = 180;
```
AhmedBM, line 27: Not sure i agree with bumping the global timeout. The timeout should be sufficient given the reporting interval on the tests is 30s, twin updates will occur before this. I suspect something else is at play here if it is necessary to bump the timeout. Maybe this was due to the stale twin when checking for the `WaitForCommandStatus` call I commented above?

AhmedBM (changes requested):

src/tests/e2etest/CommandRunnerTests.cs
AhmedBM: After looking at this, we might want to bring this into the base `E2eTest` base class. We use this for brokering the twin updates, so it makes sense to bring this into all tests.
I would suggest adding this to `UpdateTwinBlockUntilUpdate`?

src/tests/e2etest/E2ETest.cs
MariusNi: These tests must continue to run w/o failure if we change the default reporting interval in the future. The tests should be agnostic to that.

AhmedBM: yeah good point. I agree it should work for all intervals, not sure how we can adjust for that though but as far as our e2etests are concerned, we always configure our agents to use 30s. I just mean to say here that having to bump the timeout for the tests to work is concerning, we shouldn't have to do this

src/tests/e2etest/E2ETest.cs
robertschaedler3: Fixed to refresh to 0.5sec and timeout to 30min

src/tests/e2etest/CommandRunnerTests.cs
robertschaedler3: Removed `ad` and `av`. Will move this to E2eTest base class when refactoring/adding E2E test tooling

src/tests/e2etest/E2ETest.cs
robertschaedler3: Bumped the timeout as per an earlier comment. When tests are refactored and tooling is added to aid E2E test development, we should be able to reduce this/abstract it away from the tests entirely.

src/tests/e2etest/CommandRunnerTests.cs
```diff
@@ -39,54 +43,219 @@ public partial class CommandArguments
...
         public partial class CommandStatus
         {
             public string CommandId { get; set; }
             public long ResultCode { get; set; }
             public string TextResult { get; set; }
             public CommandState CurrentState { get; set; }
         }
+        public CommandStatus WaitForCommandStatus(string commandId, CommandState commandState, int maxWaitSeconds = 180)
+        {
+            var reportedCommandStatus = JsonSerializer.Deserialize<CommandStatus>(GetTwin().Properties.Reported[componentName][commandStatus].ToString());
```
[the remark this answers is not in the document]
robertschaedler3: Not necessarily, a new twin is obtained during each iteration of the loop below. I tried with `GetNewTwin()` there and received hit some errors due to the way the `twin` is stored in the `E2eTest` base class. I already have plans to address this during a subsequent PR to add tooling/*hide the twin* from each individual test.

AhmedBM (approved):

MariusNi (approved):
:shipit:

PR #101, opened 2022-02-23
Refactor E2ETest base class

robertschaedler3:
## Description

This PR introduces a refactored `E2ETest` base class starting with the command runner E2E tests. This change standardizes the interface used by individual tests to interact with the twin by encapsulating all twin related actions (update desired, wait for reported, access to module twin, access to registry manager) in the base test class. This also ensures that the most up to date twin is polled while waiting for twin updates. 

## Checklist


src/tests/e2etest/CommandRunnerTests.cs
```diff
@@ -1,19 +1,279 @@
...
 using System;
 using System.Text.Json;
 using System.Threading.Tasks;
 
 namespace E2eTesting
 {
-    public class CommandRunnerTests : E2eTest
+
+    [TestFixture]
+    public abstract class E2ETest
```
hachotze, line 15: Why is this class defined in CommandRunnerTests.cs instead of E2ETest.cs?

src/tests/e2etest/CommandRunnerTests.cs
```diff
@@ -1,19 +1,279 @@
...
+
+        private readonly string _iotHubConnectionString = Environment.GetEnvironmentVariable("E2E_OSCONFIG_IOTHUB_CONNSTR")?.Trim('"');
+        private readonly string _moduleId = "osconfig";
+        private readonly string _deviceId = Environment.GetEnvironmentVariable("E2E_OSCONFIG_DEVICE_ID");
+
+        private readonly string _sasToken = Environment.GetEnvironmentVariable("E2E_OSCONFIG_SAS_TOKEN");
+        private readonly string _uploadUrl = Environment.GetEnvironmentVariable("E2E_OSCONFIG_UPLOAD_URL")?.Trim('\"');
+        private readonly string _resourceGroupName = Environment.GetEnvironmentVariable("E2E_OSCONFIG_RESOURCE_GROUP_NAME")?.Trim('\"');
+
+        public partial class GenericResponse<T>
```
hachotze, line 27: Why is GenericResponse defined as a partial class?

src/tests/e2etest/CommandRunnerTests.cs
```diff
@@ -1,19 +1,279 @@
...
+                if (200 != ackCode)
+                {
+                    Console.WriteLine("[ExecuteCommandViaCommandRunner] Failed to set desired state");
+                    success = false;
+                }
+                else
+                {
+                    Func<CommandRunnerTests.CommandStatus, bool> condition = (CommandRunnerTests.CommandStatus status) =>
+                    {
+                        return status.CommandId == command.CommandId && status.CurrentState == CommandRunnerTests.CommandState.Succeeded;
```
hachotze, line 91: Consider adding parentheses around equality tests.

src/tests/e2etest/CommandRunnerTests.cs
robertschaedler3: The plan of action for now is to try out these changes with the CommandRunner E2E tests first, since this is where we are seeing most of the timeout issues, then bring these changes to the `E2ETest.cs` and refactor the other E2E tests to use this new proposed base class.

src/tests/e2etest/CommandRunnerTests.cs
robertschaedler3: Good catch, this does not need to be marked as `partial`(was leftover from when I was testing something).

src/tests/e2etest/CommandRunnerTests.cs
```diff
@@ -1,19 +1,279 @@
...
+
+                string fileName = String.Format("{0}-{1}.tar.gz", _resourceGroupName, _deviceId);
+                string fullURL = String.Format("{0}{1}?{2}", _uploadUrl, fileName, _sasToken);
+
+                string tempFileCommand = "temp_file=`sudo mktemp`";
+                string tarCommand = "sudo tar -cvzf $temp_file osconfig*.log";
+                string curlCommand = String.Format("curl -X PUT -T $temp_file -H \"x-ms-date: $(date -u)\" -H \"x-ms-blob-type: BlockBlob\" \"{1}\"", fileName, fullURL);
+                string uploadCommand = String.Format("cd /var/log && {0} && {1} && {2}", tempFileCommand, tarCommand, curlCommand);
+
+                // Temporarily commented out to avoid uploading logs to blob store twice
```
AhmedBM: Im confused about where the duplicate happens? How are the logs now uploaded to blob store?

src/tests/e2etest/CommandRunnerTests.cs
```diff
@@ -1,19 +1,279 @@
...
+            bool success = true;
+            var command = CommandRunnerTests.CreateCommand(arguments);
+
+            try
+            {
+                var desiredResult = SetDesired<CommandRunnerTests.CommandArguments>("CommandRunner", "CommandArguments", command);
+                desiredResult.Wait();
+                int ackCode = desiredResult.Result;
+
+                if (200 != ackCode)
```
AhmedBM, line 55: consider using the `HttpStatusCode::OK` constant - see [OK](https://docs.microsoft.com/en-us/dotnet/api/system.net.httpstatuscode?view=net-6.0#system-net-httpstatuscode-ok)

src/tests/e2etest/CommandRunnerTests.cs
```diff
@@ -1,19 +1,279 @@
...
+                    success = false;
+                }
+                else
+                {
+                    Func<CommandRunnerTests.CommandStatus, bool> condition = (CommandRunnerTests.CommandStatus status) =>
+                    {
+                        return status.CommandId == command.CommandId && status.CurrentState == CommandRunnerTests.CommandState.Succeeded;
+                    };
+
+                    var reportedResult = GetReported<CommandRunnerTests.CommandStatus>("CommandRunner", "CommandStatus", condition, 180);
```
AhmedBM, line 67: what does the `180` represent?

src/tests/e2etest/CommandRunnerTests.cs
AhmedBM: Why not just try them with all of them and not just the CommandRunner tests?

src/tests/e2etest/CommandRunnerTests.cs
```diff
@@ -1,19 +1,279 @@
...
+            if (!reported.Contains(componentName))
+            {
+                Assert.Warn("{0} not found in reported properties", componentName);
+            }
+
+            GenericResponse<T> response = JsonSerializer.Deserialize<GenericResponse<T>>(reported[componentName].ToString());
+            return response.ac;
+        }
+
+        protected async Task<int> SetDesired<T>(string componentName, string objectName, T value, int ackCode = 200, int maxWaitSeconds = 90)
```
AhmedBM, line 130: im not crazy about these hardcoded defaults, can we define a global default which is overridable from the pipeline (Environment Var maybe?) and use that as the default?

AhmedBM (commented):
Great refactor! Looks great, but why just scope this to CommandRunner tests, lets test it out for all of them. We can easily trigger it from our topic branches to test reliability

src/tests/e2etest/CommandRunnerTests.cs
AhmedBM: double of your 90 timeout im guessing? Can we find a constant to use based on others?

src/tests/e2etest/CommandRunnerTests.cs
robertschaedler3: Duplicate would be due to the separate base classes for CommandRunner and all other tests. I am moving forward with using this refactored class for all tests and there will be a single log upload from `E2ETest.cs`.

src/tests/e2etest/CommandRunnerTests.cs
robertschaedler3: Good call! I'll base this number (and others) off of the env variable previously used for timeout.

robertschaedler3:
Next iteration will contain the changes to use this refactored base test class for all the E2E tests

AhmedBM:
Awesome! Thanks @robertschaedler3

src/tests/e2etest/E2ETest.cs
```diff
@@ -1,294 +1,375 @@
...
+
+        private readonly string _sasToken = Environment.GetEnvironmentVariable("E2E_OSCONFIG_SAS_TOKEN");
+        private readonly string _uploadUrl = Environment.GetEnvironmentVariable("E2E_OSCONFIG_UPLOAD_URL")?.Trim('\"');
+        private readonly string _resourceGroupName = Environment.GetEnvironmentVariable("E2E_OSCONFIG_RESOURCE_GROUP_NAME")?.Trim('\"');
+
+        private const int POLL_INTERVAL_MS = 1000;
+        private const int MAX_WAIT_TIME_SECONDS = 90;
+
+        protected const int ACK_SUCCESS = 200;
+        protected const int ACK_ERROR = 400;
```
MariusNi, line 31: Can we expand these to treat any 2xx value as success and and any 4xx value as error?

MariusNi (dismissed):
:shipit:

src/tests/e2etest/CommandRunnerTests.cs
```diff
@@ -130,23 +105,26 @@ public CommandStatus CreateCommandStatus(string commandId, string textResult = "
             };
         }
 
-        public string CreateCommandId()
+        public void SendCommand(CommandArguments command, int expectedAckCode = ACK_SUCCESS)
         {
-            return Convert.ToBase64String(Guid.NewGuid().ToByteArray()).Substring(0, 4);
-        }
+            int ackCode = -1;
```
AhmedBM, line 110: Can we use a constant here?

src/tests/e2etest/CommandRunnerTests.cs
```diff
@@ -155,7 +133,7 @@ public void CancelCommand(string commandId)
             SendCommand(CreateCancelCommand(commandId));
         }
 
-        public void RefreshCommandStatus(string commandId, int responseCode = 200)
+        public void RefreshCommandStatus(string commandId, int ackCode = ACK_SUCCESS)
```
AhmedBM, line 136: can we use `returnCode` instead?

AhmedBM (dismissed):
Looks good! Minor comments on variable naming.

AhmedBM:
Ohh and make sure the E2E pipelines are not broken before merging :)

src/tests/e2etest/E2ETest.cs
robertschaedler3: For this I recommend we keep specific values since each test expects a specific ack code depending on value that is set. Also it is the responsibility of each individual test to check this value, so if a range is expected, the test must validate the returned ack code against the range. In the tests we have now, all are expecting a specific value.

MariusNi (approved):
:shipit:

PR #110, opened 2022-03-03
Update MIM file for PackageManagerConfiguration module

Ioana37:
## Description

The purpose of this PR is to update the MIM file for the PackageManagerConfiguration module.

## Checklist

### DIFF
--- a/src/modules/mim/packagemanagerconfiguration.json
+++ b/src/modules/mim/packagemanagerconfiguration.json
@@ -15,15 +15,8 @@
               {
                 "name": "Packages",
                 "schema": {
-                  "type": "map",
-                  "mapKey":  {
-                    "name": "Name",
-                    "schema": "string"
-                  },
-                  "mapValue": {
-                    "name": "Version",
-                    "schema": "string"
-                  }
+                  "type": "array",
+                  "elementSchema": "string"
                 }
               },
               {
@@ -32,46 +25,18 @@
               },
               {
                 "name": "SourcesFingerprint",
+                "schema": "string"
+              },
+              {
+                "name": "SourcesFilenames",
                 "schema": {
-                  "type": "map",
-                  "mapKey":  {
-                    "name": "Id",
-                    "schema": "string"
-                  },
-                  "mapValue": {
-                    "name": "Fingerprint",
-                    "schema": "string"
-                  }
+                  "type": "array",
+                  "elementSchema": "string"
                 }
               },
               {
                 "name": "ExecutionState",
-                "schema": {
-                  "type": "enum",
-                  "valueSchema": "integer",
-                  "enumValues": [
-                    {
-                      "name": "Unknown",
-                      "enumValue": 0
-                    },
-                    {
-                      "name": "Running",
-                      "enumValue": 1
-                    },
-                    {
-                      "name": "Succeeded",
-                      "enumValue": 2
-                    },
-                    {
-                      "name": "Failed",
-                      "enumValue": 3
-                    },
-                    {
-                      "name": "TimedOut",
-                      "enumValue": 4
-                    }
-                  ]
-                }
+                "schema": "string"
               }
             ]
           }
--- a/src/modules/mim/packagemanagerconfiguration.json
+++ b/src/modules/mim/packagemanagerconfiguration.json
@@ -36,6 +36,76 @@
               },
               {
                 "name": "ExecutionState",
+                "schema": {
+                  "type": "enum",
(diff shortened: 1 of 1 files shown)


src/modules/mim/packagemanagerconfiguration.json
```diff
@@ -15,15 +15,8 @@
...
-                  "type": "map",
-                  "mapKey":  {
-                    "name": "Name",
-                    "schema": "string"
-                  },
-                  "mapValue": {
-                    "name": "Version",
-                    "schema": "string"
-                  }
+                  "type": "array",
```
MariusNi, line 18: >"type": "array", [](http://example.com/codeflow?start=18&length=16)

We are still discussing to resolve the problem Magda hit with being unable to remove map elements. That should be supported and we are hoping is just wrong serialization. We should have maps for reported to match desired.

So Ioana I hope will be OK to wait a bit longer on this PR.

Thanks

Salazander: Using an array here is a conscious choice: We need "replace" behavior, not "merge", hence the array. The other discussion thread with Magda is related to the Sample module, I'll come back to that in a separate e-mail.

We could think of using a map in the future if any customer were to ask to have the current package version of all previously installed packages. For that we'd have to implement proper caching to disk to keep track of all packages that were installed in the past. This additional complexity we'd like to avoid at this point in time, because the current customer will essentially always install the same two packages with updated versions. Desired states are captured as DEB packages containing a long list of dependencies using the control-field in the DEB packages' metadata.

MariusNi: Just a comment on "We could think of using a map in the future". Once this MIM is done, a PnP interface will be published for it. To make a change a new OSConfig model will have to be published with a new version of the interface. The original interface version would remain. 

And speaking of this, you are planning to prepare a PnP interface for this MIM once the module implementation is complete, yes?

For more details see:  [https://github.com/Azure/azure-osconfig/blob/main/docs/modules.md#14-publishing-dtdl-for-the-module](https://github.com/Azure/azure-osconfig/blob/main/docs/modules.md#14-publishing-dtdl-for-the-module)

src/modules/mim/packagemanagerconfiguration.json
```diff
@@ -32,46 +25,18 @@
...
-                      "name": "Failed",
-                      "enumValue": 3
-                    },
-                    {
-                      "name": "TimedOut",
-                      "enumValue": 4
-                    }
-                  ]
-                }
+                "schema": "string"
```
MariusNi, line 39: >"schema": "string" [](http://example.com/codeflow?start=16&length=18)

"State" implies a well defined enumeration of  values or similar. A string is not that. Why are you changing from the enum of integers to plain string? What will this string contain?

Salazander: The module's folder will contain a README.md explaining all the details. The PR is being worked on at the moment, if you want to catch an early [glimpse](https://github.com/Salazander/azure-osconfig/pull/16).

Generally speaking, our module has a general state and a substate. When an error occurs, we want to provide information about which package/source caused it. When running through the package installation process, we want to report back which package is being processed. This gives meaningful insights to the operator in case of healthy and unhealthy operating conditions. 

Packing this into one string allows a concise human- and machine-readable way to encode state and context.

src/modules/mim/packagemanagerconfiguration.json
Salazander: Thanks for the heads-up! 
Yes, we're planning on publishing the PnP interface once the module has made its way back to upstream.

src/modules/mim/packagemanagerconfiguration.json
MariusNi: The MIM and DTDL have well defined structure to define such configuration. This helps the remote administrators. Otherwise all reported would be a single string.

Please split this string into meaningful MIM settings such as an integer for error|status code, enumeration of integers for reason, string for status description, etc.

Salazander: yes, we will make this change, updated MIM is coming soon.

MariusNi (approved):
:shipit:

src/modules/mim/packagemanagerconfiguration.json
MariusNi: Looks good to me. Thanks

robertschaedler3 (approved):

PR #115, opened 2022-03-07
Fixed E2E tests when multiple packages exist

AhmedBM:
## Description

* Multiple packages are being produced by the `Package Build` pipeline, both signed and unsigned as such, the `DownloadPipelineArtifact` stage should pick one of the packages. Added `head -n 1` to capture the first package.
* Removed `continueOnError` to have an explicit failure when packages fail to be copied or device identities fail to be created as they would fail anyways at further stages.

## Checklist

### DIFF
--- a/devops/pipeline/e2etest.yaml
+++ b/devops/pipeline/e2etest.yaml
@@ -162,14 +162,12 @@ stages:
           itemPattern: $(packagePattern)
 
       - script: |
-          cp `find . -name '*.deb'` $(Build.SourcesDirectory)/devops/terraform/host/osconfig.deb
+          cp `find . -name '*.deb' | head -n 1` $(Build.SourcesDirectory)/devops/terraform/host/osconfig.deb
         displayName: Stage OSConfig
-        continueOnError: true
         workingDirectory: $(Pipeline.Workspace)
 
       - task: AzureCLI@2
         displayName: Create device identity
-        continueOnError: true
         name: azcli_device_identity
         inputs:
           azureSubscription: $(SERVICE_CONNECTION)


devops/pipeline/e2etest.yaml
```diff
@@ -162,14 +162,12 @@ stages:
           itemPattern: $(packagePattern)
 
       - script: |
-          cp `find . -name '*.deb'` $(Build.SourcesDirectory)/devops/terraform/host/osconfig.deb
+          cp `find . -name '*.deb' | head -n 1` $(Build.SourcesDirectory)/devops/terraform/host/osconfig.deb
```
robertschaedler3, line 165: Is this a direct result of adding the signed packages, or was this also possible before?

robertschaedler3: I'm curious what the artifact name for signed vs unsigned looks like?

robertschaedler3 (approved):

robertschaedler3:
Also @AhmedBM, can you trigger a package build/E2E run for this branch before merging?

AhmedBM:
@robertschaedler3 i actually already tested my changes, you might not have seen the package build because i changed the filter in the tests to just take from `main` so i don't have to wait as long to test the changes but it has been tested.

devops/pipeline/e2etest.yaml
AhmedBM: its a direct result of adding them. We stage the package based on the `packagePattern` but this pattern now matches TWO packages, the signed and the unsigned package ex using pattern "**\*bionic_x86_64"
```
./OSConfig_1.0.1.2022030701_SIGNED/osconfig_1.0.1.20220307_bionic_x86_64.deb
./OSConfig_1.0.1.2022030701_ubuntu18.04_amd64/osconfig_1.0.1.20220307_bionic_x86_64.deb
```

PR #132, opened 2022-03-16
Allow null for objects

Salazander:
## Description

The customer noticed that setting objects to null is not passing the JSON validation imposed by mim.object.schema.json.

Here's an example json payload (Desired Properties) that doesn't pass the validation:
`{"Packages":["mariadb-client","cowsay-"], "Sources": null }`

which results in this validation output:
`jsonschema --instance wrong.json mim.object.schema.json
{'Packages': ['mariadb-client', 'cowsay-'], 'Sources': None}: {'Packages': ['mariadb-client', 'cowsay-'], 'Sources': None} is not valid under any of the given schemas`

This PR adds `null` as valid object value. We leave it at your discretion to create a new branch to make more changes if required.

## Checklist

- [] ~~I added unit-tests to validate my changes.~~ **All unit tests are passing: NO** This test fails: `TEST_F(CommonUtilsTest, OsProperties) `because `/sys/devices/virtual/dmi/id/` doesn't exist on our RPI4 running Ubuntu 20.04. This is not related to the schema change proposed by this PR.

### DIFF
--- a/src/modules/schema/mim.object.schema.json
+++ b/src/modules/schema/mim.object.schema.json
@@ -65,6 +65,9 @@
           },
           {
             "$ref": "#/definitions/integerMap"
+          },
+          {
+            "type": "null"
           }
         ]
       }


src/modules/schema/mim.object.schema.json
```diff
@@ -65,6 +65,9 @@
           },
           {
             "$ref": "#/definitions/integerMap"
+          },
+          {
+            "type": "null"
```
MariusNi, line 70: This is invalid schema.

MIM does not allow for null object payloads. 

When there is no data to be reported, the module must fill in with blanks or default values and report that (values of  0 and empty strings, etc, or else) using correct object schema.

This PR cannot be accepted.

MariusNi (changes requested):
This is invalid schema. This PR cannot be accepted.

src/modules/schema/mim.object.schema.json
Salazander: https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-device-twins#back-end-operations
clearly states that desired properties must be removed by setting them to 'null'.

Our module cannot respond to 'null' values because the payload never reaches the module, the payload gets discarded before that. It cannot fill blanks in any way.

Salazander: Here's where the code aborts the processing of the payload: https://github.com/Azure/azure-osconfig/blob/4bd316b58e094cc635c29437c1a6c6fb88dc3f4e/src/agents/pnp/MpiProxy.c#L65
And here are the verbose logs for completeness

```
[2022-03-17 05:45:43] [PnpUtils.c:386] ModuleTwinCallback: received {"PackageManagerConfiguration":{"DesiredState":{"Packages":["cowsay"],"OldProperty":null}},"$version":34} (105 bytes)
[2022-03-17 05:45:43] [PnpUtils.c:335] Queued desired payload of 105 bytes at slot 3
[2022-03-17 05:45:43] [PnpAgent.c:295] Processing desired twin updates
[2022-03-17 05:45:43] [PnpUtils.c:223] ProcessJsonFromTwin: DEVICE_TWIN_UPDATE_PARTIAL
[2022-03-17 05:45:43] [PnpUtils.c:142] PropertyUpdateFromIotHubCallback: invoking PackageManagerConfiguration for property DesiredState, version 34
[2022-03-17 05:45:43] [PnpUtils.c:647] PackageManagerConfiguration.DesiredState: received {"Packages":["cowsay"],"OldProperty":null} (42 bytes)
[2022-03-17 05:45:43] [CommonUtils.cpp:242] [ERROR] MIM object JSON payload is invalid according to the schema
[2022-03-17 05:45:43] [CommonUtils.cpp:249] [ERROR] Invalid JSON: '{"Packages":["cowsay"],"OldProperty":null}' (42 bytes)
[2022-03-17 05:45:43] [MpiProxy.c:77] MpiSet(0xaaab1511e690, PackageManagerConfiguration, DesiredState, {"Packages":["cowsay"],"OldProperty":null}, 42 bytes) returned 22
[2022-03-17 05:45:43] [PnpUtils.c:660] [ERROR] PackageManagerConfiguration.DesiredState: MpiSet failed with 22
```

Salazander: Please read my comments again, I've been talking of **desired** properties.

Salazander: We will instruct the customer to modify the payload they send since setting a field within a MIM-object to null is not supported.

PR #137, opened 2022-03-18
Add PackageManagerConfiguration module skeleton

MagdaPaj:
## Description

* Add skeleton of the PackageManagerConfiguration module. There is an empty implementation of the Set and Get methods, only status `MMM_OK` is returned. Future Pull Request will contain proper implementation. The skeleton is based on the Sample module code. 

## Checklist

- [x] I added unit-tests to validate my changes. All unit tests are passing apart from CommonUtilsTest.OsProperties which fails on the Raspberry Pi 4 and it's a known issue.


src/modules/packagemanagerconfiguration/src/lib/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+cmake_minimum_required(VERSION 3.2)
+
+set (CMAKE_CXX_STANDARD 17)
+
```
MariusNi, line 7: C++ 17?

Use instead what the rest of OSConfig does:

cmake_minimum_required(VERSION 3.2)
set (CMAKE_CXX_STANDARD 11)

here and elsewhere

src/modules/CMakeLists.txt
```diff
@@ -7,6 +7,7 @@ set(MODULES_INSTALL_DIR "/usr/lib/osconfig" CACHE FILEPATH "Directory used for i
 add_subdirectory(commandrunner)
 add_subdirectory(firewall)
 add_subdirectory(networking)
+add_subdirectory(packagemanagerconfiguration)
```
MariusNi, line 10: >packagemanagerconfiguration [](http://example.com/codeflow?start=17&length=27)

This name is excessively long and this results in the coed and file names for this module to be also super long. Can we replace it with a shorter name? 

How about: "PmConfig"? Let's discuss

MagdaPaj: We agreed on that name in the MIM json file, so the code and file names are consistent with it.

Would you like to completely change this name also in the MIM json file? Or only file names and classes name?

src/modules/packagemanagerconfiguration/src/lib/CMakeLists.txt
MagdaPaj: The reason why we need C++ 17 is a [filesystem library](https://en.cppreference.com/w/cpp/filesystem) for performing operations on file systems.
C++ 17 was released already 5 years ago, so what are your concerns with using that version?

MariusNi: C11 is the standard for OSConfig. C99 is the standard for the PnP SDK. Please change to C11.

src/modules/CMakeLists.txt
MariusNi: That long name is a description. Imagine if instead of OSConfig it would be OperatingSystemConfiguration.

Please change to a shorter name in the code and also in the MIM if you want them to match (they do not have to match exactly)

MagdaPaj: OS is way more obvious that it means OperatingSystem vs PM meaning PackageManager.

PmConfig is not clear at all that it is referring to packages. Previously we had several other proposals:
- AptInstall
- PackageTool
- PackageUpdate
- PackageController
- PackageManager
- PackageHandler
- Packages
Could you let us know which one you would find more suitable?

/cc @MatthewMWR

MatthewMWR: There are multiple interleaved considerations here, let's try separating them out first and then consolidating.

1. "Config" vs "Configuration": Subjectively I prefer Config (shorter, still obvious meaning, aligns with OSConfig name), but I thought at one point someone had asserted that abbreviations weren't allowed in the in object model? Am I remembering wrong? Is there such a rule? If so, should we consider changing it?

2. "PackageManager" vs "PM" vs "PacMan" vs "Apps" vs "Applications", vs "Components", etc. I think we had a long debate about this a month ago (approx), and settled on PackageManager for various reasons. For example, in OS/Linux context "apps" or "applications" have different connotations from "packages" which are often libraries. "Apt" is too specific some something that could be a cross-OS API/model, etc.

As of now, I feel most aligned to "PackageManagerConfig" as the name in the object model. I have no opinion about the names of classes, etc.

MariusNi: The MIM name can remain "PackageManagerConfiguration". The MIM was already approved in a separate PR.

This is about the names used for the source code and the binaries.

I am proposing to do here similar to ZTSI. We also have a very long name here. Let's use PMC. Aka for example: "pmclib" etc.

MagdaPaj: changed as suggested

src/modules/packagemanagerconfiguration/src/lib/CMakeLists.txt
MagdaPaj: fixed as suggested

src/modules/pmc/src/lib/PmcBase.cpp
```diff
@@ -0,0 +1,106 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include "PmcBase.h"
+
+constexpr const char ret[] = R""""({
+    "Name": "PackageManagerConfiguration Module",
```
MariusNi, line 7: >PackageManagerConfiguration Module [](http://example.com/codeflow?start=13&length=34)

"PackageManagerConfiguration (PMC) Module"?

So we tell users what our short alias is.

src/modules/pmc/tests/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+project(pmctests)
+
+cmake_minimum_required(VERSION 3.2.0)
+
```
MariusNi, line 7: This is now specified in the root CMakeList.txt for the project. Let's omit it from here.

MagdaPaj: All other tests files have this line, that's why I left it. But now I've deleted it.

src/modules/pmc/src/lib/PmcBase.h
```diff
@@ -0,0 +1,49 @@
...
+// Licensed under the MIT License.
+
+#include <string>
+#include <rapidjson/stringbuffer.h>
+
+#include <CommonUtils.h>
+#include <Logging.h>
+#include <Mmi.h>
+
+#define PMC_LOGFILE "/var/log/osconfig_packagemanagerconfiguration.log"
```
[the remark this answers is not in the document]
MagdaPaj: So originally I left the long name because I consider this as admin facing data. And for admins it may be more obvious to search for the osconfig_packagemanagerconfiguration.log rather than osconfig_pmc.log.

But for the moment I've changed it as you suggested.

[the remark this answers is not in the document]
MariusNi: The local logs are not facing remote admin data. They are local and need to match the local binary name first and foremost. The remote admin can deduct what pmc means from the description in the PnP interface, please remember to add it there

MariusNi (approved):
:shipit:

PR #140, opened 2022-03-22
Consolidating CMake and C/C++ Standard for all of OSConfig

MariusNi:
## Description

The OSConfig code was not uniformly applying same C/C++ standard and as result, we had code that was compiled in whatever local gcc compiler support was available. This PR consolidates that in a single place in the root CMakeList.txt. The only exception is the Module Manager tests that currently can only compile in C14 or newer (we need to fix this).


AhmedBM (approved):
👌

PR #144, opened 2022-03-25
Add logic for updating packages in PMC module

Ioana37:
## Description

The purpose of this PR is to add the logic, in the PackageManagerConfiguration module, for updating packages . Package updating is done by executing apt-get commands (apt-get update, apt-get install). The expected package format is the same as for apt-get install [command](https://linux.die.net/man/8/apt-get).
The PR contains a README.md file with details about the functionality of the module and the execution states and substates.

## Checklist


src/modules/pmc/src/lib/ExecutionState.h
```diff
@@ -0,0 +1,42 @@
...
+
+enum SubStateComponent
+{
+    None,
+    DeserializingJsonPayload,
+    DeserializingDesiredState,
+    DeserializingPackages,
+    UpdatingPackagesLists,
+    InstallingPackages
+};
```
MariusNi, line 23: Can these be made internal to the ExecutionState class or a parent namespace be added? Just to not have the enums as global and risk to collide with others.

src/modules/pmc/src/lib/ExecutionState.h
```diff
@@ -0,0 +1,42 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include <string>
```
MariusNi, line 4: Try to include headers from source files, not from headers. Move this out from here.

src/modules/pmc/src/lib/ExecutionState.cpp
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include "ExecutionState.h"
```
MariusNi, line 4: Include string.h before. Here and elsewhere this is needed

src/modules/pmc/src/lib/Pmc.h
```diff
@@ -2,10 +2,13 @@
 // Licensed under the MIT License.
 
 #include <PmcBase.h>
+#include <string>
```
MariusNi, line 6: You can also create a common header for all includes this module needs and include that everywhere.

src/modules/pmc/src/lib/Pmc.h
```diff
@@ -2,10 +2,13 @@
...
 
 class Pmc : public PmcBase
 {
 public:
     Pmc(unsigned int maxPayloadSizeBytes);
     ~Pmc() = default;
+
+    int RunCommand(const char* command, bool replaceEol, std::string* textResult, unsigned int timeoutSeconds) override;
```
MariusNi, line 13: >bool replaceEol, [](http://example.com/codeflow?start=40&length=16)

Do you really need to use this? For all internal commands this could be hardcoded

MariusNi, line 13: >, unsigned int timeoutSeconds) [](http://example.com/codeflow?start=80&length=30)

This module won't need timeouts either other than the default 60 seconds that is automatically applied when timeout to commonutils!ExecuteCommand is 0.

src/modules/pmc/src/lib/Pmc.cpp
```diff
@@ -2,8 +2,35 @@
...
+        if (buffer && textResult)
+        {
+            *textResult = buffer;
+        }
+    }
+    else if (IsFullLoggingEnabled())
+    {
+        OsConfigLogError(PmcLog::Get(), "RunCommand failed with status: %d and output '%s'", status, buffer);
+    }
+
```
MariusNi, line 30: Why duplicate the log without adding any new information? commonutils logs. See if the module can add anything else useful.

src/modules/pmc/src/lib/Pmc.cpp
```diff
@@ -2,8 +2,35 @@
...
+    if (buffer)
+    {
+        free(buffer);
+    }
```
MariusNi, line 34: Don't do this. Instead, call the commonutils FREE_MEMORY

src/modules/pmc/src/lib/Pmc.cpp
```diff
@@ -2,8 +2,35 @@
...
+#include <CommonUtils.h>
+#include <Mmi.h>
+#include <string>
 
 Pmc::Pmc(unsigned int maxPayloadSizeBytes)
     : PmcBase(maxPayloadSizeBytes)
 {
+}
+
+int Pmc::RunCommand(const char* command, bool replaceEol, std::string* textResult, unsigned int timeoutSeconds)
```
MariusNi, line 14: >Pmc::RunCommand(const char* command, bool replaceEol, std::string* textResult, unsigned int timeoutSeconds) [](http://example.com/codeflow?start=4&length=107)

Skip the replaceEol and timeoutSeconds and assume defaults for this module (such as: true and 0)

src/modules/pmc/src/lib/PmcBase.h
```diff
@@ -36,14 +42,45 @@ class PmcLog
...
+    std::vector<std::string> GetReportedPackages(std::vector<std::string> packages);
+    int DeserializeDesiredState(rapidjson::Document& document, DesiredState& object);
+    int ValidateAndGetPackagesNames(std::vector<std::string> packagesLines);
+    static std::vector<std::string> ListFiles(const char* dir, const char* fileNameExtension);
+    static int SerializeState(State reportedState, MMI_JSON_STRING* payload, int* payloadSizeBytes, unsigned int maxPayloadSizeBytes);
+    static int CopyJsonPayload(rapidjson::StringBuffer& buffer, MMI_JSON_STRING* payload, int* payloadSizeBytes);
+    static std::vector<std::string> Split(const std::string& str, const std::string& delimiter);
+    static std::string TrimStart(const std::string& str, const std::string& trim);
+    static std::string TrimEnd(const std::string& str, const std::string& trim);
+    static std::string Trim(const std::string& str, const std::string& trim);
```
MariusNi, line 80: >str [](http://example.com/codeflow?start=47&length=3)

Use full words. No "str" please. Use "value" or something else.

src/modules/pmc/src/lib/PmcBase.h
```diff
@@ -36,14 +42,45 @@ class PmcLog
...
 
 private:
+    bool CanRunOnThisPlatform();
+    int ExecuteUpdate(const std::string& value);
+    int ExecuteUpdates(const std::vector<std::string> packages, bool updatePackageLists);
+    std::string GetFingerprint();
+    std::vector<std::string> GetReportedPackages(std::vector<std::string> packages);
+    int DeserializeDesiredState(rapidjson::Document& document, DesiredState& object);
+    int ValidateAndGetPackagesNames(std::vector<std::string> packagesLines);
+    static std::vector<std::string> ListFiles(const char* dir, const char* fileNameExtension);
```
MariusNi, line 74: >dir [](http://example.com/codeflow?start=58&length=3)

"directory" - use full words

src/modules/pmc/tests/PmcTests.cpp
```diff
@@ -1,42 +1,298 @@
...
 #include <PmcBase.h>
 
-class PmcTests : public PmcBase
+class PmcTestImpl : public PmcBase
 {
 public:
-    PmcTests(unsigned int maxPayloadSizeBytes);
-    ~PmcTests() = default;
+    PmcTestImpl(unsigned int maxPayloadSizeBytes);
+    ~PmcTestImpl() = default;
```
MariusNi, line 15: Do you really need these?

MariusNi, line 15: > ~PmcTestImpl() = default; [](http://example.com/codeflow?start=3&length=26)

Can get rid of this

src/modules/pmc/tests/PmcTests.cpp
```diff
@@ -1,42 +1,298 @@
...
+    UNUSED(timeoutSeconds);
+
+    std::map<std::string, std::string>::const_iterator it = m_textResults.find(command);
+    if (it != m_textResults.end())
+    {
+        if (it->second == "COMMAND_ERROR")
+        {
+            return 100;
+        }
+        else if (it->second == "TIMEOUT_ERROR")
```
MariusNi, line 45: Do you convert numeric errors such as ETIME to string ("TIMEOUT_ERROR")? This is not efficient. Use normal numeric values for error codes. Referring to the code under test, not just the unittest

src/modules/pmc/src/lib/Pmc.h
Ioana37: For the `g_commandExecuteUpdate` we specify a timeout of 600 seconds. Could we keep it?

src/modules/pmc/src/lib/Pmc.cpp
Ioana37: I did remove the `replaceEol` but as mentioned in a previous comment, for one of our commands we need a timeout of 600 of seconds.

src/modules/pmc/src/lib/PmcBase.h
Ioana37: Applied the change, thank you! Could you consider, in the future, to move the helper methods for string handling in a common class? We have duplicated code with the HostName module.

src/modules/pmc/tests/PmcTests.cpp
Ioana37: Actually, we convert our mocked `TIMEOUT_ERROR` to `ETIME` in order to simulate an error of `RunCommand` method.

src/modules/pmc/src/lib/Pmc.h
MariusNi: Is it really necessary to have a timeout so long? This is 10 minutes. Default (when specified with a 0 value) is 60 seconds aka 1 minute.

If you need 10 minutes OK, but then recommend to hard-code that to some constant and remove this timeoutSeconds argument from this internal method.

src/modules/pmc/src/lib/Pmc.cpp
MariusNi: Hardcode that 600 (if you really need 10 minutes) to a constant and use that internally, skipping this argument.

src/modules/pmc/src/lib/Pmc.cpp
MariusNi: FREE_MEMORY does not need the check on line 27. Remove.

src/modules/pmc/tests/PmcTests.cpp
MariusNi: Why mock as a string instead of the plain ETIME?

src/agents/pnp/daemon/osconfig.json
```diff
@@ -59,6 +59,10 @@
       "ComponentName": "Networking",
       "ObjectName": "NetworkConfiguration"
     },
+    {
+      "ComponentName": "PackageManagerConfiguration",
+      "ObjectName": "State"
+    },
     {
```
MariusNi, line 66: You need to pull from main. This changed since this PR started. All object names are now camelCased instead PascalCased. Same for settings and settings values.

MariusNi (changes requested):
:clock1:

src/modules/pmc/tests/PmcTests.cpp
Ioana37: The goal of the test implementation `RunCommand` function is to mock the expected outcome without actually executing any commands. As such, we have a string-to-string map called `textResults` that specifies as a key the command to be mocked and as a value the expected output of the respective command. We followed the same approach as in the HostName module tests. One thing that is different in our tests, is the fact that we wanted to also simulate errors and that is why we used the hardcoded error strings. Since we are not executing any command, we cannot obtain a real error code, so we simulate the two categories - timeout and non-timeout related command errors. With these special strings (`TIMEOUT_ERROR`, `COMMAND_ERROR`), we were able to simulate both happy and unhappy executions of `RunCommand` using only our string-to-string map and with no additional changes.

To make it clearer, we removed the special strings mentioned above and enhanced the string to string map to a map of string to `tuple<int, string>` as in the example:
  ```
const std::map<std::string, std::tuple<int, std::string>> textResults =
        {
            {"command -v apt-get", std::tuple<int, std::string>(MMI_OK, "")},
            {"command -v apt-get", std::tuple<int, std::string>(ETIME, "")}
        }
```

Let us know if this approach looks better.

File: src/agents/pnp/daemon/osconfig.json [hunk 14]
```diff
@@ -59,6 +59,10 @@
       "ComponentName": "Networking",
       "ObjectName": "networkConfiguration"
     },
+    {
+      "ComponentName": "PackageManagerConfiguration",
+      "ObjectName": "State"
```
MariusNi, line 64: > "ObjectName": "State" [](http://example.com/codeflow?start=5&length=22)

This needs to be "state"

src/modules/pmc/src/lib/Pmc.cpp
```diff
@@ -1,9 +1,39 @@
...
+    char* buffer = nullptr;
+    int status;
+
+    if (isLongRunning)
+    {
+        status = ExecuteCommand(nullptr, command, true, true, 0, TIMEOUT_LONG_RUNNING, &buffer, nullptr, PmcLog::Get());
+    }
+    else
+    {
+        status = ExecuteCommand(nullptr, command, true, true, 0, 0, &buffer, nullptr, PmcLog::Get());
```
MariusNi, line 25: These can be combined in a single line:

 status = ExecuteCommand(nullptr, command, true, true, 0, isLongRunning : TIMEOUT_LONG_RUNNING : 0, &buffer, nullptr, PmcLog::Get());

MariusNi (approved):
:shipit:

robertschaedler3 (approved):

PR #149, opened 2022-03-31
Add logic for configuring sources in Pmc module

MagdaPaj:
## Description

* Add logic for configuring sources in the Pmc module
* Simplify ExecuteUpdates method. There is no need to RunCommand `g_commandAptUpdate` there as this code is done in the ConfigureSources method
* Update tests
* Update Readme with sources information and change names to camelCase

## Checklist


src/modules/pmc/src/lib/PmcBase.cpp
```diff
@@ -561,3 +607,130 @@ std::string PmcBase::Trim(const std::string& str, const std::string& trim)
 {
     return TrimStart(TrimEnd(str, trim), trim);
 }
+
+std::string PmcBase::GetSourcesFingerprint(const char* sourcesDirectory)
+{
+    std::string hashString = "";
+    std::string command = std::regex_replace(g_commandGetSourcesFingerprint, std::regex("\\$value"), sourcesDirectory);
+    int status = RunCommand(command.c_str(), &hashString);
```
MariusNi, line 615: > int status = RunCommand(command.c_str(), &hashString); [](http://example.com/codeflow?start=3&length=55)

Use the commonutils!HashString instead

src/modules/pmc/src/lib/PmcBase.cpp
```diff
@@ -561,3 +607,130 @@ std::string PmcBase::Trim(const std::string& str, const std::string& trim)
...
+        // Delete file when provided map value is empty
+        if (source.second.empty())
+        {
+            if (FileExists(sourcesFilePath.c_str()))
+            {
+                status = remove(sourcesFilePath.c_str());
+            }
+            else
+            {
+                OsConfigLogInfo(PmcLog::Get(), "Nothing to delete. Source(s) file: %s does not exist", sourcesFilePath.c_str());
```
MariusNi, line 693: > OsConfigLogInfo(PmcLog::Get(), "Nothing to delete. Source(s) file: %s does not exist", sourcesFilePath.c_str()); [](http://example.com/codeflow?start=15&length=113)

Don't log this. Or only log if full logging is enabled - but only if this is useful.

src/modules/pmc/src/lib/PmcBase.cpp
```diff
@@ -561,3 +607,130 @@ std::string PmcBase::Trim(const std::string& str, const std::string& trim)
...
+
+    if (status != MMI_OK)
+    {
+        OsConfigLogError(PmcLog::Get(), "Refresh package lists failed with status %d", status);
+        status == ETIME ? m_executionState.SetExecutionState(ExecutionState::StateComponent::timedOut, ExecutionState::SubStateComponent::updatingPackagesLists)
+                        : m_executionState.SetExecutionState(ExecutionState::StateComponent::failed, ExecutionState::SubStateComponent::updatingPackagesLists);
+    }
+    else
+    {
+        OsConfigLogInfo(PmcLog::Get(), "Successfully updated package lists");
```
MariusNi, line 731: > OsConfigLogInfo(PmcLog::Get(), "Successfully updated package lists"); [](http://example.com/codeflow?start=7&length=70)

Don't log this or log only if full logging is enabled

src/modules/pmc/tests/PmcTests.cpp
```diff
@@ -141,6 +159,36 @@ namespace OSConfig::Platform::Tests
...
+            {"dpkg-query --showformat='${Package} (=${Version})\n' --show | sha256sum | head -c 64", std::tuple<int, std::string>(MMI_OK, "25abefbfdb34fd48872dea4e2339f2a17e395196945c77a6c7098c203b87fca4")},
+            {"apt-cache policy cowsay | grep Installed", std::tuple<int, std::string>(MMI_OK, "  Installed: (none) ")},
+            {"apt-cache policy sl | grep Installed", std::tuple<int, std::string>(MMI_OK, "  Installed: (none) ")},
+            {"apt-cache policy bar | grep Installed", std::tuple<int, std::string>(MMI_OK, "  Installed: (none) ")},
+            {"find sources/ -type f -name '*.list' -exec cat {} \\; | sha256sum | head -c 64", std::tuple<int, std::string>(MMI_OK, "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b877")}
+        };
+        char reportedJsonPayload[] = "{\"packagesFingerprint\":\"25abefbfdb34fd48872dea4e2339f2a17e395196945c77a6c7098c203b87fca4\",\"packages\":[\"cowsay=(none)\",\"sl=(none)\",\"bar=(none)\"],\"executionState\":3,\"executionSubState\":6,\"executionSubStateDetails\":\"\",\"sourcesFingerprint\":\"e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b877\",\"sourcesFilenames\":[\"key.list\"]}";
```
MariusNi, line 176: Break this super long string constant in substrings

"                              ";

->

"  "
"  "
"  ";

src/modules/pmc/tests/PmcTests.cpp
```diff
@@ -153,9 +201,10 @@ namespace OSConfig::Platform::Tests
...
-            {"apt-cache policy bar | grep Installed", std::tuple<int, std::string>(MMI_OK, "  Installed: (none) ")}
+            {"apt-cache policy bar | grep Installed", std::tuple<int, std::string>(MMI_OK, "  Installed: (none) ")},
+            {"find sources/ -type f -name '*.list' -exec cat {} \\; | sha256sum | head -c 64", std::tuple<int, std::string>(MMI_OK,"e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b877")}
         };
-        char reportedJsonPayload[] = "{\"packagesFingerprint\":\"25abefbfdb34fd48872dea4e2339f2a17e395196945c77a6c7098c203b87fca4\",\"packages\":[\"cowsay=(none)\",\"sl=(none)\",\"bar=(none)\"],\"executionState\":4,\"executionSubState\":5,\"executionSubStateDetails\":\"cowsay=3.03+dfsg2-7:1 sl\"}";
+        char reportedJsonPayload[] = "{\"packagesFingerprint\":\"25abefbfdb34fd48872dea4e2339f2a17e395196945c77a6c7098c203b87fca4\",\"packages\":[\"cowsay=(none)\",\"sl=(none)\",\"bar=(none)\"],\"executionState\":4,\"executionSubState\":7,\"executionSubStateDetails\":\"cowsay=3.03+dfsg2-7:1 sl\",\"sourcesFingerprint\":\"e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b877\",\"sourcesFilenames\":[\"key.list\"]}";
```
MariusNi, line 207: Same as commented earlier. Break long string constant on multi line substrings. Here and elsewhere.

src/modules/pmc/src/lib/PmcBase.cpp
MagdaPaj: So we have one concern regarding using `commonutils!HashString` for getting the sources fingerprint. 
From [std::hash documentation](https://en.cppreference.com/w/cpp/utility/hash):

> Hash functions are only required to produce the same result for the same input within a single execution of a program; this allows salted hashes that prevent collision denial-of-service attacks.

So it seems that there is no guarantee that the sources fingerprint computed with the `std::hash` will remain the same  for the same input after OsConfig restart. And this problem doesn't happen when using `sha256sum`.  Also the same approach with `sha256sum` is used in the [Firewall](https://github.com/Azure/azure-osconfig/blob/main/src/modules/firewall/src/lib/Firewall.cpp#L14) module.  So wouldn't be better to keep using `sha256sum`?

src/modules/pmc/src/lib/PmcBase.cpp
MagdaPaj: I deleted some of the LogInfo lines and the rest of the LogInfo in that file is only used with full logging enabled.

src/modules/pmc/src/lib/PmcBase.cpp
MariusNi: Correct. If you indeed need a hash that persists from one process instance to another you need something like is done in FirewallObjectBase::GetFingerprint() in the Firewall module.

As now more than 1 module needs this we should consider adding this new function to commoutils. 

Try to make this local function completely detached from rest of the code, make it a global function even. Class member is fine as long as it does not access any class members. And also have unittests specific for it. Once that is done and this PR is merged, we can move the function and its unitests to commonutils.

src/modules/pmc/src/lib/PmcBase.cpp
MariusNi: Success traces can overflow the log repeating same trace again and again during periodic invocation of MmiGet and MmiSet. 

So in general try to not log anything informal from inside these two functions, MmiGet and MmiSet. 

Even full logging enabled, think if indeed that trace is useful, esp when it will repeat itself many times.

This would be a good occasion to review all informal traces written from inside these 2 MMI calls and their invoked internal helpers.

src/modules/pmc/src/lib/PmcBase.cpp
MariusNi: Another alternative would be while you review the use of informal traces during MmiGet and MmiSet, we could look into adding the new hash function to commonutils so you can use that one in this module. Doing that just now.

MariusNi: Merge from main and look for commonutils!PersistentHashString. Try to use this one and let us/me know if you encounter any problem.

MagdaPaj: Thanks for implementing the `commonutils!PersistentHashString`. I have 2 problems with it.

Problem 1.
For getting the sources fingerprint, I'm not able to get the same fingerprint as before.
So before I was executing this command: `find /etc/apt/sources.list.d/ -type f -name '*.list' -exec cat {} \; | sha256sum | head -c 64` and the returned fingerprint was: `f4f6cf81ab525b669be96f5d29f329600c4c2fb0b84b54ac38d39b8c1077d4c2`.

When using `commonutils!PersistentHashString` I need to break my command and first get sources content and then apply PersistentHashString function. The sources content I'm getting with this command: `find /etc/apt/sources.list.d/ -type f -name '*.list' -exec cat {} \;`. When running this command locally in bash it returns 3 lines output:
```
deb [arch=amd64,arm64,armhf] https://packages.microsoft.com/ubuntu/18.04/multiarch/prod testing main
deb [trusted=yes] https://mirrors.gigenet.com/mariadb/repo/10.5/ubuntu focal main
deb [trusted=yes] https://mirrors.gigenet.com/mariadb/repo/10.5/ubuntu focal main
```

But when running this command with `RunCommand` in the OsConfig it seems to return one line: 
```
deb [arch=amd64,arm64,armhf] https://packages.microsoft.com/ubuntu/18.04/multiarch/prod testing main deb [trusted=yes] https://mirrors.gigenet.com/mariadb/repo/10.5/ubuntu focal main deb [trusted=yes] https://mirrors.gigenet.com/mariadb/repo/10.5/ubuntu focal main
```

so then when `PersistentHashString` is executed it runs:

```
echo deb [arch=amd64,arm64,armhf] https://packages.microsoft.com/ubuntu/18.04/multiarch/prod testing main deb [trusted=yes] https://mirrors.gigenet.com/mariadb/repo/10.5/ubuntu focal main deb [trusted=yes] https://mirrors.gigenet.com/mariadb/repo/10.5/ubuntu focal main  | sha256sum | head -c 64
```
and returns `b15226514c42f33c64eaadb564cbb2b8499fd882aa0683ae0cf245f2ccd9be74`

But the correct command that should be executed is this:
```
echo -e 'deb [arch=amd64,arm64,armhf] https://packages.microsoft.com/ubuntu/18.04/multiarch/prod testing main\ndeb [trusted=yes] https://mirrors.gigenet.com/mariadb/repo/10.5/ubuntu focal main\ndeb [trusted=yes] https://mirrors.gigenet.com/mariadb/repo/10.5/ubuntu focal main' | sha256sum | head -c 64
```
because it returns the original fingerprint `f4f6cf81ab525b669be96f5d29f329600c4c2fb0b84b54ac38d39b8c1077d4c2`.

So there is a problem with multiline input: 
* the RunCommand makes multiple lines text result as a single line
* when applying `echo` for multiple lines it seems that `-e` parameter is required.

Problem 2:
We also need to get a fingerprint for packages. And for that we had: `dpkg-query --showformat='${Package} (=${Version})\n' --show  | sha256sum | head -c 64`. When I changed that, and first executed `dpkg-query --showformat='${Package} (=${Version})\n' --show` with `RunCommand` and then I applied `PersistentHashString`, then OsConfig crashed completely. The `echo` command with the string output from `dpkg-query` returned 2. And then in the Pnp agent I got this error:
```
[ERROR] OSConfig crash due to abnormal termination (SIGABRT) during MpiGet to PackageManagerConfiguration.state
````

The text result of `dpkg-query --showformat='${Package} (=${Version})\n' --show` is probably just too big. In my case it has 817 lines.

So when getting a fingerprint of huge text result, it seems more effective to do it in a single command `dpkg-query --showformat='${Package} (=${Version})\n' --show  | sha256sum | head -c 64`, instead of first getting very long string and then appling  `PersistentHashString`.

src/modules/pmc/src/lib/PmcBase.cpp
MagdaPaj: I did some more cleanup and deleted LogInfo messages that weren't really necessary.

src/modules/pmc/src/lib/PmcBase.cpp
MariusNi: #1 commonutils!ExecuteCommand has the argument replaceEol which when is true, the end of line characters '\n' are replaced with spaces, thus making multiple lines appear as one single line. Use the correct value for this replaceEol argument when executing the command, to allow you receive the data in the format (multiple or single line) that your module needs.

#2 Please paste here or send e-mail with the exact text that you pass into commonutils!PersistentHashString to make it crash. I will try to repro the same. Do not need the command that produces that text, need just the text that is passed into hashing function.

MagdaPaj:
Closing as the approach in the #169 is preferred.

PR #170, opened 2022-04-05
Repurposing commonutills!PersistentHahsString to commonutills!HashCommand

MariusNi:


src/common/tests/CommonUtilsUT.cpp
```diff
@@ -1002,170 +1002,25 @@ TEST_F(CommonUtilsTest, DuplicateString)
...
 
-    EXPECT_NE(nullptr, hashOne = PersistentHashString(testOne, nullptr));
-    EXPECT_NE(nullptr, hashTwo = PersistentHashString(testTwo, nullptr));
-    EXPECT_NE(nullptr, hashThree = PersistentHashString(testOne, nullptr));
+    EXPECT_NE(nullptr, hashOne = HashCommand(testOne, nullptr));
+    EXPECT_NE(nullptr, hashTwo = HashCommand(testTwo, nullptr));
+    EXPECT_NE(nullptr, hashThree = HashCommand(testOne, nullptr));
 
     EXPECT_NE(hashOne, hashTwo);
     EXPECT_STREQ(hashOne, hashThree);
```
robertschaedler3, line 1021: Since `sha256sum` will give the same output each time for a given string, would it be possible to also check the hash returned by `HashCommand()` here?

robertschaedler3 (approved):

src/common/tests/CommonUtilsUT.cpp
MariusNi: Are we 100% sure the sha356sum will always give the same output for same input on all platforms where this test will be run on? We've been burned before assuming constant responses.

PR #176, opened 2022-04-13
Add GPG functionality

Salazander:
## Description

- Added GPG functionality (deserialize, download and link with package source entry, see README.md for details)
- Added GPG unit tests
- Use PascalCase in code
- Updated README with GPG instructions
- Use custom PMC_0K variable as **internal** status code for successful operations
- Pass all classes by reference and as const if possible
- Split deserialization to improve readability
- Override CanRunOnThisPlatform in PmcTestImpl to simplify unit testing

## Checklist


src/modules/pmc/src/lib/PmcBase.cpp
```diff
@@ -18,20 +18,26 @@ static const std::string g_reportedObjectName = "state";
...
 static const std::string g_packagesFingerprint = "packagesFingerprint";
 static const std::string g_sourcesFingerprint = "sourcesFingerprint";
 static const std::string g_sourcesFilenames = "sourcesFilenames";
-static const std::string g_requiredTools[] = {"apt-get", "apt-cache", "dpkg-query"};
 
-constexpr const char* g_commandCheckToolPresence = "command -v $value";
 constexpr const char* g_commandAptUpdate = "apt-get update";
 constexpr const char* g_commandExecuteUpdate = "apt-get install $value -y --allow-downgrades --auto-remove";
 constexpr const char* g_commandGetInstalledPackageVersion = "apt-cache policy $value | grep Installed";
+constexpr const char* g_commandDownloadGpgKey = "bash -c 'set -o pipefail && curl $url | gpg --dearmor | tee $destination'";
```
AhmedBM, line 32: would `sh -c` be better here? Not all distros have bash.

src/modules/pmc/src/lib/PmcBase.cpp
```diff
@@ -684,29 +752,130 @@ int PmcBase::ConfigureSources(const std::map<std::string, std::string> sources)
...
+}
+
+bool PmcBase::ValidateAndUpdatePackageSource(std::string& packageSource, const std::map<std::string, std::string>& gpgKeys)
+{
+    std::smatch sourceMatches;
+
+    if (std::regex_match(packageSource, sourceMatches, std::regex(g_regexSources)))
+    {
+        std::smatch signedByMatches;
+        std::string matchedString = sourceMatches[2].matched ? sourceMatches[2].str() : std::string();
```
AhmedBM, line 801: Are these indices always valid on partial matches? Would it be safer to validate `sourceMatches.size() >= 2`?

src/modules/pmc/src/lib/PmcBase.cpp
```diff
@@ -684,29 +752,130 @@ int PmcBase::ConfigureSources(const std::map<std::string, std::string> sources)
...
+        if (std::regex_match(matchedString, signedByMatches, std::regex(g_regexSignedByOption)))
+        {
+            if (signedByMatches[1].matched)
```
AhmedBM, line 804: Similar to above, is every always at least 1 indices? Would it be safer to validate `signedByMatches.size() > 0`?

AhmedBM (commented):
Looks good just have some concerns about those indices I mentioned.

src/modules/pmc/src/lib/ExecutionState.cpp
```diff
@@ -23,22 +23,22 @@ void ExecutionState::SetExecutionState(StateComponent stateComponent, SubStateCo
     SetExecutionState(stateComponent, subStateComponent, "");
 }
 
-bool ExecutionState::IsSuccessful()
+bool ExecutionState::IsSuccessful() const
 {
-    return m_stateComponent == StateComponent::succeeded;
+    return m_stateComponent != StateComponent::Failed && m_stateComponent != StateComponent::TimedOut;
```
MariusNi, line 28: >m_stateComponent != StateComponent::Failed && m_stateComponent != StateComponent::TimedOut; [](http://example.com/codeflow?start=11&length=91)

Add parenthesis

src/modules/pmc/src/lib/ExecutionState.h
```diff
@@ -6,34 +6,36 @@ class ExecutionState
...
     void SetExecutionState(StateComponent stateComponent, SubStateComponent subStateComponent, std::string processingArgument);
     void SetExecutionState(StateComponent stateComponent, SubStateComponent subStateComponent);
-    bool IsSuccessful();
-    StateComponent GetExecutionState();
-    SubStateComponent GetExecutionSubState();
-    std::string GetExecutionSubStateDetails();
+    bool IsSuccessful() const;
+    StateComponent GetExecutionState() const;
+    SubStateComponent GetExecutionSubState() const;
+    std::string GetExecutionSubStateDetails() const;
```
MariusNi, line 38: substate is a noun. Everywhere we can have "Substate" or "substate" instead of "SubState" and respectively "subState".

src/modules/pmc/src/lib/Pmc.cpp
```diff
@@ -18,9 +21,9 @@ Pmc::Pmc(unsigned int maxPayloadSizeBytes)
 int Pmc::RunCommand(const char* command, std::string* textResult, bool isLongRunning)
 {
     char* buffer = nullptr;
-    int status = ExecuteCommand(nullptr, command, true, true, 0, isLongRunning ? TIMEOUT_LONG_RUNNING : 0, &buffer, nullptr, PmcLog::Get()); 
+    int status = ExecuteCommand(nullptr, command, true, true, 0, isLongRunning ? TIMEOUT_LONG_RUNNING : 0, &buffer, nullptr, PmcLog::Get());
```
MariusNi, line 24: >true, [](http://example.com/codeflow?start=56&length=5)

Why use true for the forJson argument? It replaces characters " and \ to spaces. Is this intended?

src/modules/pmc/src/lib/PmcBase.cpp
```diff
@@ -118,7 +127,7 @@ int PmcBase::GetInfo(const char* clientName, MMI_JSON_STRING* payload, int* payl
 
 int PmcBase::Set(const char* componentName, const char* objectName, const MMI_JSON_STRING payload, const int payloadSizeBytes)
 {
-    if(!CanRunOnThisPlatform())
+    if (!CanRunOnThisPlatform())
     {
         return ENODEV;
```
MariusNi, line 132: >return ENODEV; [](http://example.com/codeflow?start=8&length=14)

This needs an error trace to tell the operator why this module refuses to run (here and elsewhere)

MariusNi: Found it in the CanRunOnThisPlatform body. Looks good

src/modules/pmc/src/lib/PmcBase.cpp
Salazander: the array size is tightly coupled to the number of capture groups present in the regex ([details here](https://www.geeksforgeeks.org/smatch-regex-regular-expressions-in-c/)).

But since you're the second reviewer with the same concerns 😉, I'll add an explicit check.

src/modules/pmc/src/lib/PmcBase.cpp
Salazander: done!

src/modules/pmc/src/lib/PmcBase.cpp
Salazander: done!

src/modules/pmc/src/lib/ExecutionState.cpp
Salazander: done!

src/modules/pmc/src/lib/ExecutionState.h
Salazander: done!

src/modules/pmc/src/lib/Pmc.cpp
Salazander: good catch, changed to false.

src/modules/pmc/src/lib/PmcBase.cpp
Salazander: I refactored the command to avoid the bash dependency while at the same time making sure that the piped commands fail when the GPG key cannot be downloaded or is not in the expected format.

MariusNi (approved):
:shipit:

AhmedBM (approved):
👌

PR #179, opened 2022-04-14
Create build-containers workflow

robertschaedler3:
## Description

Create GitHub Actions workflow to build and push devops containers to the OSConfig Azure CR. This workflow will be triggered when changes to `devops/docker/**/Dockerfile` are merged to main.

## Checklist


.github/workflows/build-containers.yml
```diff
@@ -0,0 +1,51 @@
...
+    outputs:
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
+    steps:
+      - uses: actions/checkout@v2
+
+      - id: set-matrix
+        working-directory: ./devops/docker
+        run: echo "::set-output name=matrix::$(ls **/Dockerfile | jq -R -s -c 'split("\n")[:-1]')"
+
+  docker:
```
AhmedBM, line 18: Is there the ability to target specific distros? It looks like all containers are built when we might be onboarding a new distro, without us needing to rebuild all of the environments. Consider adding options so we can selectively build any container. Similar to what we currently have in Azure Pipelines
image

.github/workflows/build-containers.yml
```diff
@@ -0,0 +1,51 @@
...
+        with:
+          username: ${{ secrets.ACR_CLIENT_ID }}
+          password: ${{ secrets.ACR_CLIENT_SECRET }}
+
+      - name: Build and push
+        uses: docker/build-push-action@v2
+        with:
+          context: ./devops/docker/${{ matrix.container }}
+          push: true
+          tags: ${{ steps.image.outputs.name }}
```
AhmedBM, line 51: Can we also add the `github.run_number` as a tag in addition to `latest` so we can version our containers.
https://docs.github.com/en/actions/learn-github-actions/contexts#github-context

AhmedBM (changes requested):
We must be able to selectively build containers and also address versioning of the build containers

.github/workflows/build-containers.yml
robertschaedler3: Updated to only build the containers containing changes merged to `main`. Any `Dockerfile` in `./devops/docker/` will be built and push if it is added, copied, modified, or renamed by PR that was merged to main.

.github/workflows/build-containers.yml
robertschaedler3: Good call!👍🏻

.github/workflows/build-containers.yml
```diff
@@ -11,9 +14,14 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
+      - name: Get changed files
```
AhmedBM, line 17: Does this give you just the changed Dockerfiles?

AhmedBM (commented):
Nice, so new changes to only the changed `Dockerfiles` are built? Would we still be allowed to re-trigger manually if needed?

.github/workflows/build-containers.yml
robertschaedler3: We can add a manual trigger, but I recommend we keep this workflow as being only triggered by added/modified Dockerfiles to require all builds of our container images to go through a PR process.

AhmedBM: i would expect this validation to be done locally first to just validate the Dockerfile. The problem however is what if we want to simply update the current `latest` packages. I think a better approach would be to use these triggers but update all our Dockerfiles to contain the explicit versions, this way we can truly have a git operation model.
This is what i meant by locking down our versions int the environments as we currently just use the latest available for most packages. If we adopt the git ops model, then every update will be clear about what the environments contain.

AhmedBM (changes requested):
Let's wait until we get the trusted runners onboarded before merging this one in

PR #228, opened 2022-05-17
Platform and MPI server bug fixes

robertschaedler3:


src/platform/inc/MpiServer.h
```diff
@@ -12,6 +12,9 @@ extern "C"
 void MpiApiInitialize(void);
 void MpiApiShutdown(void);
 
+int LoadModules(void);
+void UnloadModules(void);
```
MariusNi, line 16: These are not right in here. They should be in a MM header.

MariusNi: We'll fix this in a separate PR

PR #249, opened 2022-05-26
Fix TPM module default reporting values

robertschaedler3:
## Description

- Fix TPM module to report default values (empty JSON string) when a TPM does not exist or can't be read
- Simplified to only query the TPM once instead of on every `MmiGet` call 
- Query TPM once during 1st MmiGet call for all properties instead of once per property
- Additional cleanup/error checking

## Checklist


MariusNi (approved):
:shipit:

PR #256, opened 2022-06-07
Generate CI test report check runs

robertschaedler3:
## Description

- Update CMake version in containers (3.21.7)
- Add `jq` to container images
- Add XML output flag to `gtest_discover_tests`
- Create gtest-xml workflow for generating JUnit XML from gtest output

## Checklist


MariusNi (dismissed):
:shipit:

AhmedBM (commented):
Looks good but can you also include the `gtest_discover_tests` changes for the `modulestest` target?

AhmedBM (commented):
once the build containers are built, we should also change the `/azure-osconfig/.github/actions/container-run/action.yml` to lockdown the container tag and not just always use the "latest"

src/modules/test/CMakeLists.txt
```diff
@@ -39,6 +39,7 @@ endif()
 
 gtest_discover_tests(modulestest
     EXTRA_ARGS ${CMAKE_CURRENT_BINARY_DIR}/testplate.json
+    XML_OUTPUT_DIR ${GTEST_OUTPUT_DIR}
```
AhmedBM, line 42: why isn't the `TestRecipes` test suite not getting picked up in CI. We clearly see the tests being run?

MariusNi:
That can be done with a separate PR in the future, right? Why block now?

---
In reply to: [998620640](https://github.com/Azure/azure-osconfig/pull/256#issuecomment-998620640) [](http://example.com/codeflow?ancestors=998620640,998620640)

src/modules/test/CMakeLists.txt
robertschaedler3: I looked at the raw output of those tests and they do not seem to be registered correctly with gtest. For example:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<testsuites tests="0" failures="0" disabled="0" errors="0" time="0" timestamp="2022-06-07T20:16:24" name="AllTests">
</testsuites>
```

is missing the `<testsuite>...</testsuite>` information and `tests="0"`.

robertschaedler3: Once the tests are registered there and outputting the complete XML results, they will automatically show in the reports.

robertschaedler3: Let's address this issue when officially integrating the functional tests into our CI.

PR #269, opened 2022-06-16
Added binary baseline validations

AhmedBM:
## Description

* Adds binary baseline validation to `Package` workflow, using `1.0.3.2022061601` as current baseline.
* Binary baselines published in `devops/binary-footprint-baseline/baseline.json`
* Added script to update binary baselines (using latest published as default or by using an optional regex filter as first argument)
* Using current delta as 10%, meaning package workflows will fail if they exceed/drop the current footprint by more than 10%
* Fixed package build to use correct `CMAKE_BUILD_TYPE` argument

## Checklist

### DIFF
--- a/devops/binary-footprint-baseline/baseline.json
+++ b/devops/binary-footprint-baseline/baseline.json
@@ -1,0 +1,56 @@
+[
+  {
+    "distro": "ubuntu-18.04",
+    "arch": "x86_64",
+    "sizeBytes": 1772612,
+    "file": "osconfig_1.0.3.2022061601_bionic_x86_64.deb"
+  },
+  {
+    "distro": "ubuntu-18.04",
+    "arch": "aarch64",
+    "sizeBytes": 1711064,
+    "file": "osconfig_1.0.3.2022061601_bionic_aarch64.deb"
+  },
+  {
+    "distro": "ubuntu-18.04",
+    "arch": "armv7l",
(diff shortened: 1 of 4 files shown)


devops/binary-footprint-baseline/baseline.json
```diff
@@ -0,0 +1,56 @@
+[
+  {
+    "distro": "ubuntu-18.04",
+    "arch": "x86_64",
+    "sizeBytes": 1772612,
```
robertschaedler3, line 5: Are the `sizeBytes` here the expected value based on our current package size?

robertschaedler3, line 5: Does this mean we will need to update our baselines over time as osconfig grows so that CI will pass?

robertschaedler3 (approved):

devops/binary-footprint-baseline/baseline.json
AhmedBM: no, these are just the current sizes. We store them as snapshots and we compare the produced local debian package, to the ones contained in the baseline

AhmedBM: yes, thats correct. We will assess to update the baselines then, its a sort of gate we will use to ensure no builds are produced that have too much of a size delta to be flagged. Its at that point we can update the baselines in order to get the gates passing again

PR #292, opened 2022-07-14
Move firewall module code from shared object file to static lib

robertschaedler3:
## Description

Move firewall module code from shared object file to static lib to make the firewall module more in line with the sample module.

## Checklist

### DIFF
--- a/src/modules/firewall/src/lib/Firewall.cpp
+++ b/src/modules/firewall/src/lib/Firewall.cpp
@@ -7,88 +7,119 @@
 #include <bits/stdc++.h>
 #include "Firewall.h"
 
-OSCONFIG_LOG_HANDLE FirewallLog::m_logFirewall = nullptr;
-const string g_iptablesUtility = "iptables";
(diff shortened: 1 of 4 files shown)


src/modules/firewall/src/lib/Firewall.cpp
```diff
@@ -7,88 +7,119 @@
...
     {
-        string fingerprint = GetFingerprint();
-        payloadString = CreateFingerprintPayload(fingerprint);
+        OsConfigLogError(FirewallLog::Get(), "Invalid object name");
+        status = EINVAL;
     }
-    else
+    else if (nullptr == payload)
     {
+        OsConfigLogError(FirewallLog::Get(), "Invalid payload");
```
MariusNi, line 53: >Invalid payload [](http://example.com/codeflow?start=46&length=15)

Add reason (null) here and on line 48

robertschaedler3: Added there and elsewhere

MariusNi (approved):
:shipit:

PR #294, opened 2022-07-15
Increasing compilation strictness for C and C++ to maximum allowed by current limits

MariusNi:


src/agents/pnp/CMakeLists.txt
```diff
@@ -44,7 +44,7 @@ else ()
 endif ()
 
 if (EXISTS ${PROJECT_SOURCE_DIR}/azure-iot-sdk-c/CMakeLists.txt)
-    target_compile_options(${target_name} PRIVATE -Wall -Wextra -Werror -Wformat-security)
+    target_compile_options(${target_name} PRIVATE -Wall -Wextra -Werror -Wformat-security -Wunused-variable -Wunused-local-typedefs -Wunused-macros -Wunused-parameter -Wunused-but-set-parameter -Wunused-value)
```
AhmedBM, line 47: These are the C projects, can we simply use the root CMakeLists and add those flags there?

eg. `add_compile_options("$<$<COMPILE_LANGUAGE:C>:-Wunused-variable;-Wunused-local-typedefs;-Wunused-macros;-Wunused-parameter;-Wunused-but-set-parameter;-Wunused-value>")`

MariusNi: We could use as baseline default for C lowest set which is this one.

MariusNi: Actually, we cannot set any default for all C code because of the SDK. The SDK cannot do -Wall and -Wextra. We can do those here because is for agent code that touches external surface of the SDK which also imposes limits. But we cannot do an absolute default for all C in the entire project because that also applies to all internal SDK code.

AhmedBM (approved):
👌

PR #309, opened 2022-07-21
Tpm module test recipe

hachotze:
## Description

Extended module test recipe for Tpm. Added cases to validate the result code for each reported object in the Tpm MIM, and cases for MMI argument validation.

## Checklist

### DIFF
--- a/src/modules/test/recipes/TpmTests.json
+++ b/src/modules/test/recipes/TpmTests.json
@@ -1,44 +1,58 @@
 [
-    {
-        "ComponentName": "Tpm",
-        "ObjectName": "tpmStatus",
-        "Desired": false,
-        "ExpectedResult": 0,
-        "WaitSeconds": 0
-    },
-    {
-        "ComponentName": "Tpm",
-        "ObjectName": "tpmVersion",
-        "Desired": false,
-        "ExpectedResult": 0,
-        "WaitSeconds": 0
-    },
-    {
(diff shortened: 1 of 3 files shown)


src/modules/test/recipes/TpmTests.json
```diff
@@ -5,5 +5,40 @@
         "Desired": false,
         "ExpectedResult": 0,
         "WaitSeconds": 0
+    },
+    {
+        "ComponentName": "Tpm",
```
MariusNi, line 10: Please use 2 spaces per tab for JSON. Not 4. See how the MIM JSON is for example.

src/modules/test/recipes/TpmTests.json
```diff
@@ -5,5 +5,40 @@
...
+        "ObjectName": "",
+        "Desired": false,
+        "ExpectedResult": 22,
+        "WaitSeconds": 0
+    },
+    {
+        "ComponentName": "",
```
MariusNi, line 31: >"ComponentName": "", [](http://example.com/codeflow?start=8&length=20)

Rather than empty strings, test for names different than the module implements

AhmedBM: i think having both included is a good idea, along with some bogus names

MariusNi (approved):
:shipit:

PR #314, opened 2022-07-26
Firewall v3 MIM

robertschaedler3:
## Description

Draft of the Firewall v3 MIM for creating, deleting, and listing firewall rules using the following schema to represent a rule:

```
{ 
    "action": 0,                        // none = 0, addRule = 1, deleteRule = 2                  
    "ruleAction": 0,                    // none= 0, allowTraffic = 1, denyTraffic= 2
    "inbound": true,                    // outbound = false
    "protocol": 0,                      // all = 0, tcp = 1, udp = 2, icmp = 3
    "sourceAddress": "<ip_address>",    // ex: "20.43.132.131/24"
    "sourcePort": <port_number>,        // ex: 1234
    "destinationAddress": "<ip_address>",
    "destinationPort": <port_number>
}
```

## Checklist

### DIFF
--- a/src/modules/mim/firewall.json
+++ b/src/modules/mim/firewall.json
@@ -1,41 +1,187 @@
-{ 
-  "name": "FirewallModel", 
+{
+  "name": "FirewallModel",
   "type": "MimModel",
-  "contents": [ 
-    { 
-      "name": "Firewall", 
-      "type": "mimComponent", 
+  "contents": [
+    {
+      "name": "Firewall",
+      "type": "mimComponent",
       "contents": [
-        { 
-          "name": "firewallState", 
-          "type": "mimObject", 
+        {
+          "name": "firewallState",
+          "type": "mimObject",
           "desired": false,
           "schema": {
-             "type": "enum",
-             "valueSchema": "integer",
-             "enumValues": [
-               { 
-                 "name": "unknown",
-                 "enumValue": 0
-               },
-               {
-                 "name": "enabled",
-                 "enumValue": 1
-               },
-               {
-                 "name": "disabled",
-                 "enumValue": 2
-               }
-             ]
-           }
+            "type": "enum",
+            "valueSchema": "integer",
+            "enumValues": [
+              {
+                "name": "unknown",
+                "enumValue": 0
+              },
+              {
+                "name": "enabled",
+                "enumValue": 1
+              },
+              {
+                "name": "disabled",
+                "enumValue": 2
+              }
+            ]
+          }
         },
-        { 
-          "name": "firewallFingerprint", 
-          "type": "mimObject", 
+        {
+          "name": "firewallFingerprint",
+          "type": "mimObject",
           "desired": false,
           "schema": "string"
+        },
+        {
+          "name": "createRule",
+          "type": "mimObject",
+          "desired": true,
+          "schema": {
+            "type": "object",
+            "fields": [
+              {
+                "name": "action",
+                "schema": {
+                  "type": "enum",
+                  "valueSchema": "integer",
+                  "enumValues": [
+                    {
+                      "name": "accept",
+                      "enumValue": 0
+                    },
+                    {
+                      "name": "drop",
+                      "enumValue": 1
+                    },
+                    {
+                      "name": "reject",
+                      "enumValue": 2
+                    }
+                  ]
+                }
+              },
+              {
+                "name": "ruleType",
+                "schema": {
+                  "type": "enum",
+                  "valueSchema": "integer",
+                  "enumValues": [
+                    {
+                      "name": "inbound",
+                      "enumValue": 0
+                    },
+                    {
+                      "name": "outbound",
+                      "enumValue": 1
+                    },
+                    {
+                      "name": "forward",
+                      "enumValue": 2
+                    }
+                  ]
+                }
+              },
+              {
+                "name": "protocol",
+                "schema": {
+                  "type": "enum",
+                  "valueSchema": "integer",
+                  "enumValues": [
+                    {
+                      "name": "all",
+                      "enumValue": 0
+                    },
+                    {
+                      "name": "tcp",
+                      "enumValue": 1
+                    },
+                    {
+                      "name": "udp",
+                      "enumValue": 2
+                    },
+                    {
+                      "name": "icmp",
+                      "enumValue": 3
+                    }
+                  ]
+                }
+              },
+              {
+                "name": "source",
+                "schema": {
+                  "type": "object",
+                  "fields": [
+                    {
+                      "name": "address",
+                      "schema": "string"
+                    },
+                    {
+                      "name": "port",
+                      "schema": "integer"
+                    },
+                    {
+                      "name": "interface",
+                      "schema": "string"
+                    }
+                  ]
+                }
+              },
+              {
+                "name": "destination",
+                "schema": {
+                  "type": "object",
+                  "fields": [
+                    {
+                      "name": "address",
+                      "schema": "string"
+                    },
+                    {
+                      "name": "port",
+                      "schema": "integer"
+                    },
+                    {
+                      "name": "interface",
+                      "schema": "string"
+                    }
+                  ]
+                }
+              },
+              {
+                "name": "state",
+                "schema": {
+                  "type": "array",
+                  "elementSchema": {
+                    "type": "enum",
+                    "valueSchema": "string"
+                  }
+                }
+              }
+            ]
+          }
+        },
+        {
+          "name": "deleteRule",
+          "type": "mimObject",
+          "desired": true,
+          "schema": {
+            // ...
+          }
+        },
+        {
+          "name": "rules",
+          "type": "mimObject",
+          "desired": false,
+          "schema": {
+            "type": "array",
+            "elementSchema": {
+              // ...
+            }
+          }
         }
       ]
     }
   ]
-} 
+}
--- a/src/modules/mim/firewall.json
+++ b/src/modules/mim/firewall.json
@@ -1,6 +1,6 @@
 {
   "name": "FirewallModel",
-  "type": "MimModel",
+  "type": "mimModel",
   "contents": [
     {
       "name": "Firewall",
@@ -36,7 +36,7 @@
           "schema": "string"
         },
         {
-          "name": "createRule",
+          "name": "firewallRule",
           "type": "mimObject",
           "desired": true,
           "schema": {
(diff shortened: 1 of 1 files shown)


src/modules/mim/firewall.json
```diff
@@ -1,41 +1,187 @@
...
+            "fields": [
+              {
+                "name": "action",
+                "schema": {
+                  "type": "enum",
+                  "valueSchema": "integer",
+                  "enumValues": [
+                    {
+                      "name": "accept",
+                      "enumValue": 0
```
MariusNi, line 53: None shall be at 0

src/modules/mim/firewall.json
```diff
@@ -1,41 +1,187 @@
...
+              },
+              {
+                "name": "ruleType",
+                "schema": {
+                  "type": "enum",
+                  "valueSchema": "integer",
+                  "enumValues": [
+                    {
+                      "name": "inbound",
+                      "enumValue": 0
```
MariusNi, line 74: None for 0

src/modules/mim/firewall.json
```diff
@@ -1,41 +1,187 @@
...
+              },
+              {
+                "name": "protocol",
+                "schema": {
+                  "type": "enum",
+                  "valueSchema": "integer",
+                  "enumValues": [
+                    {
+                      "name": "all",
+                      "enumValue": 0
```
MariusNi, line 95: None for 0

src/modules/mim/firewall.json
```diff
@@ -1,41 +1,187 @@
...
+                      "enumValue": 3
+                    }
+                  ]
+                }
+              },
+              {
+                "name": "source",
+                "schema": {
+                  "type": "object",
+                  "fields": [
```
MariusNi, line 116: Objects within objects is not supported

MariusNi (changes requested):
:clock1:

src/modules/mim/firewall.json
```diff
@@ -1,41 +1,187 @@
...
+                }
+              },
+              {
+                "name": "state",
+                "schema": {
+                  "type": "array",
+                  "elementSchema": {
+                    "type": "enum",
+                    "valueSchema": "string"
+                  }
```
MariusNi, line 159: Array of enumeration if strings not supported also. Array or strings maybe?

src/modules/mim/firewall.json
```diff
@@ -1,41 +1,187 @@
...
-          "name": "firewallFingerprint", 
-          "type": "mimObject", 
+        {
+          "name": "firewallFingerprint",
+          "type": "mimObject",
           "desired": false,
           "schema": "string"
+        },
+        {
+          "name": "createRule",
```
MariusNi, line 39: >createRule [](http://example.com/codeflow?start=19&length=10)

Not right for an object name. This name is for an action, an object name should be for an object.

MariusNi: desiredFirewallRule?

src/modules/mim/firewall.json
```diff
@@ -1,41 +1,187 @@
...
+                    "type": "enum",
+                    "valueSchema": "string"
+                  }
+                }
+              }
+            ]
+          }
+        },
+        {
+          "name": "deleteRule",
```
MariusNi, line 166: >deleteRule [](http://example.com/codeflow?start=19&length=10)

Rather than a separate object, have an action Add/Remove on the desiredFirewallRule

src/modules/mim/firewall.json
```diff
@@ -1,41 +1,187 @@
...
+          "type": "mimObject",
+          "desired": true,
+          "schema": {
+            // ...
+          }
+        },
+        {
+          "name": "rules",
```
MariusNi, line 174: >rules [](http://example.com/codeflow?start=19&length=5)

firewallRules

src/modules/mim/firewall.json
MariusNi: Add "desired" to name to tell that this is not a reported object.

src/modules/mim/firewall.json
```diff
@@ -1,41 +1,305 @@
...
+                    {
+                      "name": "none",
+                      "enumValue": 0
+                    },
+                    {
+                      "name": "create",
+                      "enumValue": 1
+                    },
+                    {
+                      "name": "delete",
```
MariusNi, line 60: >delete [](http://example.com/codeflow?start=31&length=6)

Opposite to create is destroy and to delete, add. Choose one

src/modules/mim/firewall.json
```diff
@@ -1,41 +1,305 @@
...
+                "name": "destinationInterface",
+                "schema": "string"
+              },
+              {
+                "name": "state",
+                "schema": {
+                  "type": "array",
+                  "elementSchema": "string"
+                }
+              }
```
MariusNi, line 175: Do we really need this?

src/modules/mim/firewall.json
```diff
@@ -1,41 +1,305 @@
...
+                "schema": "integer"
+              },
+              {
+                "name": "destination",
+                "schema": "string"
+              },
+              {
+                "name": "destinationPort",
+                "schema": "integer"
+              },
```
MariusNi, line 160: Why not just one "port", either destination or source depending on rule direction?

src/modules/mim/firewall.json
```diff
@@ -1,41 +1,305 @@
...
+                      "enumValue": 1
+                    },
+                    {
+                      "name": "reject",
+                      "enumValue": 2
+                    },
+                    {
+                      "name": "drop",
+                      "enumValue": 3
+                    }
```
MariusNi, line 87: What's the diff between reject and drop?

src/modules/mim/firewall.json
robertschaedler3: This is used to pass symbols to the iptables state module (`established`, `related` etc...). We can omit this, but it will limit the types of rules that can be set.

src/modules/mim/firewall.json
robertschaedler3: This is because the port is not known just by the rule direction. For example, rules to allow all incoming SSH traffic will look like this:
```
sudo iptables -A INPUT -p tcp --dport 22 -m conntrack --ctstate NEW,ESTABLISHED -j ACCEPT
sudo iptables -A OUTPUT -p tcp --sport 22 -m conntrack --ctstate ESTABLISHED -j ACCEPT
```
however, a rule to allow outgoing SSH traffic will look like this:
```
sudo iptables -A OUTPUT -p tcp --dport 22 -m conntrack --ctstate NEW,ESTABLISHED -j ACCEPT
sudo iptables -A INPUT -p tcp --sport 22 -m conntrack --ctstate ESTABLISHED -j ACCEPT
```

In the first example incoming traffic (`INPUT`) uses the destination port (`-dport`) while and outgoing traffic (`OUTPUT`) uses the source port (`--sport`). But in the second example it is reversed (outgoing -> destination port, incoming -> source port).

robertschaedler3: This also shows the uses for `state` (`NEW`, `ESTABLISHED`)

src/modules/mim/firewall.json
robertschaedler3: Reject and drop essentially do the same thing: ending the rule traversal and "block" that traffic. The difference is if a packet is `dropped` no response is given to the sender, but when it is `rejected` an error packet is sent in response to the matched packet.

src/modules/mim/firewall.json
```diff
@@ -1,41 +1,305 @@
...
+                    {
+                      "name": "outbound",
+                      "enumValue": 2
+                    },
+                    {
+                      "name": "forward",
+                      "enumValue": 3
+                    }
+                  ]
+                }
```
MariusNi, line 114: Can replace this enum with a boolean "outbound" (true out, false in, FW does not really matter)

src/modules/mim/firewall.json
```diff
@@ -1,41 +1,305 @@
...
+                    {
+                      "name": "delete",
+                      "enumValue": 2
+                    }
+                  ]
+                }
+              },
+              {
+                "name": "command",
+                "schema": {
```
MariusNi, line 68: Rename command to action

src/modules/mim/firewall.json
```diff
@@ -1,41 +1,305 @@
...
+                  "type": "enum",
+                  "valueSchema": "integer",
+                  "enumValues": [
+                    {
+                      "name": "none",
+                      "enumValue": 0
+                    },
+                    {
+                      "name": "add",
```
MariusNi, line 56: >"add", [](http://example.com/codeflow?start=30&length=6)

addRule

src/modules/mim/firewall.json
```diff
@@ -1,41 +1,305 @@
...
+                      "enumValue": 1
+                    },
+                    {
+                      "name": "delete",
```
MariusNi, line 60: >delete", [](http://example.com/codeflow?start=31&length=8)

deleteRule

src/modules/mim/firewall.json
```diff
@@ -1,41 +1,305 @@
...
+                "schema": {
+                  "type": "enum",
+                  "valueSchema": "integer",
+                  "enumValues": [
+                    {
+                      "name": "none",
+                      "enumValue": 0
+                    },
+                    {
+                      "name": "all",
```
MariusNi, line 127: > "all" [](http://example.com/codeflow?start=29&length=6)

"any"

src/modules/mim/firewall.json
```diff
@@ -1,41 +1,305 @@
...
+                "schema": "integer"
+              },
+              {
+                "name": "destination",
+                "schema": "string"
+              },
+              {
+                "name": "destinationPort",
+                "schema": "integer"
+              },
```
MariusNi, line 160: source and destination can be combined in a single foo and port, direction being indicated by the other settings.

src/modules/mim/firewall.json
```diff
@@ -1,41 +1,305 @@
...
+                {
+                  "name": "state",
+                  "schema": {
+                    "type": "array",
+                    "elementSchema": "string"
+                  }
+                }
+              ]
+            }
+          }
```
MariusNi, line 300: Is there a P0 requirement for this immense object?

This thing has no use for GS and can blow up the Twins with its sheer size.

Delete

src/modules/mim/firewall.json
MariusNi: Remove one, keep the other. Simplify

src/modules/mim/firewall.json
MariusNi: An out rule does not need source info - the source is the local device,
An in rule does not need destination info - the destination is the local device.
Simplify and merge the two sets of ports and targets into one, whose direction is dictated by the rule kind, in or out.

src/modules/mim/firewall.json
MariusNi: action: addRule, deleteRule
ruleAction: allow, block

src/modules/mim/firewall.json
```diff
@@ -1,41 +1,305 @@
...
+                    },
+                    {
+                      "name": "drop",
+                      "enumValue": 3
+                    }
+                  ]
+                }
+              },
+              {
+                "name": "type",
```
MariusNi, line 92: >type [](http://example.com/codeflow?start=25&length=4)

direction

src/modules/mim/firewall.json
MariusNi: Keep them separate and add "Address" to source and destination

src/modules/mim/firewall.json
MariusNi: We discussed to keep 2 sets

src/modules/mim/firewall.json
```diff
@@ -1,41 +1,305 @@
...
+                "schema": "integer"
+              },
+              {
+                "name": "interface",
+                "schema": "string"
+              },
+              {
+                "name": "destinationInterface",
+                "schema": "string"
+              },
```
MariusNi, line 168: Let's remove. Always to be all interfaces for the port.

src/modules/mim/firewall.json
MariusNi: Let's omit the state array

robertschaedler3: Removed for simplification

src/modules/mim/firewall.json
robertschaedler3: Changed to inbound (default true) since this is the more common rule type

src/modules/mim/firewall.json
robertschaedler3: Removed large reported object

MariusNi (approved):
:shipit:

AhmedBM (approved):
👌

PR #316, opened 2022-07-27
Fixed pmc module recipe naming

AhmedBM:
## Description

* Renamed `packagemanagerconfigurationmodule` to `pmc`
* Removed unsupported types in `modulestest` as it conflicted with complex mimObjects

## Checklist


src/modules/test/RecipeInvoker.cpp
```diff
@@ -64,10 +64,6 @@ void RecipeInvoker::TestBody()
                         JSON_Value *value = json_object_get_value(jsonObject, setting.second.name.c_str());
                         EXPECT_EQ(JSONBoolean, json_value_get_type(value)) << "Expecting '" << m_recipe.m_objectName << "' to contain boolean setting '" << setting.second.name << "'" << std::endl << "JSON: " << payloadStr;
                     }
-                    else
-                    {
-                        FAIL() << "Unsupported type: " << setting.second.type << std::endl << "JSON: " << payloadStr;
```
hachotze, line 69: Do we lose any important coverage, thinking long-term, by removing this check?

AhmedBM: short answer, is yes, but this would block your work. Essentially, we support validating MimObject primitive types (string, number, boolean) but not complex objects which are supported in the MimObject, as a result we just fail if we validate any array or object type. This was found after including the pmc module which has a _state_ object it is validating.
There is already a bug tracking this... [Bug 40652731](https://microsoft.visualstudio.com/OS/_workitems/edit/40652731): [OSConfig][modulestest] MimObject Arrays and Objects settings are not validated correctly

hachotze (approved):

PR #328, opened 2022-08-08
9 kelvin n/sample rust module open close

9KelvinN:
## Description
Implemented MmiOpen() and MmiClose(). Fixed type names to Rust convention. Added unit tests for the sample struct. Initialized constants to be used in later implementation

## Checklist


src/modules/samples/rust/src/sample.rs
```diff
@@ -1,2 +1,89 @@
...
+    removed_string_map_setting_keys: Vec<String>,
+    removed_integer_map_setting_keys: Vec<i32>,
+}
+
+pub struct Sample {
+    max_payload_size_bytes: u32,
+}
+
+impl Sample {
+    pub fn new(max_payload_size_bytes: u32) -> Sample {
```
robertschaedler3, line 68: Is it better to use an explicit return type like `Sample` or `Self`? Both will give the same result, just curious if one is preferred over the other?

src/modules/samples/rust/src/sample.rs
```diff
@@ -1,2 +1,89 @@
...
+    fn get_max_payload_size_bytes(&self) -> u32 {
+        self.max_payload_size_bytes
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn build_sample() {
+        let test = super::Sample::new(16);
```
robertschaedler3, line 86: You should be able to add something like `use super::*;` at the top of `mod tests { ... }` to help with the namespacing/readability/maintainability of tests.

src/modules/samples/rust/src/sample.rs
```diff
@@ -1,2 +1,89 @@
...
+enum IntegerEnumeration {
+    None,
+    Value1,
+    Value2,
+}
+
+// A sample object with all possible setting types
+struct Object {
+    string_setting: String,
+    integer_etting: i32,
```
robertschaedler3, line 49: Missing `s` in `integer_etting` -> `integer_setting`

src/modules/samples/rust/src/sample.rs
```diff
@@ -1,2 +1,89 @@
...
+    max_payload_size_bytes: u32,
+}
+
+impl Sample {
+    pub fn new(max_payload_size_bytes: u32) -> Sample {
+        // Initializes a Sample struct
+        Sample {
+            max_payload_size_bytes: max_payload_size_bytes,
+        }
+        // The result is returned if the ending semicolon is omitted
```
robertschaedler3, line 73: Can this comment be moved to a different line? It is normal/common to omit the `;` in favor of not using the explicit `return` keyword. It might be confusing for others to see the comment after the return is implicitly *happening*.

src/modules/samples/rust/src/lib.rs
```diff
@@ -28,38 +30,44 @@ pub extern "C" fn MmiGetInfo(
...
     max_payload_size_bytes: c_uint,
-) -> MMI_HANDLE {
-    unimplemented!("MmiOpen is not yet implemented");
+) -> MmiHandle {
+    if client_name.is_null() {
+        println!("MmiOpen called with null clientName");
+        return ptr::null_mut() as *mut c_void;
+    }
+    let sample_box: Box<Sample> = Box::<Sample>::new(Sample::new(max_payload_size_bytes));
+    return Box::into_raw(sample_box) as *mut c_void;
```
robertschaedler3, line 39: Try to have a single execution path/only return once at the end of a function. This will make it easier to read/debug.

src/modules/samples/rust/src/sample.rs
```diff
@@ -1,2 +1,89 @@
...
+    }
+
+    #[cfg(test)]
```
robertschaedler3, line 76: Does this annotation need to be here (outside of the tests module)?

src/modules/samples/rust/src/lib.rs
```diff
@@ -1,24 +1,26 @@
 // Copyright (c) Microsoft Corporation. All rights reserved..
 // Licensed under the MIT License.
 
+use sample::Sample;
 use std::ffi::CStr;
 use std::ffi::CString;
 use std::os::raw::{c_char, c_int, c_uint, c_void};
+use std::ptr;
```
robertschaedler3, line 8: What is the convention for ordering `use`? More specifically is there a preference for where `use sample::Sample;` should be placed (before/after `mod sample`)?

src/modules/samples/rust/src/sample.rs
robertschaedler3: Discussed offline: this method will only be built with the tests. Very cool!

robertschaedler3 (approved):

PR #329, opened 2022-08-08
Implemented MmiGetInfo, added unit tests, shifted to libc

9KelvinN:
## Description
Implemented MmiGetInfo, added unit tests, shifted to libc

## Checklist


src/modules/samples/rust/src/lib.rs
```diff
@@ -23,14 +22,46 @@ pub extern "C" fn MmiGetInfo(
...
+        println!("MmiGetInfo called with null payload");
+        EINVAL
+    } else if payload_size_bytes.is_null() {
+        println!("MmiGetInfo called with null payloadSizeBytes");
+        EINVAL
+    } else {
+        // Borrow the client_name ptr
+        let client_name_cstr: &CStr = unsafe { CStr::from_ptr(client_name) };
+        let client_name_str_slice: &str;
+        match client_name_cstr.to_str() {
```
robertschaedler3, line 38: Please try to have a single return point in this function. You can return from a `match` block like how a status code `libc::c_int` is returned from each branch of the if-else statement.

robertschaedler3: clippy and rust-analyzer may have some good suggestions here too for the late initialization of the `&str` declared before each `match` block. Lets discuss

src/modules/samples/rust/src/sample.rs
```diff
@@ -72,6 +72,12 @@ impl Sample {
         }
     }
 
+    pub fn get_info(client_name: &str) -> &str {
+        // This sample module makes no use of the client_name, but
```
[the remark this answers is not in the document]
robertschaedler3: Same for line 17-18 (just noticed this now)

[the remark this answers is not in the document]
9KelvinN: Really? Combining the line puts the comment > 100 chars. And at 96 for second comment. What line length do we typically do?

src/modules/samples/rust/src/sample.rs
```diff
@@ -72,10 +73,11 @@ impl Sample {
         }
     }
 
-    pub fn get_info(client_name: &str) -> &str {
+    pub fn get_info(_client_name: &str) -> Result<&str, i32> {
```
robertschaedler3, line 76: Can the same `MmiError` type be used here too?

src/modules/samples/rust/src/lib.rs
```diff
@@ -14,6 +17,46 @@ pub const MMI_OK: i32 = 0;
 pub type MmiHandle = *mut c_void;
 pub type MmiJsonString = *mut c_char;
 
+#[derive(Debug)]
+enum MmiError {
+    FailedRead(Utf8Error),
+    Code(i32),
+    FailedAllocate(NulError),
+}
+impl From<Utf8Error> for MmiError {
```
robertschaedler3, line 26: Is there a more generic trait implementation that will work for converting from error?

robertschaedler3: What will happen if a new Error type needs to be captured? What does the process look like for a developer to integrate that into `MmiError`?

src/modules/samples/rust/src/lib.rs
```diff
@@ -14,6 +17,46 @@ pub const MMI_OK: i32 = 0;
...
+    fn from(err: Utf8Error) -> MmiError {
+        MmiError::FailedRead(err)
+    }
+}
+impl From<i32> for MmiError {
```
robertschaedler3, line 31: Here and elsewhere, please add spaces between different blocks

src/modules/samples/rust/src/lib.rs
9KelvinN: Discussed offline, options were box<Error>, struct, or this. Liked this for simplicity and that the other options may not actually simplify.

robertschaedler3 (approved):
Awesome 👍🏻

PR #331, opened 2022-08-10
Functional test recipes for PMC and HostName modules

hachotze:
## Description
- Added functional test recipes for PMC and HostName modules.
- Removed non-required WaitSeconds parameters from Networking and TPM recipes.
- Added zero case to check for 'payload too large' in HostNameBase::Set().
## Checklist


src/modules/test/RecipeInvoker.cpp
```diff
@@ -96,6 +96,8 @@ void RecipeInvoker::TestBody()
         std::cout << "Waiting for " << m_recipe.m_waitSeconds << " seconds" << std::endl;
         std::this_thread::sleep_for(std::chrono::seconds(m_recipe.m_waitSeconds));
     }
+
+    std::remove("/etc/osconfig/osconfig_commandrunner.cache");
```
robertschaedler3, line 100: Is it possible to save a copy of this file before running the tests, then restore the cache instead of removing it completely?

robertschaedler3: Small corner case to consider that could impact a dev environment significantly: 

1. Send Reboot/Shutdown via the IoTHub
2. Stop OSConfig (agent/platform) 
3. Run `modulestest`
4. Restart OSConfig **<-** *this will cause your dev machine to Reboot/Shutdown*

robertschaedler3:
It looks like renaming the hostname test recipe is breaking the build:

```
-- Adding module 'Hostname' to the build.
CMake Error at modules/CMakeLists.txt:22 (message):
  Module Hostname must have a recipe file at:
  /__w/azure-osconfig/azure-osconfig/src/modules/test/recipes/HostnameTests.json
Call Stack (most recent call first):
  modules/CMakeLists.txt:29 (add_module_custom_target)
  modules/CMakeLists.txt:41 (add_module)
```

src/modules/test/RecipeInvoker.cpp
AhmedBM: yes, it would affect it if that scenario was run. But this is a temporary workaround until we have a good setup/teardown solution

AhmedBM (commented):
I think the recipe rename will break the recipe configuration generation. You might have tyo resort to using `add_module_custom_target` in order to be explicit for the naming of your targets/mim/recipes

src/modules/test/RecipeInvoker.cpp
MariusNi: This is now done in main, please remove from this PR. Do not merge with this in, will duplicate the change (the one in main is on a different line)

src/modules/test/recipes/HostNameTests.json
```diff
@@ -0,0 +1,140 @@
...
+    "ObjectName": "name",
+    "Desired": true,
+    "Payload": "{}",
+    "ExpectedResult": 22
+  },
+  {
+    "ComponentName": "CommandRunner",
+    "ObjectName": "commandArguments",
+    "Desired": true,
+    "Payload": "{\"commandId\": \"saveHostNameInTempFile\", \"arguments\": \"echo $(hostname) > HostNameTestsTemp.txt\", \"timeout\": 0, \"singleLineTextResult\": true, \"action\": 3}",
```
MariusNi, line 55: >"Payload": "{\"commandId\": \"saveHostNameInTempFile\", \"arguments\": \"echo $(hostname) > HostNameTestsTemp.txt\", \"timeout\": 0, \"singleLineTextResult\": true, \"action\": 3}", [](http://example.com/codeflow?start=4&length=181)

There are no space separators in actual payloads. Suggest removing, here and elsewhere.

hachotze: Agree, addressed this in the latest iteration.

src/modules/test/RecipeInvoker.cpp
hachotze: Removed this in the latest iteration.

MariusNi (dismissed):
:shipit:

MariusNi (approved):
:shipit:

MariusNi (approved):
:shipit:

src/modules/test/recipes/PmcTests.json
```diff
@@ -0,0 +1,234 @@
...
+    "ObjectName": "desiredState",
+    "Desired": true,
+    "Payload": "{}",
+    "ExpectedResult": 22
+  },
+  {
+    "ComponentName": "CommandRunner",
+    "ObjectName": "commandArguments",
+    "Desired": true,
+    "Payload": "{\"commandId\":\"addMicrosoftPublicKey\",\"arguments\":\"apt-key adv --keyserver keyserver.ubuntu.com --recv-keys EB3E94ADBE1229CF\",\"timeout\":0,\"singleLineTextResult\":true,\"action\":3}",
```
AhmedBM, line 49: why are we using commandRunner to add the key? PMC supports the `gpgKeys` see https://github.com/Azure/azure-osconfig/blob/main/src/modules/pmc/README.md

hachotze: Agree, removed this check.

AhmedBM (approved):
Looks good 👌

PR #333, opened 2022-08-15
Sample rust module set

9KelvinN:
## Description

Finished the Rust Sample module. Implemented MmiSet and unit tests. 

## Checklist


src/modules/samples/rust/src/sample.rs
```diff
@@ -19,17 +19,8 @@ const REPORTED_OBJECT_NAME: &str = "reportedObject";
...
-const BOOLEAN_SETTING_NAME: &str = "booleanSetting";
-const INTEGER_ENUMERATION_SETTING_NAME: &str = "integerEnumerationSetting";
-const STRING_ARRAY_SETTING_NAME: &str = "stringsArraySetting";
-const INTEGER_ARRAY_SETTING_NAME: &str = "integerArraySetting";
-const STRING_MAP_SETTING_NAME: &str = "stringMapSetting";
-const INTEGER_MAP_SETTING_NAME: &str = "integerMapSetting";
-
 // r# Denotes a Rust Raw String
-const INFO: &str = r#"""({
+const INFO: &str = r#""""({
```
robertschaedler3, line 23: For this raw string, is it necessary to have the addtional `"""( ... )"""` at the beginning/end?  `r#" ... "` should be sufficient without the additional characters.

src/modules/samples/rust/src/lib.rs
```diff
@@ -94,7 +96,37 @@ pub extern "C" fn MmiSet(
...
 ) -> c_int {
-    unimplemented!("MmiSet is not yet implemented");
+    if client_session.is_null() {
+        println!("MmiSet called with null clientSession");
+        EINVAL
+    } else {
+        let sample: &mut Sample = unsafe { &mut *(client_session as *mut Sample) };
+        let component_name: &CStr = unsafe { CStr::from_ptr(component_name) };
+        let object_name: &CStr = unsafe { CStr::from_ptr(object_name) };
+        let payload: &CFixedStr = unsafe { CFixedStr::from_ptr(payload, payload_size_bytes as usize) };
```
robertschaedler3, line 106: Here and elsewhere, does it make sense to wrap these 4 lines in an `unsafe { }` block? What is the benefit to using unsafe for each line? If there is no difference, it might be clearer to group these things together in one block.

src/modules/samples/rust/Cargo.toml
```diff
@@ -10,6 +10,7 @@ edition = "2021"
 libc = "0.2"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
+c_fixed_string = "0.2"
```
robertschaedler3, line 13: Maybe sort these alphabetically?

src/modules/samples/rust/src/lib.rs
9KelvinN: I'd argue this better. If we did one big block, we'd have to place safe functions (like mmi_set_helper) in the unsafe block because of scoping: all of the variables would be in the unsafe block scope so the helper method would have to be in the block to access them.

robertschaedler3 (approved):
👍🏻

PR #339, opened 2022-08-19
Daemon Config Module MIM

9KelvinN:
## Description
Added the MIM for the Daemon Configuration Module

## Checklist

### DIFF
--- a/src/modules/mim/daemonconfiguration.json
+++ b/src/modules/mim/daemonconfiguration.json
@@ -1,0 +1,135 @@
+{
+    "name": "DaemonConfigurationModel",
+    "type": "mimModel",
+    "contents": [
+      {	
+        "name": "DaemonConfiguration",
+        "type": "mimComponent",
+        "contents": [
+          {
+            "name": "daemons",
+            "type": "mimObject",
+            "desired": false,
+            "schema": {
+              "type": "array",
+              "elementSchema": {
+                "type": "object",
+                "fields": [
+                  {
+                    "name": "name",
+                    "schema": "string"
+                  },
+                  {
+                    "name": "state",
+                    "schema": {
+                      "type": "enum",
+                      "valueSchema": "integer",
+                      "enumValues": [
+                        {
+                          "name": "unknown",
+                          "enumValue": 0
+                        },
+                        {
+                          "name": "running",
+                          "enumValue": 1
+                        },
+                        {
+                          "name": "failed",
+                          "enumValue": 2
+                        },
+                        {
+                          "name": "stopped",
+                          "enumValue": 3
+                        }
+                      ]
+                    }
+                  },
+                  {
+                    "name": "status",
+                    "schema": {
+                      "type": "enum",
+                      "valueSchema": "integer",
+                      "enumValues": [
+                        {
+                          "name": "unknown",
+                          "enumValue": 0
+                        },
+                        {
+                          "name": "enabled",
+                          "enumValue": 1
+                        },
+                        {
+                          "name": "disabled",
+                          "enumValue": 2
+                        },
+                        {
+                          "name": "static",
+                          "enumValue": 3
+                        },
+                        {
+                          "name": "indirect",
+                          "enumValue": 4
+                        }
+                      ]
+                    }
+                  }
+                ]
+              }
+            }
+          },
+          {
+            "name": "daemonCommands",
+            "type": "mimObject",
+            "desired": true,
+            "schema": {
+              "type": "array",
+              "elementSchema": {
+                "type": "object",
+                "fields": [
+                  {
+                    "name": "name",
+                    "schema": "string"
+                  },
+                  {
+                    "name": "state",
+                    "schema": {
+                      "type": "enum",
+                      "valueSchema": "integer",
+                      "enumValues": [
+                        {
+                          "name": "run",
+                          "enumValue": 0
+                        },
+                        {
+                          "name": "stop",
+                          "enumValue": 1
+                        }
+                      ]
+                    }
+                  },
+                  {
+                    "name": "status",
+                    "schema": {
+                      "type": "enum",
+                      "valueSchema": "integer",
+                      "enumValues": [
+                        {
+                          "name": "enable",
+                          "enumValue": 0
+                        },
+                        {
+                          "name": "disable",
+                          "enumValue": 1
+                        }
+                      ]
+                    }
+                  }
+                ]
+              }
+            }
+          }
+        ]
+      }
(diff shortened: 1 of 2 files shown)


src/modules/mim/daemonconfiguration.json
```diff
@@ -0,0 +1,143 @@
...
+                    "schema": "string"
+                  },
+                  {
+                    "name": "state",
+                    "schema": {
+                      "type": "enum",
+                      "valueSchema": "integer",
+                      "enumValues": [
+                        {
+                          "name": "run",
```
MariusNi, line 108: >run [](http://example.com/codeflow?start=35&length=3)

start is opposite of stop, use start instead of run

src/modules/mim/daemonconfiguration.json
```diff
@@ -0,0 +1,143 @@
...
+                        }
+                      ]
+                    }
+                  }
+                ]
+              }
+            }
+          },
+          {
+            "name": "daemonCommands",
```
MariusNi, line 89: >daemonCommands [](http://example.com/codeflow?start=21&length=14)

Suggested: "desiredDaemonConfigfuration"

src/modules/mim/daemonconfiguration.json
```diff
@@ -0,0 +1,143 @@
+{
+    "name": "DaemonConfigurationModel",
+    "type": "mimModel",
+    "contents": [
+      {	
+        "name": "DaemonConfiguration",
+        "type": "mimComponent",
+        "contents": [
+          {
+            "name": "daemons",
```
MariusNi, line 10: >daemons [](http://example.com/codeflow?start=21&length=7)

daemonConfiguration

src/modules/mim/daemonconfiguration.json
```diff
@@ -0,0 +1,143 @@
...
+                          "enumValue": 1
+                        }
+                      ]
+                    }
+                  }
+                ]
+              }
+            }
+          }
+        ]
```
MariusNi, line 139: How does the desired object specify desired state and desired status for each existing daemon? 

Cannot expect the order in the both arrays to be same at all times.  The desired object may need the names at least.

src/modules/mim/daemonconfiguration.json
```diff
@@ -0,0 +1,143 @@
...
+                        },
+                        {
+                          "name": "stop",
+                          "enumValue": 1
+                        }
+                      ]
+                    }
+                  },
+                  {
+                    "name": "status",
```
MariusNi, line 119: Merge state and status in one?

enable, disable, start, stop, can be issued one at a time. 

Thus this can be a single enumeration for desired state of the deamon named by the name setting:

0 (none)
1 (enable)
2 (disable)
3 (start)
4 (stop)

src/modules/mim/daemonconfiguration.json
```diff
@@ -0,0 +1,143 @@
...
+                        },
+                        {
+                          "name": "dead",
+                          "enumValue": 4
+                        }
+                      ]
+                    }
+                  },
+                  {
+                    "name": "status",
```
MariusNi, line 52: "state" and "status" here for reported object may need to be separate indeed but the names are basically the same. Either think of more different names or merge the two into a single enum, like suggested for the desired object "desired state".

src/modules/mim/daemonconfiguration.json
```diff
@@ -0,0 +1,143 @@
+{
+    "name": "DaemonConfigurationModel",
```
MariusNi, line 2: >Daemon [](http://example.com/codeflow?start=13&length=6)

This module is just for systemd daemons, correct?

Then add that in description:

"SystemdDaemonConfiguration"

Can also replace "daemon" by "service" here and elsewhere

src/modules/mim/daemonconfiguration.json
```diff
@@ -0,0 +1,143 @@
...
+                          "enumValue": 0
+                        },
+                        {
+                          "name": "enabled",
+                          "enumValue": 1
+                        },
+                        {
+                          "name": "enabledRuntime",
+                          "enumValue": 2
+                        },
```
MariusNi, line 68: What is the difference between 1 and 2?

src/modules/mim/daemonconfiguration.json
```diff
@@ -0,0 +1,143 @@
...
+                          "name": "disabled",
+                          "enumValue": 3
+                        },
+                        {
+                          "name": "static",
+                          "enumValue": 4
+                        },
+                        {
+                          "name": "indirect",
+                          "enumValue": 5
```
MariusNi, line 79: May want to use more descriptive names for these two, or remove

MariusNi:
Unlike the Rust sample, this one is not appropriate for the same branch or main repo. Instead, this module should be solely in the prototypes repo.

MariusNi (dismissed):
:clock1:

MariusNi:
As discussed offline, we will target separate branches for the sample and the daemon module and keep main repo to have all infra support.

---
In reply to: [1221082959](https://github.com/Azure/azure-osconfig/pull/339#issuecomment-1221082959) [](http://example.com/codeflow?ancestors=1221082959,1221082959)

src/modules/mim/daemonconfiguration.json
MariusNi: As discussed offline, we'll do for the desired object:

string name;
boolean enabled;
boolean started;

src/modules/mim/daemonconfiguration.json
MariusNi: If we rename the component and model adding "systemd" in names, then we can use for the 2 objects respectively:

"daemonConfiguration"
"desiredDaemonConfiguration"

src/modules/mim/daemonconfiguration.json
```diff
@@ -0,0 +1,143 @@
+{
```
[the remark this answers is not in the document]
MariusNi: Correct. Source code uses 4 spaces per tab, JSON use 2 spaces per tab (throughout the entire OSConfig)

9KelvinN:
Table from the systemctl man:
image

static and indirect are quite different from enable and disable, so I think it's correct to leave them in. It's also not substantial additional work to leave them in. 
Similarly for exited vs dead. exited means that it ran successfully and exited (but it's a subset of active) while dead means the service was and is inactive.

Agreed that enabled runtime doesn't seem much different from enabled so removed that one

MariusNi (approved):
:shipit:

PR #345, opened 2022-08-22
Added build-time MIM validation

AhmedBM:
## Description

* Changed default container to use `289651` (added jsonschema) and not `latest`. We want to make sure we use a specific version tag and not latest which might represent a broken reference.
* Added build-time validation of MIMs when jsonschema is installed (optional, CMake outputs a warning if jsonschema is not installed)
* Corrected `commandrunner` and `deviceinfo` MIMs to pass schema validation

## Checklist

### DIFF
--- a/.github/actions/container-run/action.yml
+++ b/.github/actions/container-run/action.yml
@@ -27,7 +27,7 @@ inputs:
   tag:
     description: The container tag.
     required: false
-    default: latest
+    default: "289651"
 
 outputs:
   id:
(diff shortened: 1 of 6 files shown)


.github/actions/container-run/action.yml
```diff
@@ -27,7 +27,7 @@ inputs:
   tag:
     description: The container tag.
     required: false
-    default: latest
+    default: "289651"
```
MariusNi, line 30: >289651 [](http://example.com/codeflow?start=14&length=6)

What is this "magic" number?

MariusNi (approved):
:shipit:

PR #348, opened 2022-08-23
Daemon config module get

9KelvinN:
## Description
Added framework for the Daemon Config Module, added the module to the cmake buildline, implemented it's get method, and added unit tests.

## Checklist


src/modules/daemonconfiguration/src/common.rs
```diff
@@ -0,0 +1,75 @@
...
+use libc::{c_int, EINVAL, ENOMEM};
+use std::ffi::NulError;
+use std::fmt;
+use std::str::Utf8Error;
+
+#[derive(Debug, PartialEq)]
+pub enum MmiError {
+    FailedRead,
+    FailedAllocate,
+    InvalidArgument,
```
robertschaedler3, line 13: Consider adding context for these errors. Ex: `InvalidArgument(String)` to provide the name of the argument in log traces.

src/modules/daemonconfiguration/src/daemonconfiguration.rs
```diff
@@ -0,0 +1,380 @@
...
+    name: String,
+    state: State,
+    auto_start_status: AutoStartStatus,
+}
+
+pub trait SystemctlInfo {
+    fn get_daemons() -> Result<Vec<Daemon>, MmiError>;
+    fn list_unit_files() -> Result<String, MmiError>;
+    fn show_substate_property(name: &str) -> Result<String, MmiError>;
+    fn create_daemon(name: &str, auto_start_status: &str) -> Result<Daemon, MmiError>;
```
robertschaedler3, line 58: Consider redefining the `Result` type to make all the return types more concise:

```Rust
type Result<T> = std::result::Result<T, MmiError>;
```

src/modules/daemonconfiguration/src/daemonconfiguration.rs
```diff
@@ -0,0 +1,380 @@
...
+            Command::new("systemctl")
+                .args(["show", name, "--property=SubState"])
+                .output()?
+        };
+        Ok(String::from(std::str::from_utf8(&output.stdout)?))
+    }
+}
+
+#[derive(Default, Debug)]
+pub struct DaemonConfiguration {
```
robertschaedler3, line 126: Consider moving the generic type to be at the struct level. This will ensure all instances of the struct will use the same underlying systemctl implementation (actual vs mock). instead of just get/set method.

src/modules/daemonconfiguration/src/daemonconfiguration.rs
```diff
@@ -0,0 +1,380 @@
...
+            // Whether the caller was booted using Systemd
+            Err(MmiError::SystemdError)
+        } else if !COMPONENT_NAME.eq(component_name) {
+            println!("Invalid component name: {}", component_name);
+            Err(MmiError::InvalidArgument)
+        } else if !OBJECT_NAME.eq(object_name) {
+            println!("Invalid object name: {}", object_name);
+            Err(MmiError::InvalidArgument)
+        } else {
+            let daemons = T::get_daemons()?;
```
robertschaedler3, line 160: A different name other that `T` may be more readable here

src/modules/CMakeLists.txt
```diff
@@ -1,6 +1,17 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 
+if (BUILD_RUST)
```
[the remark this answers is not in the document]
MariusNi: Basically, to be moved up in main CMakeList.txt, where BUILD_RUST would be conditionally set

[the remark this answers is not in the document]
9KelvinN: Yeah, this fetch could just be in the base level Cmake, but atm it's only used for modules. Shouldn't it be left at the lowest level? Would prevent being fetched if modules aren't being built

src/modules/daemonconfiguration/src/common.rs
9KelvinN: We'd discussed this for the Sample module! More specific logging is left at the rust lib level (specific logging can be done before the error is returned). But it's hard to use the specific String arg with a wide catching match statement because the String can't be incorporated into the fmt::display return string

src/modules/daemonconfiguration/src/daemonconfiguration.rs
9KelvinN: So did some experimentation and may have done it wrong, but then the struct needs to have fields of the generic type? (which would be unnecessary because an actual instance isn't needed, only the static methods). Also, leaving it get/set level prevents needing to specify the generic type for new() and getInfo()

robertschaedler3: Instead of adding to the struct itself, it should be possible to add on the `impl` block for the struct: 

https://dev.to/somedood/generic-impl-blocks-are-kinda-like-macros-1aa0

*Also check out the optimization section of that article*

src/modules/CMakeLists.txt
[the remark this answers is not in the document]
MariusNi: One single place, at the root CMakeList.txt is best. Same one place where the BUILD_RUST will be set, conditioned (in the future) on the presence of RUST dependencies.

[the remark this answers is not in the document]
robertschaedler3: The fetch should be as high up as possible, ideally only being fetched if Rust components are being built.

src/modules/daemonconfiguration/src/daemonconfiguration.rs
9KelvinN: Am I misunderstanding? impl<T> requires that the struct have a generic type (DaemonConfiguration<T>) so it'd have to be added to the struct. Also which optimization section are you referring to?

9KelvinN: Discussed offline, current implementation is fine practice (tie the generic to the method instead of the struct) for code cleanliness and abstraction

robertschaedler3 (approved):

PR #351, opened 2022-08-23
Sample rust module cleanup

9KelvinN:
## Description
Renamed a variable and removed print debug line

## Checklist

### DIFF
--- a/src/modules/samples/rust/src/sample.rs
+++ b/src/modules/samples/rust/src/sample.rs
@@ -111,35 +111,35 @@ impl Sample {
         &mut self,
         component_name: &str,
         object_name: &str,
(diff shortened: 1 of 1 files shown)


MariusNi:
>                     println!("Invalid object name: {}", object_name);

The sample should use OSConfig's logging library if possible and log correctly, like all other modules. 

Can this sample use the C macros (via a C adapter) so it can do proper logging?

---
Refers to: src/modules/samples/rust/src/sample.rs:176 in fe33612. [](commit_id = fe33612885d71435749df5c3d6011f67c2d7953e, deletion_comment = False)

MariusNi (approved):
:shipit:

PR #359, opened 2022-08-29
Sample rust module payload size exceeded

9KelvinN:
## Description
Added a new payload size exceeded error type and message.

## Checklist

### DIFF
--- a/src/modules/samples/rust/src/sample.rs
+++ b/src/modules/samples/rust/src/sample.rs
@@ -181,11 +181,11 @@ impl Sample {
             if self.max_payload_size_bytes != 0
             && payload.len() as u32 > self.max_payload_size_bytes {
                 // In other modules you could prioritize parts of the payload
-                println!("Payload size exceeded max payload size bytes in get");
-                Err(MmiError::PayloadSizeExceeded)
-            } else {
-                Ok(payload)
+                println!("Payload size exceeded max payload size bytes in get so it was truncated.");
+                let payload_bytes = payload.into_bytes();
+                let payload = &payload_bytes[0..max_payload_size_bytes];
             }
+            Ok(payload)
         } else {
             println!("Invalid component name: {}", component_name);
             Err(MmiError::InvalidArgument)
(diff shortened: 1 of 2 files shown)


MariusNi:
>     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {

Modules cannot display errors to the user. They run in machine session and they can only log and return error codes to MMI calls. 

Suggest clear this up throughout the project to remove any impression that this module may display something to the user or similar.

---
Refers to: src/modules/samples/rust/src/common.rs:47 in 74bc447. [](commit_id = 74bc447a31e7214f45562ed07a58b4815cfd7142, deletion_comment = False)

src/modules/samples/rust/src/sample.rs
```diff
@@ -181,7 +181,8 @@ impl Sample {
             if self.max_payload_size_bytes != 0
             && payload.len() as u32 > self.max_payload_size_bytes {
                 // In other modules you could prioritize parts of the payload
-                Err(MmiError::InvalidArgument)
+                println!("Payload size exceeded max payload size bytes in get");
+                Err(MmiError::PayloadSizeExceeded)
```
MariusNi, line 185: For MmiGet, exceeding the maximum size cannot result in an error. For MmiGet the module needs to truncate reported payload so that it fits the maximum size indicated by MmiOpen.

Please fix this.

Failing MmiSet for payload that exceeds maximum size is ok.

MariusNi (changes requested):
:clock1:

src/modules/samples/rust/src/sample.rs
9KelvinN: Changed to truncate the payload although this branch should never trigger based on the behavior of the sample module. It saves in the object whatever was passed into set. And the set payload must be below max_payload_size_bytes, so when the stored value is reconverted into a string payload, it will still be below max_payload_size_bytes.

robertschaedler3 (approved):

MariusNi (approved):
:shipit:

PR #360, opened 2022-08-29
Daemon config module payload filter

9KelvinN:
## Description
Added payload filtering to only report enabled daemons that don't have state "other" or "dead". Added a new payload size exceeded error type and message. Fixed wrong component and object names. Changed unit tests to reflect filtering.

## Checklist

### DIFF
--- a/src/modules/daemonconfiguration/src/common.rs
+++ b/src/modules/daemonconfiguration/src/common.rs
@@ -5,6 +5,7 @@ use libc::{c_int, EINVAL, ENOMEM};
 use std::ffi::NulError;
 use std::fmt;
 use std::str::Utf8Error;
+use std::string::FromUtf8Error;
 
 #[derive(Debug, PartialEq)]
 pub enum MmiError {
@@ -15,6 +16,12 @@ pub enum MmiError {
     SerdeError,
     SystemctlError,
     SystemdError,
+}
+
+impl From<FromUtf8Error> for MmiError {
+    fn from(_err: FromUtf8Error) -> MmiError {
+        MmiError::FailedRead
+    }
 }
 
 impl From<Utf8Error> for MmiError {
--- a/src/modules/daemonconfiguration/src/common.rs
+++ b/src/modules/daemonconfiguration/src/common.rs
@@ -11,6 +11,7 @@ pub enum MmiError {
     FailedRead,
     FailedAllocate,
     InvalidArgument,
+    PayloadSizeExceeded,
     SerdeError,
     SystemctlError,
     SystemdError,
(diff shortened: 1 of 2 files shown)


src/modules/daemonconfiguration/src/common.rs
```diff
@@ -11,6 +11,7 @@ pub enum MmiError {
     FailedRead,
     FailedAllocate,
     InvalidArgument,
+    PayloadSizeExceeded,
```
MariusNi, line 14: >PayloadSizeExceeded, [](http://example.com/codeflow?start=4&length=20)

This is only valid (as an error) for MmiSet. MmiGet cannot return this error, instead, it must truncate payload.

src/modules/daemonconfiguration/src/daemonconfiguration.rs
```diff
@@ -83,7 +83,10 @@ impl SystemctlInfo for Systemctl {
                 continue;
             }
             let daemon = Self::create_daemon(&service["service"], &service["status"])?;
-            services.push(daemon);
+            // Only report enabled daemons with State not being Other
+            if daemon.state != State::Other && daemon.state != State::Dead && daemon.auto_start_status == AutoStartStatus::Enabled {
```
MariusNi, line 87: >f daemon.state != State::Other && daemon.state != State::Dead && daemon.auto_start_status == AutoStartStatus::Enabled  [](http://example.com/codeflow?start=13&length=118)

Does Rust have operator precedence? Then, don't rely on it and separate statement with extra parenthesis?

robertschaedler3: Yes, however adding extra parentheses often results in linting errors during `cargo build` which will show up as warnings when osconfig is build via `cmake`. @9KelvinN I suggest adding parentheses wherever possible without creating build warnings like @MariusNi suggests (we should not suppress these types of warnings).

robertschaedler3: Also, the opposite is true: the compiler will warn you if the precedence between 2 operators is unclear and suggest adding parenthesis where necessary to avoid such operator precedence errors.

https://doc.rust-lang.org/reference/expressions.html

9KelvinN: Heeded the suggestions, added parentheses to explicitize order of operations

src/modules/daemonconfiguration/src/common.rs
9KelvinN: Changed

robertschaedler3 (approved):

PR #365, opened 2022-08-31
Firewall v3 implementation

robertschaedler3:
## Description

Implement Firewall v3 following the updated MIM. 

## Checklist


src/modules/firewall/src/lib/Firewall.cpp
```diff
@@ -196,4 +679,309 @@ std::string IpTables::Fingerprint() const
...
+        m_policyStatusMessage = "";
+    }
+
+    return status;
+}
+
+std::vector<IpTablesPolicy> IpTables::GetDefaultPolicies() const
+{
+    const char* command = "iptables -S | grep -E '^-P (INPUT|OUTPUT)'";
+    static const std::regex policyRegex("^-P\\s+(\\S+)\\s+(\\S+)$");
```
AhmedBM, line 887: does this work across all our known distros? I suggest adding a recipe that exercises this code

src/modules/firewall/src/lib/Firewall.h
```diff
@@ -39,9 +40,128 @@ class FirewallLog
...
+    virtual int DirectionFromChain(const std::string& str);
+};
+
+class IpTablesPolicy : public IpTablesRule
+{
+public:
+    typedef GenericRule::Action Action;
+    typedef GenericRule::Direction Direction;
+
+    IpTablesPolicy() = default;
```
AhmedBM, line 143: There shouldn't be a need to explicitly define these default functions, is there?

src/modules/firewall/src/lib/Firewall.h
```diff
@@ -103,22 +287,62 @@ class FirewallModule : public FirewallModuleBase
...
+    virtual int GetDefaultPolicies(rapidjson::Writer<rapidjson::StringBuffer>& writer) const override
+    {
+        std::vector<Policy> policies = m_firewall.GetDefaultPolicies();
+        writer.StartArray();
+        for (const Policy& policy : policies)
+        {
+            policy.Serialize(writer);
+        }
+        writer.EndArray();
+        return 0;
```
AhmedBM, line 317: return `EXIT_SUCCESS`?

src/modules/firewall/src/lib/Firewall.h
```diff
@@ -103,22 +287,62 @@ class FirewallModule : public FirewallModuleBase
...
+    }
+
+    virtual int GetConfigurationStatus(rapidjson::Writer<rapidjson::StringBuffer>& writer) const override
+    {
+        Status status = m_firewall.GetStatus();
+        writer.String((status == Status::Success) ? "success" : ((status == Status::Failure) ? "failure" : "unknown"));
+        return 0;
```
AhmedBM, line 324: return `EXIT_SUCCESS`?

src/modules/firewall/src/lib/Firewall.h
```diff
@@ -103,22 +287,62 @@ class FirewallModule : public FirewallModuleBase
...
+    }
+
+    virtual int GetConfigurationStatusDetail(rapidjson::Writer<rapidjson::StringBuffer>& writer) const override
+    {
+        std::string statusDetail = m_firewall.GetStatusMessage();
+        writer.String(statusDetail.c_str());
+        return 0;
```
AhmedBM, line 331: return `EXIT_SUCCESS`?

src/modules/firewall/src/lib/Firewall.h
```diff
@@ -103,22 +287,62 @@ class FirewallModule : public FirewallModuleBase
...
+    }
+
+    virtual int SetDefaultPolicies(rapidjson::Document& document) override
+    {
+        std::vector<Policy> policies = ParseArray<Policy>(document);
+        return (!policies.empty()) ? m_firewall.SetDefaultPolicies(policies) : -1;
```
AhmedBM, line 337: `<expr> ? EXIT_SUCCESS : EXIT_FAILURE`

src/modules/firewall/src/lib/Firewall.h
```diff
@@ -103,22 +287,62 @@ class FirewallModule : public FirewallModuleBase
...
+    }
+
+    virtual int SetRules(rapidjson::Document& document) override
+    {
+        std::vector<Rule> rules = ParseArray<Rule>(document);
+        return (!rules.empty()) ? m_firewall.SetRules(rules) : -1;
```
AhmedBM, line 343: `<expr> ? EXIT_SUCCESS : EXIT_FAILURE`

AhmedBM (commented):
Make sure to also include a test recipe to exercise the behavior across all our supported distros

src/modules/firewall/src/lib/Firewall.cpp
```diff
@@ -141,18 +191,451 @@ int FirewallModuleBase::Get(const char* componentName, const char* objectName, M
...
+    {
+        OsConfigLogError(FirewallLog::Get(), "Invalid payload size: %d", payloadSizeBytes);
+        status = EINVAL;
+    }
+    else
+    {
+        std::string payloadJson = std::string(payload, payloadSizeBytes);
+        size_t payloadHash = HashString(payloadJson.c_str());
+
+        if (payloadHash != m_lastPayloadHash)
```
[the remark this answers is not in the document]
robertschaedler3: Yes, it is mainly there to prevent re-running a bunch of system commands just to check if the desired rules/policies are the same.

src/modules/firewall/src/lib/Firewall.cpp
robertschaedler3: Yes, the firewall module already has a dependency on `iptables` (for fingerprint and enabled state) so the same will work for default policies etc.

robertschaedler3: Agreed, I will add test recipes to this PR

src/modules/firewall/src/lib/Firewall.h
robertschaedler3: Good point! I will remove all these default constructors

src/modules/firewall/src/lib/Firewall.h
```diff
@@ -39,32 +40,167 @@ class FirewallLog
...
+        Out
+    };
+
+    enum class Protocol
+    {
+        Any,
+        Tcp,
+        Udp,
+        Icmp,
+    };
```
MariusNi, line 71: If we proceed with enum of strings in MIM, why not use enum of strings in implementation as well?

src/modules/firewall/src/lib/Firewall.h
```diff
@@ -39,32 +40,167 @@ class FirewallLog
...
+
+    virtual int SetActionFromString(const std::string& str);
+    virtual int SetDirectionFromString(const std::string& str);
+    virtual int SetStateFromString(const std::string& str);
+    virtual int SetProtocolFromString(const std::string& str);
+
+    virtual std::string ActionToString() const = 0;
+    virtual std::string DirectionToString() const = 0;
+    virtual std::string ProtocolToString() const = 0;
+
```
MariusNi, line 100: If using enums of strings in code same as in MIM, these won't be necessary.

MariusNi: You can define your own type for enumeration of strings.

src/modules/firewall/src/lib/Firewall.cpp
```diff
@@ -15,8 +15,46 @@ const std::string FirewallModuleBase::m_moduleInfo = R"""({
...
+const char g_protocolTCP[] = "tcp";
+const char g_protocolUDP[] = "udp";
+const char g_protocolICMP[] = "icmp";
+
+const char g_targetAccept[] = "ACCEPT";
+const char g_targetReject[] = "REJECT";
+const char g_targetDrop[] = "DROP";
+
+const char g_chainInput[] = "INPUT";
+const char g_chainOutput[] = "OUTPUT";
```
MariusNi, line 57: Can do const char* g_chain[] = {"INPUT", "OUTPUT"}; or similar. Idea being: arrays of strings simulating enumerations, rather than standalone constants.

src/modules/firewall/src/lib/Firewall.h
```diff
@@ -39,32 +40,167 @@ class FirewallLog
...
+        Accept,
+        Drop,
+        Reject
+    };
+
+    enum class Direction
+    {
+        In,
+        Out
+    };
```
MariusNi, line 63: Why not arrays of string constants that would be directly used for parsing and assembling payloads? These enums are no different from the plain enums used before, they still need conversion to string constants. See the other comment for an example.

src/modules/firewall/src/lib/Firewall.h
```diff
@@ -39,32 +40,265 @@ class FirewallLog
...
+    virtual bool IsValid() const = 0;
+
+    std::string ToString() const
+    {
+        return m_value;
+    }
+
+    bool operator==(const char* other) const
+    {
+        return m_value.compare(other) == 0;
```
MariusNi, line 59: >m_value.compare(other) == 0; [](http://example.com/codeflow?start=15&length=28)

nit: add () around, here and on line 64 and similar.

You can also use cast to bool instead ==/!= 0.

return static_cast<bool>(m_value.compare(other));

etc

src/modules/firewall/src/lib/Firewall.h
```diff
@@ -39,32 +40,265 @@ class FirewallLog
...
+    std::string m_value;
+};
+
+class GenericRule
+{
+public:
+    class State : public StringEnum
+    {
+    public:
+        static const std::array<std::string, 2> m_values;
```
MariusNi, line 83: >2> m_values [](http://example.com/codeflow?start=45&length=11)

Why 2? Define a constant for this instead?

src/modules/firewall/src/lib/Firewall.h
```diff
@@ -39,32 +40,265 @@ class FirewallLog
...
+
+        State() = default;
+        State(const std::string& value) : StringEnum(value) {}
+
+        bool IsValid() const override
+        {
+            return std::find(m_values.begin(), m_values.end(), m_value) != m_values.end();
```
MariusNi, line 90: >std::find(m_values.begin(), m_values.end(), m_value) != m_values.end(); [](http://example.com/codeflow?start=19&length=71)

Add some () here (in general, do not skip on ())

src/modules/firewall/src/lib/Firewall.h
```diff
@@ -39,32 +40,265 @@ class FirewallLog
...
+        }
+    };
+
+    class Action : public StringEnum
+    {
+    public:
+        static const std::array<std::string, 3> m_values;
```
MariusNi, line 97: Define constant? Same in all other similar places

src/modules/firewall/src/lib/Firewall.h
```diff
@@ -39,32 +40,265 @@ class FirewallLog
...
+
+    class Direction : public StringEnum
+    {
+    public:
+        static const std::array<std::string, 2> m_values;
+
+        Direction() = default;
+        Direction(const std::string& value) : StringEnum(value) {}
+
+        bool IsValid() const override
```
MariusNi, line 117: >bool IsValid() const override [](http://example.com/codeflow?start=8&length=29)

Should there be error logging in this code to explain failures for MmiGet and MmiSet?

src/modules/firewall/src/lib/Firewall.h
```diff
@@ -39,32 +40,265 @@ class FirewallLog
...
+private:
+    State m_desiredState;
+};
+
+class GenericPolicy
+{
+public:
+    class Action : public StringEnum
+    {
+    public:
```
MariusNi, line 176: Is it necessary class definition within a class definition? This makes the code more complicated than it is. Untangle and befriend, or ...?

src/modules/firewall/tests/FirewallTests.cpp
```diff
@@ -8,52 +8,149 @@
...
+        rapidjson::Document expected;
 
-    class MockUtility : public GenericFirewall
+        if (expected.Parse(expectedJson.c_str()).HasParseError())
+        {
+            return testing::AssertionFailure() << "expected JSON is not valid JSON";
+        }
+        else if (actual.Parse(actualJson.c_str()).HasParseError())
+        {
+            return testing::AssertionFailure() << "actual JSON is not valid JSON";
```
MariusNi, line 22: These should never happen in an unit-test. Either positive or negative the test shall pass without extra assertions.

src/modules/firewall/tests/FirewallTests.cpp
```diff
@@ -8,52 +8,149 @@
...
     {
-        return State::Enabled;
+        return (!m_rules.empty() && !m_defaultPolicies.empty()) ? State::Enabled : State::Disabled;
     }
 
-    std::string MockUtility::Fingerprint() const
+    std::string MockFirewall::Fingerprint() const
     {
-        return expectedFingerprint;
+        // For testing purposes, return the number of rules + the number of default policies.
```
MariusNi, line 63: >+ [](http://example.com/codeflow?start=60&length=1)

"plus" (use full English in comments for consistency with rest of the code)

src/modules/firewall/src/lib/Firewall.h
robertschaedler3: Depending on where the string enum is used (parsing, serializing, etc) `IsValid()` is used similarly to `HasParseError()`

src/modules/firewall/tests/FirewallTests.cpp
robertschaedler3: This is a custom assertion for comparing the JSON values returned by `Firewall::Get()`

src/modules/firewall/src/lib/Firewall.h
robertschaedler3: Agreed, move out of class definition to simplify 👍🏻

src/modules/firewall/tests/FirewallTests.cpp
```diff
@@ -8,52 +8,149 @@
...
+            return testing::AssertionFailure() << "actual JSON is not valid JSON" << actualJson;
+        }
+        else if (actual == expected)
+        {
+            return testing::AssertionSuccess();
+        }
+        else
+        {
+            return testing::AssertionFailure() << "expected:\n\t" << expectedJson << "\n but got:\n\t" << actualJson;
+        }
```
MariusNi, line 31: Rather than many exists from function, assign to some internal variable from if-else and return once.

src/modules/firewall/tests/FirewallTests.cpp
```diff
@@ -73,15 +170,288 @@ namespace tests
...
+        std::string rule = "{\"desiredState\": \"" + desiredState + "\", \"action\": \"" + action + "\", \"direction\": \"" + direction + "\"";
+        if (!protocol.empty()) {
+            rule += ", \"protocol\": \"" + protocol + "\"";
+        }
+        if (!src.empty()) {
+            rule += ", \"src\": \"" + src + "\"";
+        }
+        if (!dst.empty()) {
+            rule += ", \"dst\": \"" + dst + "\"";
+        }
```
MariusNi, line 253: Break {}

statement 
{

}

Not:

statement {

}

MariusNi (approved):
:shipit:

PR #373, opened 2022-09-08
Conditional rust build and Corrosion

9KelvinN:
## Description
Corrosion is no longer fetched, instead it's a subdirectory. BUILD_RUST is no longer a manual flag, it's conditionally set on the presence of cargo and rustc

## Checklist

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -120,7 +120,6 @@ find_program(RUSTC_EXECUTABLE rustc
 mark_as_advanced(RUSTC_EXECUTABLE)
 
 # Fetch Corrosion to link the Rust Modules with the CMake Build Line
-
 if(CARGO_EXECUTABLE AND RUSTC_EXECUTABLE)
     set(BUILD_RUST true)
     add_subdirectory(corrosion)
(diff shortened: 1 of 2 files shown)


src/CMakeLists.txt
```diff
@@ -94,6 +93,40 @@ endif()
 set(PLATFORM_INC_DIR ${CMAKE_CURRENT_SOURCE_DIR}/platform/inc)
 set(MODULES_INC_DIR ${CMAKE_CURRENT_SOURCE_DIR}/modules/inc)
 
+if(CMAKE_HOST_WIN32)
```
robertschaedler3, line 96: No need to check for WIN32, osconfig is only for linux

src/CMakeLists.txt
```diff
@@ -94,6 +93,40 @@ endif()
...
 set(MODULES_INC_DIR ${CMAKE_CURRENT_SOURCE_DIR}/modules/inc)
 
+if(CMAKE_HOST_WIN32)
+	set(USER_HOME "$ENV{USERPROFILE}")
+else()
+	set(USER_HOME "$ENV{HOME}")
+endif()
+
+if(NOT DEFINED CARGO_HOME)
+	if("$ENV{CARGO_HOME}" STREQUAL "")
```
robertschaedler3, line 103: Spacing here and elsewhere looks off (tabs instead of 4 spaces)

robertschaedler3 (approved):

PR #375, opened 2022-09-09
Removed module identity on package removal

AhmedBM:
## Description

* Removed osconfig module identity on Debian/RPM package removal

## Checklist

### DIFF
--- a/devops/debian/prerm
+++ b/devops/debian/prerm
@@ -44,4 +44,9 @@ if [ -d /run/systemd/system ] && [ "$1" = remove ]; then
 fi
 # End automatically added section
 
+# MSFT change: Remove module identity on package removal
+if [ "$1" = remove ]; then
+    rm /etc/aziot/identityd/config.d/osconfig.toml >/dev/null || true
+fi
+
 exit 0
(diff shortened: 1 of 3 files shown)


devops/debian/prerm
```diff
@@ -44,4 +44,9 @@ if [ -d /run/systemd/system ] && [ "$1" = remove ]; then
 fi
 # End automatically added section
 
+# MSFT change: Remove module identity on package removal
```
MariusNi, line 47: >MSFT [](http://example.com/codeflow?start=2&length=4)

Do we need "MSFT" in comments? MSFT is an internal term that should not appear in our public repo.

We can use "OSConfig Core Team" if needed.

AhmedBM: This was the pattern we used for any of the changes made after applying the debhelper templates, its not the only place with the "MSFT changes", they are also defined in https://github.com/Azure/azure-osconfig/blob/main/devops/debian/postinst

If we change, we should change all instances.

MariusNi: Yes, please lets' change all comments and names stating "MSFT" which are not really appropriate for our public repro. Instead use "Core Team" or "OsConfig Core Team" as referenced in out documents in this repo. Thanks!

MariusNi (approved):
:shipit:

PR #382, opened 2022-09-12
Changed modulestest clientName to use default clientName

AhmedBM:
## Checklist

### DIFF
--- a/src/modules/test/RecipeInvoker.h
+++ b/src/modules/test/RecipeInvoker.h
@@ -4,7 +4,7 @@
 #ifndef RECIPEINVOKER_H
 #define RECIPEINVOKER_H
 
-static const std::string g_defaultClient = "RecipeInvoker";
+static const std::string g_defaultClient = "Azure OsConfig";
 
 static const std::string g_moduleName = "ModuleName";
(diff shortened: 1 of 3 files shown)


src/modules/test/RecipeInvoker.h
```diff
@@ -4,7 +4,7 @@
 #ifndef RECIPEINVOKER_H
 #define RECIPEINVOKER_H
 
-static const std::string g_defaultClient = "RecipeInvoker";
+static const std::string g_defaultClient = "Azure OsConfig";
```
robertschaedler3, line 7: This will still fail the validation used by the module here: https://github.com/Azure/azure-osconfig/blob/main/src/common/commonutils/CommonUtils.cpp#L12

We will need a full version string in this form: `Azure OSConfig <model version>;<major>.<minor>.<patch>.<yyyymmdd><build>`

robertschaedler3: That form will be the same string as is passed from the platform to the modules during `MmiOpen()` at runtime.

AhmedBM: Good catch!

MariusNi (approved):
:shipit:

PR #399, opened 2022-09-26
Fix payload size check in HostName module

simonjaeger:
## Description

Fixed a check that failed in the *HostName* module on *MmiSet* when a session was created with a maximum payload size of zero. Corrected by skipping the check if the maximum payload size is zero and created a test to verify fix. 

## Checklist

### DIFF
--- a/src/modules/hostname/tests/HostNameBaseTests.cpp
+++ b/src/modules/hostname/tests/HostNameBaseTests.cpp
@@ -118,6 +118,27 @@ namespace OSConfig::Platform::Tests
         delete payload;
     }
 
+    TEST(HostNameBaseTests, GetNameWithZeroPayloadByteSize)
+    {
+        const std::map<std::string, std::string> textResults =
+            {
+                {"cat /etc/hostname", "device"},
+            };
+
+        MMI_JSON_STRING payload;
+        int payloadSizeBytes;
+
+        HostNameBaseTests testModule(textResults, 0);
+        int status = testModule.Get(&testModule, HostNameBase::m_componentName, HostNameBase::m_propertyName, &payload, &payloadSizeBytes);
+
+        std::string result(payload, payloadSizeBytes);
+
+        EXPECT_EQ(status, MMI_OK);
+        EXPECT_STREQ(result.c_str(), "\"device\"");
+
+        delete payload;
+    }
+
     TEST(HostNameBaseTests, GetHosts)
     {
         const std::map<std::string, std::string> textResults =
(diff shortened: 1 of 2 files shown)


MariusNi (approved):
:shipit:

src/modules/hostname/tests/HostNameBaseTests.cpp
```diff
@@ -118,6 +118,27 @@ namespace OSConfig::Platform::Tests
...
+
+        HostNameBaseTests testModule(textResults, 0);
+        int status = testModule.Get(&testModule, HostNameBase::m_componentName, HostNameBase::m_propertyName, &payload, &payloadSizeBytes);
+
+        std::string result(payload, payloadSizeBytes);
+
+        EXPECT_EQ(status, MMI_OK);
+        EXPECT_STREQ(result.c_str(), "\"device\"");
+
+        delete payload;
```
robertschaedler3, line 139: nit: use `FREE_MEMORY()` macro from common utils for consistency

robertschaedler3 (dismissed):
👍🏻

src/modules/hostname/tests/HostNameBaseTests.cpp
simonjaeger: Good point, thanks! Fixed.

src/modules/hostname/tests/HostNameBaseTests.cpp
```diff
@@ -118,6 +118,27 @@ namespace OSConfig::Platform::Tests
         delete payload;
```
MariusNi, line 118: >delete payload; [](http://example.com/codeflow?start=8&length=15)

Apply same as on line 139 everywhere else now.

MariusNi:
>         int payloadSizeBytes;

nit, for consistency with rest of project. always explicitly initialize variables. Here:

MMI_JSON_STRING payload = nullptr;
int payloadSizeBytes = 0;

---
Refers to: src/modules/hostname/tests/HostNameBaseTests.cpp:157 in bffd8a5. [](commit_id = bffd8a5ae8f2546bc46370c01f3dc526611fc25a, deletion_comment = False)

MariusNi:
>         int payloadSizeBytes;

In all instances in this file, please.

---
In reply to: [1258619152](https://github.com/Azure/azure-osconfig/pull/399#issuecomment-1258619152) [](http://example.com/codeflow?ancestors=1258619152)

---
Refers to: src/modules/hostname/tests/HostNameBaseTests.cpp:157 in bffd8a5. [](commit_id = bffd8a5ae8f2546bc46370c01f3dc526611fc25a, deletion_comment = False)

src/modules/hostname/tests/HostNameBaseTests.cpp
simonjaeger: Good point, thanks!

PR #412, opened 2022-10-05
New OSConfig Configuration module

MariusNi:
May still need bit of fit and finish aka stabilization but here it is so we can also test the CI and see where it cracks.


dtmi/osconfig/configuration-1.json
```diff
@@ -0,0 +1,109 @@
...
+        "@type": "Enum",
+        "valueSchema": "integer",
+        "enumValues": [
+          {
+            "name": "auto",
+            "enumValue": 0
+          },
+          { 
+            "name": "mqtt",
+            "enumValue": 1
```
simonjaeger, line 98: >"enumValue": 1 [](http://example.com/codeflow?start=12&length=14)

Should this DTMI property be different (enum names/values) than the reported MIM object?

src/common/commonutils/ConfigUtils.c
```diff
@@ -84,35 +89,44 @@ static int GetIntegerFromJsonConfig(const char* valueName, const char* jsonStrin
                 if (0 == valueToReturn)
                 {
                     valueToReturn = defaultValue;
-                    OsConfigLogInfo(log, "GetIntegerFromJsonConfig: %s value not found or 0, using default (%d)", valueName, defaultValue);
+                    if (IsFullLoggingEnabled())
+                    {
+                        OsConfigLogInfo(log, "GetIntegerFromJsonConfig: %s value not found or 0, using default (%d)", valueName, defaultValue);
```
simonjaeger, line 94: >OsConfigLogInfo [](http://example.com/codeflow?start=24&length=15)

nit: Is *Info* the right log level for this? Others are *Error*.

src/modules/configuration/src/lib/Configuration.c
```diff
@@ -0,0 +1,545 @@
...
+    }
+}
+
+int ConfigurationMmiGetInfo(const char* clientName, MMI_JSON_STRING* payload, int* payloadSizeBytes)
+{
+    int status = EINVAL;
+
+    if ((NULL == payload) || (NULL == payloadSizeBytes))
+    {
+        return status;
```
simonjaeger, line 287: >return status; [](http://example.com/codeflow?start=8&length=14)

Would it make sense to log this error?

src/modules/configuration/src/lib/Configuration.c
```diff
@@ -0,0 +1,545 @@
...
+    *payload = NULL;
+    *payloadSizeBytes = 0;
+
+    if (!IsValidSession(clientSession))
+    {
+        OsConfigLogError(ConfigurationGetLog(), "MmiGet(%s, %s) called outside of a valid session", componentName, objectName);
+        status = EINVAL;
+    }
+    
+    if ((MMI_OK == status) && (strcmp(componentName, g_configurationComponentName)))
```
simonjaeger, line 336: >strcmp [](http://example.com/codeflow?start=31&length=6)

Would it be clearer if this read as `0 != strcmp`?

src/modules/configuration/src/lib/Configuration.c
```diff
@@ -0,0 +1,545 @@
...
+        OsConfigLogError(ConfigurationGetLog(), "Failed to apply new configuration: %s", IsFullLoggingEnabled() ? newConfiguration : "-");
+    }
+        
+    if (jsonValue)
+    {
+        json_value_free(jsonValue);
+    }
+
+    FREE_MEMORY(existingConfiguration);
+    FREE_MEMORY(newConfiguration);
```
simonjaeger, line 249: >FREE_MEMORY(newConfiguration); [](http://example.com/codeflow?start=4&length=30)

Is it preferred to free this using `json_free_serialized_string` instead? As it was created with `json_serialize_to_string_pretty`.

src/modules/configuration/src/lib/Configuration.c
```diff
@@ -0,0 +1,545 @@
...
+        }
+
+        if (MMI_OK == status)
+        {
+            if (0 != (modelVersion = (int)json_object_get_number(jsonObject, g_modelVersionObject)))
+            {
+                g_modelVersion = modelVersion;
+            }
+
+            if (0 != (refreshInterval = (int)json_object_get_number(jsonObject, g_refreshIntervalObject)))
```
robertschaedler3, line 496: Should there be a check with `json_object_get_type()` before each `json_object_get_xyz()` ?

src/modules/configuration/tests/ConfigurationTests.cpp
```diff
@@ -0,0 +1,381 @@
...
+
+#include <gtest/gtest.h>
+#include <version.h>
+#include <Mmi.h>
+#include <CommonUtils.h>
+#include <Configuration.h>
+
+using namespace std;
+
+class ConfigurationTest : public ::testing::Test
```
robertschaedler3, line 12: Should unit tests be placed in `namespace Tests { ... }` like other modules?

dtmi/osconfig/configuration-1.json
```diff
@@ -0,0 +1,109 @@
...
+                  "enumValue": 2
+                }
+              ]
+            }
+          }
+        ]
+      },
+      "writable": true
+    },
+    {
```
robertschaedler3, line 55: Why not use an object for all reported properties (here and in MIM)? 

This would significantly cut down on the number of `MmiGet` calls to the module (configuration may grow larger in the future). Also, could allow for the reported object to match 1:1 with `osconfig.json`.

src/common/commonutils/ConfigUtils.c
MariusNi: No, because this also happens when the value is not found - and that's not an error. Notice this trace is not newly introduced now.

src/modules/configuration/src/lib/Configuration.c
MariusNi: Why? Would a != combined with a == make things clearer? Both checks are positive here and code looks clear to me.

dtmi/osconfig/configuration-1.json
MariusNi: This MIM was reviewed and signed off earlier in team. 

Reason is that we want to use flatten models where it is possible. The configuration settings can be queried separately on their own fine. However when applying new settings, osconfig needs to be restarted. We do not want if somebody enables full and command logging to restart osconfig twice, for example. Also, some of the settings can be grouped, if not now, in the future (the problem that desired have for HostName, Ztsi, etc).

src/modules/configuration/tests/ConfigurationTests.cpp
MariusNi: The Test namespace is used in many places, including for several modules test:  

src\common\tests\CommonUtilsUT.cpp(20):class CommonUtilsTest : public ::testing::Test
src\modules\adhs\tests\AdhsTests.cpp(12):class AdhsTest : public ::testing::Test
src\modules\commandrunner\tests\CommandRunnerTests.cpp(14):    class CommandRunnerTests : public ::testing::Test
src\modules\configuration\tests\ConfigurationTests.cpp(12):class ConfigurationTest : public ::testing::Test
src\modules\deliveryoptimization\tests\DeliveryOptimizationTests.cpp(12):class DeliveryOptimizationTest : public ::testing::Test
src\modules\deviceinfo\tests\DeviceInfoTests.cpp(12):class DeviceInfoTest : public ::testing::Test
src\modules\firewall\tests\FirewallTests.cpp(126):    class FirewallTests : public ::testing::Test
src\modules\test\RecipeInvoker.h(12):class RecipeFixture : public ::testing::Test {};
src\modules\tpm\tests\TpmTests.cpp(57):    class TpmTests : public ::testing::Test
src\platform\tests\ManagementModuleTests.cpp(15):    class ManagementModuleTests : public ::testing::Test
src\platform\tests\ModulesManagerTests.cpp(26):    class ModuleManagerTests : public ::testing::Test
src\platform\tests\MpiServerTests.cpp(14):    class MpiServerTests : public ::testing::Test {};
src\platform\tests\MpiTests.cpp(18):    class MpiTests : public ::testing::Test

src/modules/configuration/src/lib/Configuration.c
```diff
@@ -0,0 +1,545 @@
...
+        status = EINVAL;
+    }
+    
+    if ((MMI_OK == status) && (strcmp(componentName, g_configurationComponentName)))
+    {
+        OsConfigLogError(ConfigurationGetLog(), "MmiGet called for an unsupported component name (%s)", componentName);
+        status = EINVAL;
+    }
+    
+    if ((MMI_OK == status) && (NULL == (configuration = LoadConfigurationFromFile(g_configurationFile))))
```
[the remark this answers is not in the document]
MariusNi: Prefer to keep these two separate in this case. If there would be a larger number of checks, it may make sense to group using that pattern. In this case choice is not to.

src/modules/configuration/tests/ConfigurationTests.cpp
MariusNi: Let's discuss this offline, we'll want to uniformize the namespace for all tests. My vote is for "Test" as this is one namespace, not many.

src/modules/configuration/src/lib/Configuration.c
MariusNi: Yes.

dtmi/osconfig/configuration-1.json
simonjaeger: Current MIM for iotHubProtocol (reported): 

```JSON
"enumValues": [
    {
        "name": "mqtt",
        "enumValue": 0
    },
    {
        "name": "mqttWebSocket",
        "enumValue": 1
    }
]
```

Current DTMI for iotHubProtocol (reported):
```JSON
"enumValues": [
    {
        "name": "auto",
        "enumValue": 0
    },
    {
        "name": "mqtt",
        "enumValue": 1
    },
    {
        "name": "mqttWebSocket",
        "enumValue": 2
    }
]
```

For example: enumValue 0 means something different in MIM (**mqtt**) than in DTMI (**auto**). Is it accepted that this differs between MIM and DTMI?

src/common/commonutils/ConfigUtils.c
simonjaeger: Understood, thanks for the context!

dtmi/osconfig/configuration-1.json
MariusNi: Yes, correct. This will be fixed, the MIM also needs to have same enums as in the PnP interface. Auto needs to be available for desired as well as it is an actual setting value in configuration JSON.

src/modules/configuration/src/lib/Configuration.c
simonjaeger: Agreed on clarity. Would it be more consistent with rest of code to use `0 != strcmp`, however?

src/modules/configuration/src/lib/Configuration.c
simonjaeger: Makes sense, thanks!

simonjaeger (dismissed):
LGTM

robertschaedler3 (approved):

src/modules/configuration/src/lib/Configuration.c
```diff
@@ -0,0 +1,547 @@
...
+    int refreshInterval = g_refreshInterval;
+    int localManagementEnabled = g_localManagementEnabled ? 1 : 0;
+    int fullLoggingEnabled = g_fullLoggingEnabled ? 1 : 0;
+    int commandLoggingEnabled = g_commandLoggingEnabled ? 1 : 0;
+    int iotHubProtocol = g_iotHubProtocol;
+
+    JSON_Value* jsonValue = NULL;
+    JSON_Object* jsonObject = NULL;
+
+    if ((NULL == componentName) || (NULL == objectName) || (NULL == payload) || (0 >= payloadSizeBytes))
```
AhmedBM, line 434: payloadSizeBytes = 0 should be valid (no limit)

src/modules/configuration/src/lib/Configuration.c
```diff
@@ -0,0 +1,547 @@
...
+
+    if ((MMI_OK == status) && (strcmp(objectName, g_desiredConfigurationObject)))
+    {
+        OsConfigLogError(ConfigurationGetLog(), "MmiSet called for an unsupported object name (%s)", objectName);
+        status = EINVAL;
+    }
+
+    if (MMI_OK == status)
+    {
+        if (NULL != (payloadString = malloc(payloadSizeBytes + 1)))
```
AhmedBM, line 461: +1 what? Shouldn't we not return null terminators?

src/modules/configuration/tests/ConfigurationTests.cpp
```diff
@@ -0,0 +1,381 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include <gtest/gtest.h>
```
AhmedBM, line 4: nit: Alphabetically sort headers

src/modules/configuration/src/lib/Configuration.c
MariusNi: MmiSet with payloadSizeBytes (aka no payload) is not valid. This is not the parameter for MmiOpen, this is MmiSet.

src/modules/configuration/src/lib/Configuration.c
MariusNi: Allocates 1 more for the terminator but returns without that last byte.

src/modules/configuration/tests/ConfigurationTests.cpp
MariusNi: That is not a rule and sometimes it could even break things (some headers require to be included in certain order).

PR #424, opened 2022-10-14
Fixing payload validation for ModulesTest app

AhmedBM:
## Description

* Fixing the recipe payload validation to use standard string comparison instead of parsons `json_value_equals` which fails on occasion despite the JSON strings being equivalent.

## Checklist

### DIFF
--- a/src/modules/test/RecipeInvoker.cpp
+++ b/src/modules/test/RecipeInvoker.cpp
@@ -32,14 +32,12 @@ void RecipeInvoker::TestBody()
                 EXPECT_NE(0, m_recipe.m_mimObjects->at(m_recipe.m_componentName)->size()) << "No MimObjects for " << m_recipe.m_componentName << "!";
                 auto map = m_recipe.m_mimObjects->at(m_recipe.m_componentName)->at(m_recipe.m_objectName);
 
-                std::string payloadStr(payload, payloadSize);
-
                 // Validate settings + supported values
                 TestLogInfo("Validating settings and supported values for '%s'", m_recipe.m_objectName.c_str());
                 if ((map.m_type.compare("array") == 0) ||
                     (map.m_type.compare("map") == 0))
                 {
-                    EXPECT_EQ(JSONArray, json_value_get_type(root_value)) << "Expecting '" << m_recipe.m_objectName << "' to contain an array" << std::endl << "JSON: " << payloadStr;
+                    EXPECT_EQ(JSONArray, json_value_get_type(root_value)) << "Expecting '" << m_recipe.m_objectName << "' to contain an array" << std::endl << "JSON: " << payloadString;
                 }
                 else
                 {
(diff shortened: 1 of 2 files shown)


src/modules/test/RecipeInvoker.cpp
```diff
@@ -82,11 +80,10 @@ void RecipeInvoker::TestBody()
...
-                EXPECT_NE(nullptr, returned_payload) << "Failed to parse returned payload" << std::endl << "JSON: " << recipe_payload_str.c_str() << std::endl;
+                EXPECT_NE(nullptr, returned_payload) << "Failed to parse returned payload" << std::endl << "JSON: " << payloadString.c_str() << std::endl;
                 EXPECT_EQ(json_value_get_type(returned_payload), json_value_get_type(recipe_payload)) << "Non matching payload types. Recipe payload: " << json_value_get_type(recipe_payload) << ", returned payload: " << json_value_get_type(returned_payload) << std::endl;
-                EXPECT_EQ(json_value_equals(recipe_payload, returned_payload), 1) << "Non matching recipe payload" << std::endl << "Recipe   payload: " << m_recipe.m_payload << std::endl << "Returned payload: " << recipe_payload_str.c_str() << std::endl;
+                EXPECT_STRCASEEQ(m_recipe.m_payload.c_str(), payloadString.c_str()) << "Non matching recipe payload" << std::endl << "Recipe   payload: " << m_recipe.m_payload << std::endl << "Returned payload: " << payloadString.c_str() << std::endl;
```
robertschaedler3, line 86: Will this also work for objects? What if two objects are equal but serialized differently (should be equal but different serialization means they will not be equal)?

robertschaedler3 (changes requested):

src/modules/test/RecipeInvoker.cpp
MariusNi: It's all strings. It'll work for object payloads as long as expected will be in same order and format with actual

AhmedBM: yeah, these will be known strings so it will not be an issue

AhmedBM: In general I agree that a json object shouldn't care about member ordering, but for the recipes, we should note that we expect the object to be ordered just as the module will be serialized. Given that the recipes are authored by the module authors themselves this shouldn't be an issue, the expected json shouldn't be a surprise.

MariusNi: Spaces vs. no spaces will break this comparison, even if same order and names

robertschaedler3: These should be compared as JSON values, it will be very difficult to debug and identify failures caused by small serialization differences in large JSON objects (difference of spacing, ordering, etc).

MariusNi: Spaces before/after elements in JSON can break this comparison even with same ordering

MariusNi: Example:

"{\"Name\" :  \"Configuration\","
    "\"Description\" :  \"Provides functionality to manage OSConfig configuration on device\","....

vs

"{\"Name\": \"Configuration\","
    "\"Description\": \"Provides functionality to manage OSConfig configuration on device\","

MariusNi: Also lines with EOLs vs not.

src/modules/test/RecipeInvoker.cpp
```diff
@@ -80,13 +78,12 @@ void RecipeInvoker::TestBody()
...
-                EXPECT_NE(nullptr, returned_payload) << "Failed to parse returned payload" << std::endl << "JSON: " << recipe_payload_str.c_str() << std::endl;
+                EXPECT_NE(nullptr, returned_payload) << "Failed to parse returned payload" << std::endl << "JSON: " << payloadString.c_str() << std::endl;
                 EXPECT_EQ(json_value_get_type(returned_payload), json_value_get_type(recipe_payload)) << "Non matching payload types. Recipe payload: " << json_value_get_type(recipe_payload) << ", returned payload: " << json_value_get_type(returned_payload) << std::endl;
-                EXPECT_EQ(json_value_equals(recipe_payload, returned_payload), 1) << "Non matching recipe payload" << std::endl << "Recipe   payload: " << m_recipe.m_payload << std::endl << "Returned payload: " << recipe_payload_str.c_str() << std::endl;
+                EXPECT_EQ(json_value_equals(recipe_payload, returned_payload), 1) << "Non matching recipe payload" << std::endl << "Recipe   payload: " << m_recipe.m_payload << std::endl << "Returned payload: " << payloadString.c_str() << std::endl;
```
robertschaedler3, line 86: If the parsing fails (`returned_payload` and/or `recipe_payload` is null), what is the expected behavior of line 85, 86? I believe passing null to `json_value_get_type` or `json_value_equals` is undefined behavior for parson (could cause crashes).

src/modules/test/RecipeInvoker.cpp
```diff
@@ -80,13 +78,12 @@ void RecipeInvoker::TestBody()
             if (m_recipe.m_payload.size())
             {
                 JSON_Value *recipe_payload = json_parse_string(m_recipe.m_payload.c_str());
-                JSON_Value *returned_payload = json_parse_string(payload);
+                JSON_Value *returned_payload = json_parse_string(payloadString.c_str());
 
-                std::string recipe_payload_str(payload, payloadSize);
                 EXPECT_NE(nullptr, recipe_payload) << "Failed to parse recipe payload" << std::endl << "JSON: " << m_recipe.m_payload.c_str() << std::endl;
-                EXPECT_NE(nullptr, returned_payload) << "Failed to parse returned payload" << std::endl << "JSON: " << recipe_payload_str.c_str() << std::endl;
+                EXPECT_NE(nullptr, returned_payload) << "Failed to parse returned payload" << std::endl << "JSON: " << payloadString.c_str() << std::endl;
```
robertschaedler3, line 84: Similar to above validation above, are we validating JSON objects against their respective MIM type?

src/modules/test/RecipeInvoker.cpp
AhmedBM: neither should crash but if the above fails, 85, 86 will also fail

src/modules/test/RecipeInvoker.cpp
AhmedBM: yes, but this happens before the JSON validation, this code is just responsible for the JSON comparison

MariusNi (approved):
:shipit:

PR #436, opened 2022-11-04
Refactoring the test recipes with explicit setup and teardown and running commands without CommandRunner

MariusNi:
## Description

- Each test recipe gets its first test object for setup of the module under test. Teardown is implied at the end of the recipe.
- CommandRunner invocation from other modules' test recipes is replaced with direct commands to be run by ModulesTest using commonutils.
- Test recipes using 4 spaces per tab are corrected to 2 spaces per tab.

## Checklist

### DIFF
--- a/src/modules/test/TestRecipeParser.cpp
+++ b/src/modules/test/TestRecipeParser.cpp
@@ -31,7 +31,7 @@ TestRecipes TestRecipeParser::ParseTestRecipe(std::string path)
 
     JSON_Array *jsonTestRecipes = json_value_get_array(root_value);
     JSON_Object *jsonTestRecipe = nullptr;
-    for (size_t i = 0; i < json_array_get_count(jsonTestRecipes); i++)
+    for (size_t i = 1; i < json_array_get_count(jsonTestRecipes); i++)
     {
         jsonTestRecipe = json_array_get_object(jsonTestRecipes, i);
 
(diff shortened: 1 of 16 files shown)


robertschaedler3 (dismissed):

AhmedBM:
This does break all the tests with the added unexpected object at the beginning of the test recipes. I would recommend merging this in only after we've added the changes to support these in `modulestest`

```
[==========] Running 0 tests from 0 test suites.
[==========] 0 tests from 0 test suites ran. (0 ms total)
[  PASSED  ] 0 tests.
```

MariusNi:
Yes, but that's PASSED at the end. The tests are not broken/failed - they pass by skipping the contents, which is another problem by itself (ModulesTest should not skip a test recipe when it hits an unsupported test object w/o flagging that as a failure).
  

---
In reply to: [1304179713](https://github.com/Azure/azure-osconfig/pull/436#issuecomment-1304179713) [](http://example.com/codeflow?ancestors=1304179713,1304179713)

AhmedBM:
agreed that this is a problem, im just suggesting we make the recipe changes after the `modulestest` changes as this will regress the current tests. The CI for modulestest is disabled so it isn't blocking PRs. But this will break them for anyone trying to validate their changes locally. I say break, but i just mean it no longer runs the tests (0 tests)

MariusNi:
It's fixed in last iteration, please take a look - ModulesTest now skips the first text object as a temp workaround

---
In reply to: [1304216516](https://github.com/Azure/azure-osconfig/pull/436#issuecomment-1304216516) [](http://example.com/codeflow?ancestors=1304216516,1304216516)

AhmedBM (approved):
👌

robertschaedler3 (approved):
👍🏻

PR #441, opened 2022-12-02
Add support for Debian 10/11

robertschaedler3:
## Description

- Added Dockerfiles for Debian 10 (buster) + 11 (bullseye) 
- Updated packaging/signing pipelines to build new packages
- Remove old/unused variables and jobs from pipelines and updated variables
- Set `DBUILD_TESTS=OFF` during package builds to improve performance

## Checklist


devops/pipeline/create_containers.yaml
```diff
@@ -39,6 +41,32 @@ parameters:
   type: boolean
   default: false
```
MariusNi, line 42: Is it by design that Debian 9 is left and not removed?

src/modules/CMakeLists.txt
```diff
@@ -32,20 +34,23 @@ function(add_module_custom_target moduleDir targetName mimName recipeName)
             add_subdirectory(${_moduleDir})
         elseif(EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/test/recipes/${recipeName}Tests.json)
             add_subdirectory(${_moduleDir})
-            add_dependencies(modulerecipeconfiguration ${_targetName})
+            if (BUILD_TESTS)
```
MariusNi, line 37: > if (BUILD_TESTS) [](http://example.com/codeflow?start=11&length=17)

Thanks for adding this! Shouldn't this be even higher, for example to cover line 33 (JQ)?

MariusNi: Actually, the entire block of testing modules, all from line 13 down to line 52

devops/docker/debian10-arm/Dockerfile
```diff
@@ -0,0 +1,34 @@
+FROM multiarch/qemu-user-static:x86_64-arm as qemu
+FROM arm32v7/debian:buster
```
AhmedBM, line 2: same, use secure mirror `mcr.microsoft.com/mirror/docker/library/debian:buster`

devops/docker/debian10-arm64/Dockerfile
```diff
@@ -0,0 +1,34 @@
+FROM multiarch/qemu-user-static:x86_64-aarch64 as qemu
+FROM arm64v8/debian:buster
```
AhmedBM, line 2: use secure mirror `mcr.microsoft.com/mirror/docker/library/debian:buster`

devops/docker/debian11-amd64/Dockerfile
```diff
@@ -0,0 +1,33 @@
+FROM debian:bullseye
```
AhmedBM, line 1: Use secure mirror `mcr.microsoft.com/mirror/docker/library/debian:bullseye`

devops/docker/debian11-arm/Dockerfile
```diff
@@ -0,0 +1,35 @@
+FROM multiarch/qemu-user-static:x86_64-arm as qemu
+FROM arm32v7/debian:bullseye
```
AhmedBM, line 2: Use secure mirror `mcr.microsoft.com/mirror/docker/library/debian:bullseye`

devops/docker/debian11-arm64/Dockerfile
```diff
@@ -0,0 +1,35 @@
+FROM multiarch/qemu-user-static:x86_64-aarch64 as qemu
+FROM arm64v8/debian:bullseye
```
AhmedBM, line 2: Use secure mirror `mcr.microsoft.com/mirror/docker/library/debian:bullseye`

devops/pipeline/binary_package.yaml
```diff
@@ -8,119 +8,116 @@ name: $(MajorVersion).$(MinorVersion).$(PatchVersion).$(Date:yyyyMMdd)$(Rev:rr)
...
-        performCoverage: false
-        ignoreTestFailures: false
-        publishArtifact: true
+        displayName: debian11_arm64
 
 variables:
   SERVICE_CONNECTION: OSConfig-ACR
   CONTAINER_REGISTRY: osconfig.azurecr.io
-  CONTAINER_TAG: 289651
+  CONTAINER_TAG: latest
```
AhmedBM, line 94: We should stick to a specific tag and not always use `latest`

AhmedBM (changes requested):
We must make use of the secure mcr mirrors as noted

devops/pipeline/binary_package.yaml
MariusNi: Can you please elaborate why ("we should stick to a specific tag")?

robertschaedler3: The issue with a "specific" tag is that it required us to build all (now 15) containers every time we need to make an update to any of the Dockerfiles. Latest will also let us "roll-back" if there is an issue with a new image that is built.

devops/pipeline/create_containers.yaml
robertschaedler3: Yes, once we "officially" determine that Debian 10/11 are ready (passed all validation etc.) we will remove support from Debian 9 everywhere in a separate PR.

devops/docker/debian10-amd64/Dockerfile
```diff
@@ -0,0 +1,32 @@
+FROM debian:buster
+ARG DEBIAN_FRONTEND=noninteractive
+RUN apt -y update && apt-get -y install software-properties-common
```
simonjaeger, line 3: nit: Any reason why *apt* over *apt-get* is used? Generally speaking, *apt* is intended for human end-users. As long as the Dockerfile only does the above, it is fine. Otherwise, it may be a good idea to use *apt-get* everywhere instead.

devops/docker/debian10-amd64/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+RUN apt -y update && apt-get -y install \
```
simonjaeger, line 4: nit: Is the package sources list not already updated from the previous line?

devops/pipeline/binary_package.yaml
AhmedBM: > Can you please elaborate why ("we should stick to a specific tag")?

This is to have explicit control over the container version. Although this was more of a problem in the past where containers were updated without our CD workflows. Now this is less of a problem given we have CD for the creating the containers we just automagically pick up the latest.

> The issue with a "specific" tag is that it required us to build all (now 15) containers every time we need to make an update to any of the Dockerfiles. Latest will also let us "roll-back" if there is an issue with a new image that is built.

Yeah i get the convenience. This was more of an issue in the past.

devops/docker/debian10-amd64/Dockerfile
AhmedBM: I somehow thought i already made those changes to use apt over _apt-get/apt-cache/apt-*_. Although Simon seems to be suggesting the opposite (apt-* over apt). I wasn't aware of any functional differences between apt and apt-* but i definitely agree that we should stick to either of them and not mix-and-match as this can lead to issues when building the images.

simonjaeger: Indeed, for the above it will not have any major differences. Typically, *apt* adds things such as progress bars can affect the logs (rarely). I also found that it removes noise in the logs that is not useful for CI (but nice for humans). Overall, this is nitpicking, most important is that we standardize on one 😄.

devops/docker/debian11-arm/Dockerfile
robertschaedler3: For this and others: no such mirror exists for arm32/46 images in MCR.

src/modules/CMakeLists.txt
robertschaedler3: I temporarily added it this way so that the current modulestest are not broken, all will change when JQ is removed as a part of the modulestest work.

robertschaedler3: We still need that `add_subdirectory` line for each module, only the dependency and building of the test its self are isolated by the `BUILD_TESTS` flag.

MariusNi: OK! Still JQ can be skipped unless test is enabled.

MariusNi (dismissed):
:shipit:

devops/docker/debian11-arm/Dockerfile
AhmedBM: I see. We should onboard these if they are not present, see https://eng.ms/docs/cloud-ai-platform/azure-core/azure-management-and-platforms/containers-bburns/microsoft-container-registry/microsoft-artifact-registry-also-known-as-microsoft-container-registry-docs/onboardingdocs/newimageonboarding

devops/docker/debian10-arm64/Dockerfile
AhmedBM: We should onboard the arm64 image - see https://eng.ms/docs/cloud-ai-platform/azure-core/azure-management-and-platforms/containers-bburns/microsoft-container-registry/microsoft-artifact-registry-also-known-as-microsoft-container-registry-docs/onboardingdocs/newimageonboarding

devops/docker/debian10-arm/Dockerfile
AhmedBM: We should onboard the arm image - see https://eng.ms/docs/cloud-ai-platform/azure-core/azure-management-and-platforms/containers-bburns/microsoft-container-registry/microsoft-artifact-registry-also-known-as-microsoft-container-registry-docs/onboardingdocs/newimageonboarding

AhmedBM (changes requested):
See comments about onboarding missing containers mirrors to MCR

AhmedBM (approved):
Looks good!

PR #442, opened 2022-12-08
Refactoring ModulesTest

robertschaedler3:
## Description

- Remove dependency on `jq`
- Remove `testplate.json`
- Revise module loading/unloading and setup/teardown
  - Load modules only based on the test definitions
  - Unload modules when the tests are complete
- Add setup/teardown scripts to run before modules are loaded and after they are unloaded
- Execute system commands/scripts between tests
- Establish path resolution hierarchy 
- Refactor test definitions to remove dependencies on `commandrunner` module

## Checklist


src/modules/test/README.md
```diff
@@ -1,60 +1,193 @@
...
+> - Shared-object binaries are placed under `build/modules/bin`
+> - Test definitions are placed under and renamed to match their module respective shared-object binary `build/test/<module>.json`
+
+```bash
+$ moduletest <file>... [--bin <path>] [--verbose] [--help]
+
+# Example
+$ moduletest path/to/module/test.json
+$ moduletest path/to/module/test.json --bin /usr/lib/osconfig
+$ moduletest path/to/module/test1.json path/to/module/test2.json
```
MariusNi, line 21: We discussed to not do this anymore. To have as the only possible command line arguments to modulestest the test recipes.

src/modules/adhs/test.json
```diff
@@ -0,0 +1,51 @@
...
+    },
+    {
+      "type": "desired",
+      "component": "Adhs",
+      "object": "desiredOptIn",
+      "payload": 1
+    },
+    {
+      "run": "grep -q 'Permission = \"Required\"' /etc/azure-device-health-services/config.toml"
+    },
```
MariusNi, line 31: So you are proposing to replace the test recipes with these?

Why aren't the test recipes removed as well?

If we deviate from the agreed path, we need to have the changes proposed and discussed in team.

src/modules/adhs/test.json
```diff
@@ -0,0 +1,51 @@
...
+  "setup": [
+    "mkdir -p /etc/azure-device-health-services/",
+    "touch /etc/azure-device-health-services/config.toml",
+    "cp /etc/azure-device-health-services/config.toml config.toml"
+  ],
+  "teardown": [
+    "cp config.toml /etc/azure-device-health-services/config.toml",
+    "rm -f config.toml"
+  ],
+  "steps": [
```
MariusNi, line 14: This does not appear right to have these test.json in the source code for each module. ModulesTest is a functional test that needs to test module binaries against test recipes. When this test is executed the OSConfig source code may not be present/available.

The contract for ModulesTest is:

1. Its binary must test module SO binaries
2. Against test recipes

src/modules/test/main.c
```diff
@@ -0,0 +1,732 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include <Common.h>
+#include <Module.h>
```
MariusNi, line 5: Can also move Module.h to Common.h.

However, is Module.h really needed? If it's only main.c that includes it, all Module.h contents could be in main.c (and this could simplify the code even more)

src/modules/test/main.c
```diff
@@ -0,0 +1,732 @@
...
+} STEP;
+
+static bool g_verbose = false;
+
+void FreeStep(STEP* step)
+{
+    if (step == NULL)
+    {
+        return;
+    }
```
MariusNi, line 88: nit: simpler to do: if (step) {} then exit

src/modules/test/main.c
```diff
@@ -0,0 +1,732 @@
...
+            free(step->data.module.name);
+            break;
+        case COMMAND:
+            free(step->data.command.arguments);
+            break;
+        case TEST:
+            free(step->data.test.component);
+            free(step->data.test.object);
+            json_value_free(step->data.test.payload);
+            break;
```
MariusNi, line 102: >break; [](http://example.com/codeflow?start=12&length=6)

This last break is not needed. Also, is best to have as the last label "default"

src/modules/test/main.c
```diff
@@ -0,0 +1,732 @@
...
+    JSON_Value* payload = NULL;
+    const char* json = NULL;
+
+    if (NULL == (type = json_object_get_string(object, RECIPE_TYPE)))
+    {
+        LOG_ERROR("Missing '%s' from test step", RECIPE_TYPE);
+        status = EINVAL;
+    }
+
+    if (strcmp(type, RECIPE_REPORTED) == 0)
```
MariusNi, line 119: Add an else here or exit after line 116 or check status on line 119

src/modules/test/main.c
```diff
@@ -0,0 +1,732 @@
...
+        status = EINVAL;
+    }
+
+    if (NULL == (payload = json_object_get_value(object, RECIPE_PAYLOAD)))
+    {
+        if (NULL != (json = json_object_get_string(object, RECIPE_JSON)))
+        {
+            payload = json_parse_string(json);
+        }
+    }
```
MariusNi, line 150: And if not (on missing else cases)? There should be errors logged

src/modules/test/main.c
```diff
@@ -0,0 +1,732 @@
...
+}
+
+int ParseModuleStep(const JSON_Object* object, MODULE_STEP* module)
+{
+    int status = 0;
+    const char* action = NULL;
+
+    if (module == NULL)
+    {
+        LOG_ERROR("Invalid (null) module");
```
MariusNi, line 203: >module [](http://example.com/codeflow?start=34&length=6)

"module name"? (and not invalid, but maybe "no module name (null)"?

src/modules/test/main.c
```diff
@@ -0,0 +1,732 @@
...
+            LOG_ERROR("Invalid action '%s'", action);
+            status = EINVAL;
+        }
+    }
+
+    if ((status == 0) && (module->action == LOAD))
+    {
+        if (NULL == (module->name = strdup(json_object_get_string(object, RECIPE_MODULE))))
+        {
+            LOG_ERROR("'%s' is required for '%s' module step", RECIPE_MODULE, action);
```
MariusNi, line 233: >module step [](http://example.com/codeflow?start=49&length=11)

What's a "module step"? May want to use in logging only terms that are defined in modules.md.

src/modules/test/main.c
```diff
@@ -0,0 +1,732 @@
...
+                {
+                    sleep(step->delay);
+                }
+
+                printf("%d/%d ", i + 1, total);
+
+                switch (step->type)
+                {
+                    case COMMAND:
+                        printf("[COMMAND]\n");
```
MariusNi, line 489: >printf("[COMMAND]\n"); [](http://example.com/codeflow?start=24&length=22)

Use the logging macros, not printf directly. These printf will log nice to console but will not be in the log file. The CI runs may suffer (when re-adding this in future)

src/modules/test/main.c
```diff
@@ -0,0 +1,732 @@
...
+                UnloadModule(module);
+                module = NULL;
+            }
+
+            printf("summary: %s\n", failures == 0 ? "PASSED" : "FAILURE");
+            printf("  passed: %d\n", total - failures);
+            printf("  skipped: %d\n", skipped);
+            printf("  failed: %d\n", failures);
+            printf("  total: %d (%d ms)\n", total, (int)(end - start));
+            printf("\n");
```
MariusNi, line 564: See earlier comment on printf. Apply everywhere.

src/modules/test/recipes/adhs.json
```diff
@@ -0,0 +1,53 @@
+[
+  {
```
MariusNi, line 2: Please keep the original recipe file names, do not change. We need to review the delta, for example, also those names were decided upon.

src/modules/test/main.c
robertschaedler3: I kept it separate to make reusing this code in the platform easier in the future if/when C++ components are converted to C.

src/modules/test/main.c
robertschaedler3: In this case it is not an error if the payload is missing, we need to treat payloads as optional for both comparison and setting "empty" desired payloads as part of the tests.

src/modules/test/main.c
robertschaedler3: There is not log file produced directly by the `moduletest` binary. For CI runs, stdout is redirected to a file:

```
$ ./moduletest blah.json > output.txt
``` 

then that file is uploaded as an artifact.

src/modules/test/main.c
```diff
@@ -0,0 +1,732 @@
...
+    }
+    else
+    {
+        status = EINVAL;
+    }
+
+    if (NULL == (test->component = strdup(json_object_get_string(object, RECIPE_COMPONENT))))
+    {
+        LOG_ERROR("'%s' is required for '%s' test step", RECIPE_COMPONENT, type);
+        status = EINVAL;
```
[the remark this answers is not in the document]
robertschaedler3: The intent for these parsing functions is to show as many parsing errors as possible to the user. instead of stopping execution on the first error when others can still be safely detected.

src/modules/test/Module.h
```diff
@@ -0,0 +1,64 @@
...
+
+    MMI_OPEN open;
+    MMI_CLOSE close;
+    MMI_GETINFO getinfo;
+    MMI_SET set;
+    MMI_GET get;
+    MMI_FREE free;
+} MANAGEMENT_MODULE;
+
+MANAGEMENT_MODULE* LoadModule(const char* client, const char* bin, const char* name);
```
MariusNi, line 61: >bin [](http://example.com/codeflow?start=62&length=3)

Use full words

src/modules/test/Module.c
```diff
@@ -0,0 +1,392 @@
...
+                        break;
+                    }
+                }
+            }
+
+            if (0 == status)
+            {
+                *moduleInfo = info;
+            }
+            else {
```
MariusNi, line 219: >{ [](http://example.com/codeflow?start=17&length=1)

New line, please (look for other instances)

src/modules/test/Module.c
```diff
@@ -0,0 +1,392 @@
...
+    {
+        LOG_ERROR("Invalid (null) bin");
+        status = EINVAL;
+    }
+    else if (NULL == name)
+    {
+        LOG_ERROR("Invalid (null) name");
+        status = EINVAL;
+    }
+    else if (NULL == (module = (MANAGEMENT_MODULE*)calloc(1, sizeof(MANAGEMENT_MODULE))))
```
MariusNi, line 256: >else if (NULL == (module = (MANAGEMENT_MODULE*)calloc(1, sizeof(MANAGEMENT_MODULE)))) [](http://example.com/codeflow?start=4&length=85)

Where is this memory freed?

MariusNi: Make sure caller does

src/modules/test/Module.h
```diff
@@ -0,0 +1,64 @@
...
+typedef struct MANAGEMENT_MODULE
+{
+    char* name;
+    void* handle;
+    MMI_HANDLE session;
+    MODULE_INFO* info;
+
+    MMI_OPEN open;
+    MMI_CLOSE close;
+    MMI_GETINFO getinfo;
```
MariusNi, line 55: >getinfo [](http://example.com/codeflow?start=16&length=7)

getInfo

MariusNi: or: mmiGetInfo (and same for all the rest)

src/modules/test/Module.c
robertschaedler3: It is freed by the caller. Ensured that the caller properly unloads modules on failure and correctly cleans up any allocated memory.

src/modules/test/README.md
MariusNi: Change "test.json" to "ModuleTestRecipe.json" or similar?

src/modules/test/README.md
```diff
@@ -1,60 +1,98 @@
...
+### Module steps
 
-Supports any of the `-gtest_*` flags. See `--help` for additional usage details.
- * `--gtest_list_tests`
- * `--gtest_filter=POSTIVE_PATTERNS[-NEGATIVE_PATTERNS]`
+Module steps are used to load and unload a module as part of the test suite. This can be useful for testing the behavior of a module when it is loaded/unloaded and gives the opportunity to verify the state of the device before/after the module is loaded/unloaded.
 
-# Authoring Test Recipes
-See *Testing* in [docs/modules.md](https://github.com/Azure/azure-osconfig/blob/main/docs/modules.md#13-testing) for more information on authoring Test Recipes (see [src/modules/test/recipes/](https://github.com/Azure/azure-osconfig/tree/main/src/modules/test/recipes) for current recipes).
+> If the unload step is omitted, the module will be unloaded after the test suite is complete.
```
MariusNi, line 85: >will be unloaded after the test suite is complete. [](http://example.com/codeflow?start=44&length=50)

Why not leave it loaded until ModuleTest process exits?

Maybe in the future some module developers would like to see if their module remains loaded and does not crash while other modules are in use.

robertschaedler3: Agreed, fixed the log trace to display a warning to make this behavior more clear.

MariusNi (approved):
:shipit:

MariusNi (approved):
:shipit:

PR #446, opened 2022-12-13
Start systemd-resolved service for DnsServers data collection

hachotze:
## Description

Added logic to check if the systemd-resolved service is running before attempting DnsServers data collection. If the service is stopped, the module will start it. If the service is not available on the system, the module will write an error in the Networking log file and continue.

## Checklist

### DIFF
--- a/src/modules/networking/src/lib/Networking.cpp
+++ b/src/modules/networking/src/lib/Networking.cpp
@@ -31,8 +31,7 @@ const char* g_getIpAddressDetails = "ip addr";
 const char* g_getDefaultGateways = "ip route";
 const char* g_getDnsServers = "systemd-resolve --status";
 
-const char* g_checkSystemdResolvedIsRunning = "systemctl list-units --type=service --state=running | grep systemd-resolved.service";
-const char* g_startSystemdResolved = "systemctl start systemd-resolved.service";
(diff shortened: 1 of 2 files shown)


src/modules/networking/src/lib/Networking.cpp
```diff
@@ -31,6 +31,9 @@ const char* g_getIpAddressDetails = "ip addr";
 const char* g_getDefaultGateways = "ip route";
 const char* g_getDnsServers = "systemd-resolve --status";
 
+const char* g_checkSystemdResolvedIsRunning = "systemctl list-units --type=service --state=running | grep systemd-resolved.service";
+const char* g_startSystemdResolved = "systemctl start systemd-resolved.service";
```
MariusNi, line 35: Don't do this. Instead, use the following commonutils library functions:

bool IsDaemonActive(const char* name, void* log)

And:

bool EnableAndStartDaemon(const char* name, void* log)

MariusNi: You can directly call just the later one, EnableAndStartDaemon.

hachotze: Thanks! This is a great improvement. Addressed.

MariusNi (approved):
:shipit:

PR #515, opened 2023-05-15
Include run attempt in package version tweak

robertschaedler3:
## Description

Fixes package version tweak in `insiders-fast` packages to include the GitHub Actions `run_attempt`.

## Checklist

### DIFF
--- a/.github/workflows/package-build.yml
+++ b/.github/workflows/package-build.yml
@@ -59,17 +59,29 @@ jobs:
           container: azure/azure-osconfig/${{ inputs.os }}-${{ inputs.arch }}
           mount: ${{ github.workspace }}:${{ env.MOUNT }}
 
+      - name: Set tweak
+        id: version
+        run: |
+          # Set the tweak version (YYMMDD + 2 digit run attempt)
+          now=$(date +'%Y%m%d')
+          number=$(printf '%02d' ${{ github.run_attempt }})
+          tweak="$(date +'%Y%m%d')$(printf '%02d' ${{ github.run_attempt }})"
+
+          echo date: $now
+          echo number: $number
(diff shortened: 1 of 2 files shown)


.github/workflows/package-build.yml
```diff
@@ -59,17 +59,29 @@ jobs:
           container: azure/azure-osconfig/${{ inputs.os }}-${{ inputs.arch }}
           mount: ${{ github.workspace }}:${{ env.MOUNT }}
 
+      - name: Set tweak
+        id: version
+        run: |
+          # Set the tweak version (YYMMDD + 2 digit run attempt)
+          now=$(date +'%Y%m%d')
+          number=$(printf '%02d' ${{ github.run_attempt }})
```
MariusNi, line 67: >run_attempt [](http://example.com/codeflow?start=44&length=11)

Would "run_number" as it was before be better? Or maybe "build_number"?

"attempt" suggests retry after failure.

MariusNi (approved):
:shipit:

.github/workflows/package-build.yml
robertschaedler3: Ideally, we should remove this altogether, but we need to replicate the way we were versioning these nightly packages in ADO in the meantime. The `run_attempt` more closely follows what we had in ADO using a build number that reset every day. The issue with `run_number` is that it never resets so we will eventually end up with a very "large" version number once the run number increases to 3+ digits (technically should only be 2 digits based on our version schema anyways).

robertschaedler3: Also, removing this will require a bit of planning. Lets discuss more before our next release (that should be a good time to make this change).

PR #553, opened 2023-09-28
Upgrade to PMC API v4

robertschaedler3:
Upgrade package publishing workflows to use PMC (packages.microsoft.com) API v4. This removes our dependency on the `repoclient` and uses the `mcr.microsoft.com/pmc/pmc-cli` Docker container in an isolated environment.


.github/actions/pmc/action.yml
```diff
@@ -0,0 +1,15 @@
+name: PMC
```
MariusNi, line 1: >PMC [](http://example.com/codeflow?start=6&length=3)

Would be good to use the full name once somewhere if possible (in a comment maybe?) To say what PMC means

MariusNi (dismissed):
:shipit:

.github/workflows/package-publish.yml
```diff
@@ -17,58 +17,34 @@ jobs:
   publish:
     runs-on: [self-hosted, 1ES.Pool=release-pool, 1ES.ImageOverride=ubuntu-22.04]
     environment: ${{ inputs.channel }}
+    env:
+      PMC_CLI_BASE_URL: ${{ secrets.PMC_CLI_BASE_URL }}
```
AhmedBM, line 21: i would suggest using the values directly in the pipeline and just avoid passing all these args in the first place,. We gain GitOps as well!

.github/workflows/package-publish.yml
```diff
@@ -17,58 +17,34 @@ jobs:
...
-          EOF
-          )
-
-          echo "$config" > config.json
-
-      - name: Publish
+      - name: Set target
+        id: target
         env:
           REPO_ID_ubuntu_1804: ${{ secrets.REPO_ID_BIONIC }}
```
AhmedBM, line 49: although this isn't part of your change, would be nice to also have these arguments baked into the pipeline instead of using the secrets, again GitOps and tracking FTW! 😂

AhmedBM (changes requested):
Can we bake in the arguments instead of using secrets?

.github/workflows/package-publish.yml
robertschaedler3: Good point! I'll add as many here as possible so that only the bare minimum are secrets. Thanks!

.github/workflows/package-publish.yml
robertschaedler3: Are we sure these repo IDs are "public" I remember at some point thinking that they were private. If they can be public, I'm all for it!

robertschaedler3: Actually I realized why these are secrets. Sort of a hack to allow us to specify different repo IDs based on the environment (insiders-fast vs prod).

.github/actions/pmc/action.yml
robertschaedler3: Good point! the `action.yml` metadata allows us to add a top-level `description` of the custom action. I will add one explaining that this is for packages.microsoft.com.

.github/workflows/package-publish.yml
robertschaedler3: Changed the repo IDs to [GitHub Environment Variables](https://docs.github.com/en/actions/learn-github-actions/contexts#vars-context) 👍

AhmedBM (approved):
👌

PR #557, opened 2023-10-04
Add --id-only flag to PMC upload command

robertschaedler3:
Fixes error in the publishing workflow during the update step. The PMC CLI expects a list of IDs during the final update step when packages are associated with the repo. Packages are then published separately in the final step.


MariusNi (approved):
:shipit:

PR #563, opened 2023-10-12
Add Ubuntu 22.04 and Debian 12 build containers

robertschaedler3:
Add Ubuntu 22.04 (`jammy`) and Debian 12 (`bookworm`) build containers (amd64 and arm64). 

_Workflow changes will follow after the containers are published to `ghcr.io`._


MariusNi (approved):
:shipit:

PR #569, opened 2023-10-21
Add Rust sample module

robertschaedler3:
Adds support for building modules in Rust:

- `#[osc_module]` macro for implementing the MMI and [Foreign Function Interface (FFI)](https://doc.rust-lang.org/nomicon/ffi.html) type conversions
  - Shared logging implementation for using `log::*` macros in modules with our existing log format
  - Module info types (`Lifetime`, `Version`, `UserAccount`)
- Integration into CMake builds (via [corrosion](https://github.com/corrosion-rs/corrosion))

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -7,7 +7,8 @@
 	},
 	// Features to add to the dev container. More info: https://containers.dev/features.
 	"features": {
-		"ghcr.io/devcontainers/features/github-cli:1": {}
+		"ghcr.io/devcontainers/features/github-cli:1": {},
+		"ghcr.io/devcontainers/features/rust:1": {}
 	},
 	"postCreateCommand": "git submodule update --init",
 	"customizations": {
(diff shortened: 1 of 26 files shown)


src/modules/samples/CMakeLists.txt
```diff
@@ -1,4 +1,5 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 
-add_subdirectory(cpp)
\ No newline at end of file
+add_subdirectory(cpp)
+add_subdirectory(rust)
```
MariusNi, line 5: >rust [](http://example.com/codeflow?start=17&length=4)

All modules samples here must follow the pattern show in this C++ sample. Basically showing how to implement all possible MIM cases in a module. If the C++ sample is very useful like this (we got such feedback), the Rust one will be seen significantly a lot more useful for developers.

src/modules/samples/CMakeLists.txt
```diff
@@ -1,4 +1,5 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 
-add_subdirectory(cpp)
\ No newline at end of file
+add_subdirectory(cpp)
```
MariusNi, line 4: We also need the sample to be published in pure C language. 

This will also allow us to compare 1:1 with Rust.

PR #570, opened 2023-10-24
Add C# to CodeQL workflow

robertschaedler3:
Add C# to CodeQL workflow for scanning E2E tests.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -20,15 +20,18 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        language: [cpp]
         # CodeQL language support https://aka.ms/codeql-docs/language-support
+        language: [cpp, csharp]
 
     steps:
-      - name: Checkout repository
-        uses: actions/checkout@v3
+      - uses: actions/checkout@v3
+        if: matrix.language == 'cpp'
         with:
           submodules: recursive
           clean: true
+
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql.yml
```diff
@@ -20,31 +20,47 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        language: [cpp]
         # CodeQL language support https://aka.ms/codeql-docs/language-support
+        language: [cpp, csharp]
```
MariusNi, line 24: >language: [cpp, csharp] [](http://example.com/codeflow?start=8&length=23)

Can we add C or at least prevent C source be scanned as C++? I expect no to both but just in case...

MariusNi (approved):
:shipit:

.github/workflows/codeql.yml
robertschaedler3: C/C++ (`C89`, `C99`, `C11`, `C17`, `C++98`, `C++03`, `C++11`, `C++14`, `C++17`, `C++20`) are all [supported](https://codeql.github.com/docs/codeql-overview/supported-languages-and-frameworks/) under the same scanner using the [`cpp` identifier](https://docs.github.com/en/code-security/code-scanning/creating-an-advanced-setup-for-code-scanning/customizing-your-advanced-setup-for-code-scanning#changing-the-languages-that-are-analyzed)

PR #571, opened 2023-11-07
Move ESRP signing script into custom action

robertschaedler3:
Move ESRP signing script into custom action to remove non-production languages from code scans


adteran (approved):

AhmedBM (approved):
👍

PR #573, opened 2023-11-10
Update CODEOWNERS

robertschaedler3:
Update CODEOWNERS to include renamed team.

## Checklist

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-* @Azure/azure-osconfig-core @MariusNi @robertschaedler3
+* @Azure/azure-osconfig-core
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-* @Azure/azure-osconfig-core
+* @Azure/osconfig-maintainers


AhmedBM (approved):

simonjaeger (approved):

.github/CODEOWNERS
```diff
@@ -1 +1 @@
-* @Azure/azure-osconfig-core @MariusNi @robertschaedler3
```
MariusNi, line 1: >@MariusNi [](http://example.com/codeflow?start=29&length=9)

Put MariusNi back please

MariusNi (dismissed):
:clock1:

.github/CODEOWNERS
```diff
@@ -1 +1 @@
...
+* @Azure/osconfig-maintainers
```
MariusNi, line 1: >@Azure/osconfig-maintainers [](http://example.com/codeflow?start=2&length=27)

As discussed, add explicitly MariusNi after the team name.

MariusNi, line 1: >osconfig-maintainer [](http://example.com/codeflow?start=9&length=19)

Why not naming this 'osconfig-core' as it was, rather than osconfig-maintainers which does not truly reflect the scope of this group?

How about:

@Azure/osconfig-core @MariusNi @robertschaedler3

So that way the two real maintainers are always explicitly listed for every PR. Or, if you want to skip yours:

@Azure/osconfig-core @MariusNi

robertschaedler3: The `osconfig-maintainers` name correlates to the GitHub role assignment given to the team. Similar to the `osconfig-admins` team receiving the `admin` role.

https://docs.github.com/en/organizations/managing-user-access-to-your-organizations-repositories/managing-repository-roles/repository-roles-for-an-organization

robertschaedler3: This has the same effect whether members are listed explicitly or not. 2 members of the team must review every pull request. The reason to use teams instead of listing all the members explicitly in this file is to prevent needing to change the file if the team structure changes. 

All members of the team will receive email notifications when a PR is opened or when the team is mentioned. It is misleading to have 2 individuals listed there when we are ultimately looking for reviews from team members.

robertschaedler3: This will not change the notification behavior or PR requirements. Specifying just the team alias is the most clear and direct way to express our desired rules and will match will what we have internally on other projects.

.github/CODEOWNERS
robertschaedler3: As described below, this will not change the notification behavior or PR requirements.

MariusNi: As discussed, we shall not change the team names and code owners for adding more people to be code reviewers. Recommend to leave all else as-is and add the new members to azure-osconfig-core (existing) and azure-osconfig-extended (new). Leave codeowners set to azure-core-osconfig *plus* at minimum MariusNi.

.github/CODEOWNERS
MariusNi: Regardless of notifications behavior and everything else stated, please put back MariusNi. And not just that, leave the azure-osconfig-team name as it was and add people to it.

MariusNi: Closing this comment, we discuss this in at least another.

MariusNi (approved):
:shipit:

PR #578, opened 2023-12-07
Continued stabilization for the SSH policy ran by the Universal NRP and the Security Baseline module

MariusNi:
## Description

Continued stabilization for the SSH policy ran by the Universal NRP and the Security Baseline module, including bug fixes and improvements. 

## Checklist

### DIFF
--- a/src/common/commonutils/CommonUtils.h
+++ b/src/common/commonutils/CommonUtils.h
@@ -41,6 +41,12 @@
     }\
 }\
 
+#define OsConfigResetReason(reason) {\
+    if (NULL != reason) {\
+        FREE_MEMORY(*reason);\
+    }\
+}\
+
 // Linefeed (LF) ASCII character
 #ifndef EOL
 #define EOL 10
--- a/src/common/commonutils/CommonUtils.h
+++ b/src/common/commonutils/CommonUtils.h
@@ -37,7 +37,7 @@
 #define OsConfigCaptureSuccessReason(reason, FORMAT, ...) {\
     if (NULL != reason) {\
         FREE_MEMORY(*reason);\
-        *reason = FormatAllocateString(FORMAT, temp, SECURITY_AUDIT_PASS, ##__VA_ARGS__); \
+        *reason = FormatAllocateString(FORMAT, SECURITY_AUDIT_PASS, ##__VA_ARGS__); \
     }\
 }\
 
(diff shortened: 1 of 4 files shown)


src/common/commonutils/SshUtils.c
```diff
@@ -528,6 +528,7 @@ int CheckSshProtocol(char** reason, void* log)
     else if (EEXIST == (status = CheckLineNotFoundOrCommentedOut(g_sshServerConfiguration, '#', protocol, log)))
     {
         OsConfigLogInfo(log, "CheckSshProtocol: '%s' is found uncommented in %s", protocol, g_sshServerConfiguration);
+        OsConfigCaptureSuccessReason(reason, "%s'%s' is found uncommented in %s", protocol, g_sshServerConfiguration);
```
simonjaeger, line 531: nit: Should the last %s be covered in single quotes for consistency with others?

simonjaeger (approved):
LGTM

src/common/commonutils/CommonUtils.h
```diff
@@ -34,6 +34,19 @@
...
 }\
 
+#define OsConfigCaptureSuccessReason(reason, FORMAT, ...) {\
+    if (NULL != reason) {\
+        FREE_MEMORY(*reason);\
+        *reason = FormatAllocateString(FORMAT, SECURITY_AUDIT_PASS, ##__VA_ARGS__); \
+    }\
+}\
+
+#define OsConfigResetReason(reason) {\
```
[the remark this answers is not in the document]
MariusNi: Macros are way more performant as they are expanded inline. Esp for small bits of code like this.

src/common/commonutils/SshUtils.c
MariusNi: This is by design and yes, it's already consistent. This last string is the SSH server configuration file and simply it looks better there in reasons w/o the '' delimiters which are reserved for values.

MariusNi: " 'Protocol 2' is found uncommented in /etc/ssh/sshd_config"

Same as for the sshd service here:

" The sshd service reports that 'clientaliveinterval' is set to '3600' (that is greater than zero)"

MariusNi: sshd and /etc/ssh/sshd_config do not get quoted, they are the two main pieces there and are always the same

The values (which can change) get quoted and that only to see if there are spaces sneaking after and such

robertschaedler3 (approved):

PR #591, opened 2024-01-23
Adding initialization for SSH policy parameter values in fallback case

MariusNi:
## Description

When the user attempts to change default parameter values for policy we need to submit pre-audit the user desired values. This change here targets the NRP fallback case. For the OSConfig case the best course of action will be to add new set of initFoo (alongside auditFoo and remediateFoo) MIM desired objects to receive the desired values.

## Checklist


robertschaedler3 (approved):

robertschaedler3 (approved):

AhmedBM (approved):
👌

PR #592, opened 2024-01-24
Making the audit for users and groups allowed and denied for SSH to accept multiple entries from sshd

MariusNi:
## Description

When multiple users or groups are configured to be allowed or denied for SSH, querying the OpenSSH server may return multiple lines, with one user or group per each. We need to support this case in order to correctly audit.

## Checklist


simonjaeger (approved):
LGTM

AhmedBM (approved):
👍

PR #601, opened 2024-01-30
Temporarily removing Ubuntu 22.04 from the E2E tests due to AIS unavailability

robertschaedler3:
- Fix E2E tests (pass connection string secret between jobs)
- Temporarily remove Ubuntu 22.04 E2E test job
- Add Ubuntu 22.04 cloudtest managed image definition

## Checklist

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -19,7 +19,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      - uses: hashicorp/setup-terraform@v2
+      - uses: hashicorp/setup-terraform@v3
         with:
           terraform_version: latest
           terraform_wrapper: false
@@ -38,7 +38,7 @@ jobs:
           terraform apply -auto-approve
 
           echo iothub_name=$(terraform output -raw iothub_name) >> $GITHUB_OUTPUT
-          echo connection_string=$(terraform output -raw connection_string) >> $GITHUB_OUTPUT
+          echo connection_string=$(terraform output -raw connection_string | base64 -w 0) >> $GITHUB_OUTPUT
 
   test:
     name: Test
@@ -47,7 +47,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        target: [ubuntu-18.04, ubuntu-20.04, ubuntu-22.04, debian-10, debian-11]
+        target: [ubuntu-18.04, ubuntu-20.04, debian-10, debian-11]
         arch: [amd64]
     secrets:
       client_id: ${{ secrets.CLIENT_ID }}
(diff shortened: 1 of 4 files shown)


.github/workflows/e2e.yml
```diff
@@ -47,7 +47,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        target: [ubuntu-18.04, ubuntu-20.04, ubuntu-22.04, debian-10, debian-11]
+        target: [ubuntu-18.04, ubuntu-20.04, debian-10, debian-11]
```
MariusNi, line 50: >target: [ubuntu-18.04, ubuntu-20.04, debian-10, debian-11] [](http://example.com/codeflow?start=8&length=58)

This PR is not fixing the E2E tests, it hides the current problem on Ubuntu 22.04 by removing it from E2E, while not addressing the issue for the other distros to be added in near future. OK with this for now, however we need to change the PR name to make the intent clear. See the suggested PR name and decide if you want to keep like that or change. Otherwise there is no problem here and I will sign off.

devops/e2e/cloudtest/ubuntu-22_04.json
```diff
@@ -0,0 +1,31 @@
...
+            "parameters": {
+                "repository": "https://packages.microsoft.com/ubuntu/22.04/prod"
+            }
+        },
+        {
+            "name": "linux-install-packages",
+            "parameters": {
+                "packages": "aziot-identity-service jq"
+            }
+        },
```
MariusNi, line 20: Why are we instrumenting Ubuntu 22.04 as it were to support AIS? Does AIS work here and does not work in the other case? Please explain.

.github/workflows/e2e.yml
robertschaedler3: Agreed with updated PR name. This does not fix the issue for Ubuntu 22.04, however the tests were failing for other reasons that are fixed in this PR

devops/e2e/cloudtest/ubuntu-22_04.json
robertschaedler3: It does get installed on the VM, however the service does not appear to be available/OSConfig is unable to connect to it and obtain a the connection string. I was able to confirm that the connection string is added to the AIS `config.toml` file, however the socket that the PnP agent uses to communicate with AIS is not available.

AhmedBM (approved):
👌

devops/e2e/cloudtest/ubuntu-22_04.json
MariusNi: Then we should skip AIS for good and do not install the package on Ubuntu 22.04 as it won't be used anyway?

MariusNi (approved):
:shipit:

PR #619, opened 2024-02-15
Adding a bash script to reinstall the openssh server and update osconfig

saicharithc:
## Description

Adding a bash script to reinstall openssh & update osconfig. This script will be useful in validation across various distros.

## Checklist

### DIFF
--- a/src/adapters/mc/sshpreview/refresh_for_preview.sh
+++ b/src/adapters/mc/sshpreview/refresh_for_preview.sh
@@ -1,0 +1,33 @@
+#!/bin/bash
+
+if [ -n "$(command -v yum)" ]; then
+        echo "using yum"
+        yum remove openssh-server -y
+        rm /etc/ssh/sshd_config*
+        rm /etc/ssh/sshd_config.d/osconfig_*
+        yum install openssh-server -y
+
+        if [[ $(yum list installed | grep -wi osconfig) ]]; then
+                echo "OSConfig Installed, updating"
+                yum update
+                yum install osconfig
+        else
+                echo "OSConfig Not Present"
+        fi
+fi
+
+if [ -n "$(command -v apt-get)" ]; then
+        echo "using apt-get"
+        apt-get remove --purge openssh-server -y
+        rm /etc/ssh/sshd_config*
+        rm /etc/ssh/sshd_config.d/osconfig_*
+        apt-get install openssh-server -y
+
+        if [[ $(apt list osconfig | grep -wi osconfig) ]]; then
+                echo "OSConfig Installed, updating"
+                apt-get update
+                apt-get install osconfig
+        else
+                echo "OSConfig Not Present"
+        fi
+fi
--- a/src/adapters/mc/sshpreview/refresh_for_preview.sh
+++ b/src/adapters/mc/sshpreview/refresh_for_preview.sh
@@ -2,32 +2,32 @@
 
 if [ -n "$(command -v yum)" ]; then
         echo "using yum"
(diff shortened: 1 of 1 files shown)


src/adapters/mc/sshpreview/refresh_for_preview.sh
```diff
@@ -0,0 +1,33 @@
...
+        rm /etc/ssh/sshd_config*
+        rm /etc/ssh/sshd_config.d/osconfig_*
+        yum install openssh-server -y
+
+        if [[ $(yum list installed | grep -wi osconfig) ]]; then
+                echo "OSConfig Installed, updating"
+                yum update
+                yum install osconfig
+        else
+                echo "OSConfig Not Present"
```
MariusNi, line 15: nit: "installed", "not present" as all lowercase.

Also would be better to match: 

present - missing
installed - uninstalled

src/adapters/mc/sshpreview/refresh_for_preview.sh
```diff
@@ -0,0 +1,33 @@
...
+        fi
+fi
+
+if [ -n "$(command -v apt-get)" ]; then
+        echo "using apt-get"
+        apt-get remove --purge openssh-server -y
+        rm /etc/ssh/sshd_config*
+        rm /etc/ssh/sshd_config.d/osconfig_*
+        apt-get install openssh-server -y
```
MariusNi, line 24: Can we add sudo in here to allow the user to not have to run the script elevated?

src/adapters/mc/sshpreview/refresh_for_preview.sh
saicharithc: updated to installed/uninstalled and to lowercase

src/adapters/mc/sshpreview/refresh_for_preview.sh
saicharithc: added sudo

MariusNi (approved):
:shipit:

src/adapters/mc/sshpreview/refresh_for_preview.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+
+if [ -n "$(command -v yum)" ]; then
+        echo "using yum"
+        sudo yum remove openssh-server -y
```
AhmedBM, line 5: i wouldn't put any `sudo` command in the script. Have them invoke with sudo instead. Makes the script much more portable for rootless environments

MariusNi: Can you explain why not?

AhmedBM: As a best practice, scripts should be written to require the least privileges necessary for the task they need to perform. If a script needs to perform actions as root, it’s often better to have the user invoke the script with sudo rather than embedding sudo in the script itself. This makes the need for root privileges explicit and keeps the control in the hands of the user.

AhmedBM: all of the build containers are rootness, mind you, we use these only for building and not running osconfig. But it is an environment to consider

simonjaeger (approved):

src/adapters/mc/sshpreview/refresh_for_preview.sh
saicharithc: removed sudo from the script

AhmedBM (approved):
👌

MariusNi (approved):
:shipit:

PR #628, opened 2024-02-28
Combining SSH and Security Baseline policy MOF manifests (draft) and temporarily reducing the test recipe for the later module

MariusNi:


src/modules/test/recipes/SecurityBaselineTests.json
```diff
@@ -110,668 +110,10 @@
     "ComponentName": "SecurityBaseline",
```
simonjaeger, line 110: Why is this being reduced? Do they not work? If so, are we tracking it?

simonjaeger (approved):

src/modules/test/recipes/Backup/SecurityBaselineTests_full.json
```diff
@@ -0,0 +1,1957 @@
+[
```
robertschaedler3, line 1: What is the difference between this backup/full test and the other test recipe? Can they be a single test?

src/modules/test/recipes/Backup/SecurityBaselineTests_full.json
robertschaedler3, line 1: If this file is under "Backup" I do not think it is run by the module test. Is this true?

robertschaedler3 (approved):

src/modules/test/recipes/SecurityBaselineTests.json
MariusNi: This test recipe in its original form changes things (access rights to system and user files among them) and there is no VM that in default form can pass all audit checks. IN order to include this test recipe to the automated functional test for OSConfig, we need to clean it of things that are likely to fail and we'll add those bask as we advance through the implementation and we make things stable.

src/modules/test/recipes/Backup/SecurityBaselineTests_full.json
MariusNi: Correct. If left under the original folder it will be picked up by the test even with a changed name. To save a copy of the full recipe to bring it back later one fragment at a time we save a copy under this separate folder.

src/modules/test/recipes/Backup/SecurityBaselineTests_full.json
MariusNi: At some point in the future when the entire Security Baseline will finished as the SSH part is, there will be one test. For now, only the SSH part can remain in the official test (also have to remove remaining audits here)

robertschaedler3 (approved):

AhmedBM:
I see it has not been tested on the [Module Test](https://github.com/Azure/azure-osconfig/actions/workflows/module-test.yml) workflow. Before merging we should make sure it works

MariusNi:
> I see it has not been tested on the [Module Test](https://github.com/Azure/azure-osconfig/actions/workflows/module-test.yml) workflow. Before merging we should make sure it works

We do not have dcumented such step for PRs. Also, we should not run functional tests during CI. Manually triggered a new test pass as you suggested and it failed. It seems it does not run with what's in this branch but possibly main: "[Test (mariner, 2, RPM, Pmc|SecurityBaseline|Sample, amd64) / test](https://github.com/Azure/azure-osconfig/actions/runs/8085034273/job/22091853946#step:5:771)"

Will complete this PR and continue to investigate any problems that may remain in main for the test recipe.

PR #629, opened 2024-02-28
Downgrade upload-artifact from v4 to v3

robertschaedler3:
Downgrade upload-artifact to v3 to allow multiple uploads to a single `unit-test` artifact (used for generating the test report).

> https://github.com/actions/upload-artifact/issues/478


MariusNi (approved):
:shipit:

simonjaeger (approved):

AhmedBM:
We have a few more instances of `actions/upload-artifact` should we also change these which also might have multiple runs, similar to our CI runs

PR #650, opened 2024-03-13
Add managed identity for self-hosted runner authentication

robertschaedler3:
- Use a [managed identity](https://learn.microsoft.com/en-us/entra/identity/managed-identities-azure-resources/how-manage-user-assigned-managed-identities?pivots=identity-mi-methods-azp) for authenticating with internal Azure resources from 1ES hosted runners to improve security (for use in `release-pool`/`e2e-pool`)
- Create separate `ci-pool` with no managed identity role assignments


.github/workflows/package-build.yml
```diff
@@ -112,7 +112,6 @@ jobs:
             ./build/*.zip
             ./build/modules/test/moduletest
             ./build/modules/bin/*.so
-            ./src/tests/universal-nrp-e2e/*.Tests.ps1
```
AhmedBM, line 115: don't remove this, this is needed for the Universal NRP Tests

AhmedBM (changes requested):
Must ensure the Universal NRP Pester Tests is present in the package build

.github/workflows/package-build.yml
robertschaedler3: It looks like it is not used there. The `universalnrp-test-run` workflow does a checkout before comying/running the tests: https://github.com/Azure/azure-osconfig/blob/457201988e19dfcf37b70793a26868be2faeb366/.github/workflows/universalnrp-test-run.yml#L35-L45

robertschaedler3: If we need it in an artifact we should probably put it in a different one since it changes the artifact output contents (adds the build directory structure + breaks E2E tests).

AhmedBM: ahh yes, didn't see that. Actually, this would fail anyhow if it wouldn't find it.

AhmedBM (approved):
👌

MariusNi (approved):
:shipit:

PR #654, opened 2024-03-18
MC E2E Distro Expansion

AhmedBM:
## Description

 * Added Machine Configuration e2e workflow targeting the following distros (daily + workflow dispatch):
   * Centos 8
   * Debian 10, 11
   * Mariner 2 (aka Azure Linux)
   * Oracle Linux 8
   * Red Hat Enterprise Linux 8, 9
   * Suse Linux Enterprise (aka sles) 15
   * Ubuntu 20.04, 22.04
 * Removed MC e2e from CI
 * Updated Cloudtest manifests

## Checklist

### DIFF
--- a/.github/workflows/universalnrp-test.yml
+++ b/.github/workflows/universalnrp-test.yml
@@ -1,7 +1,6 @@
 name: Universal NRP Test
 
 on:
-  push:
   workflow_dispatch:
   schedule:
     - cron: '0 20 * * *' # Every day at 12pm PST (UTC-8)
--- a/.github/workflows/universalnrp-test.yml
+++ b/.github/workflows/universalnrp-test.yml
@@ -1,6 +1,9 @@
 name: Universal NRP Test
 
-on: [workflow_dispatch]
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 8 * * *' # Every day at 8am UTC (12am PST)
 
 jobs:
   test:
@@ -11,7 +14,18 @@ jobs:
       matrix:
         target:
           [
+            { os: centos, version: 8, package-type: RPM, tag: ''},
+            { os: debian, version: 10, package-type: DEB, tag: ''},
+            { os: debian, version: 11, package-type: DEB, tag: ''},
+            # { os: debian, version: 12, package-type: DEB, tag: ''},
+            { os: mariner, version: 2, package-type: RPM, tag: ''},
+            { os: oraclelinux, version: 8, package-type: RPM, tag: ''},
+            { os: rhel, version: 8, package-type: RPM, tag: ''},
+            { os: rhel, version: 9, package-type: RPM, tag: ''},
+            # { os: rockylinux, version: 9, package-type: RPM, tag: ''},
+            { os: sles, version: 15, package-type: RPM, tag: ''},
             { os: ubuntu, version: 20.04, package-type: DEB, tag: ''},
+            { os: ubuntu, version: 22.04, package-type: DEB, tag: ''},
           ]
         arch: [amd64]
     with:
--- a/.github/workflows/universalnrp-test.yml
+++ b/.github/workflows/universalnrp-test.yml
@@ -1,9 +1,10 @@
 name: Universal NRP Test
 
 on:
+  push:
   workflow_dispatch:
   schedule:
-    - cron: '0 8 * * *' # Every day at 8am UTC (12am PST)
+    - cron: '0 20 * * *' # Every day at 12pm PST (UTC-8)
 
 jobs:
   test:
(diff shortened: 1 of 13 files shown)


.github/workflows/universalnrp-test.yml
```diff
@@ -11,7 +14,18 @@ jobs:
       matrix:
         target:
           [
+            { os: centos, version: 8, package-type: RPM, tag: ''},
+            { os: debian, version: 10, package-type: DEB, tag: ''},
+            { os: debian, version: 11, package-type: DEB, tag: ''},
+            # { os: debian, version: 12, package-type: DEB, tag: ''},
```
MariusNi, line 20: ># { os: debian, version: 12, package-type: DEB, tag: ''}, [](http://example.com/codeflow?start=12&length=57)

Why are Debian 12 and Rocky Linux 9 disabled here?

.github/workflows/universalnrp-test.yml
```diff
@@ -1,6 +1,9 @@
 name: Universal NRP Test
 
-on: [workflow_dispatch]
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 8 * * *' # Every day at 8am UTC (12am PST)
```
MariusNi, line 6: >Every day at 8am UTC (12am PST) [](http://example.com/codeflow?start=26&length=31)

Suggest 12 AM PST which is 3 AM EST. This time that is now is too late for PST.

.github/workflows/universalnrp-test.yml
```diff
@@ -11,7 +14,18 @@ jobs:
       matrix:
         target:
```
MariusNi, line 15: > target: [](http://example.com/codeflow?start=7&length=8)

Where we differentiate the targets among the set with OSConfig vs the set without OSConfig?

MariusNi: We need both sets run on separate VMs every run.

.github/workflows/universalnrp-test.yml
AhmedBM: Although the images are created, they have issues booting. Im Fixing this and left them commented out for now

.github/workflows/universalnrp-test.yml
AhmedBM: All of these sets are without osconfig, should i also include all these targets with it installed as well?

MariusNi: Yes, please. We need the set duplicated with equal number of distros: with and without the OSConfig package installed.

.github/workflows/universalnrp-test.yml
MariusNi: We also need Amazon Linux added.

AhmedBM: Removed distros with issues in automation. Will have new PR with them reintroduced when working.

MariusNi (approved):
:shipit:

.github/workflows/universalnrp-test.yml
```diff
@@ -1,6 +1,10 @@
 name: Universal NRP Test
 
-on: [workflow_dispatch]
+on:
+  push:
```
robertschaedler3, line 4: We should probably restrict this a bit. The default `push` event will run on a push to any branch. If the goal is to get this to run on feature branches, `workflow_dispatch` is best. But if the goal is to also run on PRs, this should stay in the CI workflow. 

Can we restrict this a bit?

.github/workflows/universalnrp-test.yml
```diff
@@ -11,11 +15,23 @@ jobs:
...
+            { os: debian, version: 10, package-type: DEB, tag: ''},
+            { os: debian, version: 11, package-type: DEB, tag: ''},
+            { os: mariner, version: 2, package-type: RPM, tag: ''},
+            { os: oraclelinux, version: 8, package-type: RPM, tag: ''},
+            { os: rhel, version: 8, package-type: RPM, tag: ''},
+            { os: rhel, version: 9, package-type: RPM, tag: ''},
+            { os: rockylinux, version: 9, package-type: RPM, tag: ''},
+            { os: sles, version: 15, package-type: RPM, tag: ''},
             { os: ubuntu, version: 20.04, package-type: DEB, tag: ''},
+            { os: ubuntu, version: 22.04, package-type: DEB, tag: ''},
```
robertschaedler3, line 28: Should we also add these jobs as required (same as CI unit test jobs)?

.github/workflows/universalnrp-test.yml
AhmedBM: Oups, that one snuck in. Good catch! Intent is to just have workflows dispatch and pr only

.github/workflows/universalnrp-test.yml
AhmedBM: Eventually yes, some are currently failing  but once they are clean I will introduce them

robertschaedler3: Awesome sounds good!

robertschaedler3 (approved):

PR #659, opened 2024-03-25
Added git ref support to workflows, fixed prod release

AhmedBM:
## Description

* Updated prod workflow to publish our previous stable `v1.0.5` distros
* Added missing arm64 targets to both `prod`, `insiders-fast` channels
* Removed release temporarily, will re-introduce in separate PR

## Checklist


.github/workflows/prod.yml
```diff
@@ -13,60 +16,24 @@ jobs:
...
-            { os: debian, version: 10, dist: buster, package-type: DEB },
-            { os: debian, version: 11, dist: bullseye, package-type: DEB },
+            { os: debian, version: 10, arch: amd64, dist: buster, package-type: DEB, repo: microsoft-debian-buster-prod-apt },
+            { os: debian, version: 10, arch: arm64, dist: buster, package-type: DEB, repo: microsoft-debian-buster-prod-apt },
+            { os: debian, version: 11, arch: amd64, dist: bullseye, package-type: DEB, repo: microsoft-debian-bullseye-prod-apt },
+            { os: debian, version: 11, arch: arm64, dist: bullseye, package-type: DEB, repo: microsoft-debian-bullseye-prod-apt },
+            { os: debian, version: 12, arch: amd64, dist: bookworm, package-type: DEB, repo: microsoft-debian-bookworm-prod-apt },
+            { os: debian, version: 12, arch: arm64, dist: bookworm, package-type: DEB, repo: microsoft-debian-bookworm-prod-apt },
+            { os: ubuntu, version: 18.04, arch: amd64, dist: bionic, package-type: DEB, repo: microsoft-ubuntu-bionic-prod-apt },
+            { os: ubuntu, version: 18.04, arch: arm64, dist: bionic, package-type: DEB, repo: microsoft-ubuntu-bionic-prod-apt },
```
MariusNi, line 26: Ubuntu 18.04 is out of all our targets as discussed offline

MariusNi: Instead, we need Ubuntu 22.04 here, right?

MariusNi (approved):
:shipit:

.github/workflows/prod.yml
```diff
@@ -1,9 +1,12 @@
 name: Publish (prod)
 
 on:
-  push:
-    tags:
-      - v*.*.*
+  workflow_dispatch:
+    inputs:
+      ref:
```
robertschaedler3, line 6: Isn't `ref` already available for `workflow_dispatch` events? Why do we need to specify this manually/duplicate the `ref`?

> When you trigger the event, you can provide the `ref` and any `inputs`

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch

> 5. Select the Branch dropdown menu and click a branch to run the workflow on

https://docs.github.com/en/actions/using-workflows/manually-running-a-workflow

.github/workflows/prod.yml
```diff
@@ -13,60 +16,24 @@ jobs:
...
+      arch: ${{ matrix.target.arch }}
+      repo: ${{ matrix.target.repo }}
       dist: ${{ matrix.target.dist }}
+      # ESRP debian based repos (deb) have channels/rings on repos.
+      # ESRP yum based repos (rpm) have no channels
+      channel: ${{ matrix.target.package-type == 'DEB' && 'prod' || '' }}
       package-type: ${{ matrix.target.package-type }}
     secrets: inherit
-
-  release:
```
robertschaedler3, line 33: Are we no longer publishing releases to GitHub?

.github/workflows/prod.yml
```diff
@@ -1,9 +1,12 @@
 name: Publish (prod)
 
 on:
-  push:
-    tags:
-      - v*.*.*
```
robertschaedler3, line 6: Can we add `workflow_dispatch` to the existing `push` trigger so that we can still easily publish with a version tag?

.github/workflows/package-build.yml
```diff
@@ -25,6 +25,10 @@ on:
         description: The type of package to build (deb | rpm).
         type: string
         required: true
+      ref:
```
robertschaedler3, line 28: Won't this workflow run in the same context as its parent (e.g. have the same implicit `ref`)?

.github/workflows/universalnrp-test-run.yml
```diff
@@ -21,6 +21,10 @@ on:
         description: 'Pester tag to include in the test run. If not provided, all tests will be run.'
         default: ''
         type: string
+      ref:
```
robertschaedler3, line 24: Same as above, this should have the same implicit `ref` (context) as the workflow that calls this. Why do we need to explicitly specify a `ref`?

robertschaedler3 (commented):
The GitHub context includes the ref/sha where the event trigger originated from (`workflow_dispatch` - from the branch where the workflow was run on | `workflow_call` - shared the context/ref of the caller). 

https://docs.github.com/en/actions/learn-github-actions/contexts#github-context

.github/workflows/prod.yml
robertschaedler3: If we want to be able to update releases on GitHub that should also be supported by the gh cli, we just need to detect if the release already exists. I remember seeing an example of this somewhere/can find a link if this is what we intend to do?

AhmedBM: We will, but we will manually create/edit them, its easier that way

.github/workflows/prod.yml
AhmedBM: nice catch, didn't know this was an available built-in, i will incorporate it!

.github/workflows/universalnrp-test-run.yml
AhmedBM: i thought you had to be explicit, just read on [Contexts](https://docs.github.com/en/actions/learn-github-actions/contexts) and your right, good catch thanks!

.github/workflows/prod.yml
MariusNi: Unclear here, so not dropping package release to GitHub, just making them manual instead of automatic? Can you please clarify?

robertschaedler3: Yes why not have them automatic/updated (or at least created automatically)? Then edit if needed?

.github/workflows/prod.yml
robertschaedler3: Awesome! Thanks!

.github/workflows/prod.yml
AhmedBM: Ok, i see we feel strongly about creating a new release through the workflows. I'll re-introduce in another PR as we are currently fixing an existing release and i don't want to complicate the workflow logic for this one-off update.

.github/workflows/prod.yml
AhmedBM: Sure, ill re-add it

.github/workflows/prod.yml
MariusNi: So then leave this portion here in this PR and change in a next one?

AhmedBM: yes

robertschaedler3: If going to re-add most of this later it is possible to just add an `if: false` to the job to skip it. I think we did something similar for the `container-build` workflow a while ago.

AhmedBM: Alright, ill re-introduce with `if: false`. All of the artifact names have changed since the workflow refactor though which is why i decided to just remove it as this will all change.

robertschaedler3 (approved):

AhmedBM:
Spoke offline about Alma Linux comment in prod workflows and ABI compatibility between Alma/Rocky/RHEL. Packages.microsoft.com (PMC) has an alias on the Rocky/Alma-9 repos so they share the same rpms set.

.github/workflows/prod.yml
robertschaedler3: Ahh okay I see! 👍

MariusNi (approved):
:shipit:

PR #680, opened 2024-04-26
Refactoring bulk of the ASB implementation code from the Security Baseline module into a new commonutils library

MariusNi:
## Description

This is bulk refactoring of the ASB implementation code from the Security Baseline module into a new separate library that gets linked and used by both the Universal NRP and by the Security Baseline module, for both the SSH Control Posture and Azure Security Baseline for Linux audit-only. This code builds and has no regressions for either, while enabling the ASB audit on all distros.

Some minor changes were necessary to this ASB implementation when moved in order to successfully compile with gcc 4.8 on Ubuntu 14 for the NRP build. This old gcc version has some quirky implementation of some of the warnings.

Includes temporary and unfinished code for policy parameter configuration for the ASB audit checks which is here done in bulk, semi-generated, for all, just so that the code compiles. Those lists will be trimmed down and then code formatted and completed. For now, this code all works for its intended purpose and is successfully deployed via test policies.

The unit-tests for the ASB implementation remain for the Security Baseline module so they continue to cover the MMI implementation for the MIM of the module.

The Security Baseline module implementation is greatly simplified.

The Universal NRP is refactored to drop the fallback and instead, via the new ASB library, always resort to direct ASB calls, unless a policy check is not supported and only then looking for and calling into OSConfig's platform MPI if available. 

The Universal NRP is also no longer installing the OSConfig package itself.

## Checklist


src/modules/securitybaseline/src/lib/CMakeLists.txt
```diff
@@ -3,11 +3,17 @@
...
 set_property(TARGET securitybaselinelib PROPERTY POSITION_INDEPENDENT_CODE ON)
 
-target_link_libraries(securitybaselinelib PRIVATE logging commonutils parsonlib)
-target_include_directories(securitybaselinelib
-    PUBLIC
-        ${MODULES_INC_DIR}
-        ${CMAKE_CURRENT_SOURCE_DIR})
\ No newline at end of file
+target_link_libraries(securitybaselinelib PRIVATE 
+    logging
```
AhmedBM, line 12: can make `logging` + `commonutils` `PUBLIC`? You can remove them from your so\CMakeLists.txt `target_link_libraries`

src/common/asb/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
...
+    Asb.c)
+
+target_compile_options(asb PRIVATE -Wno-psabi)
+
+target_include_directories(asb PUBLIC
+    ${CMAKE_CURRENT_SOURCE_DIR}
+    ${PLATFORM_INC_DIR})
+
+target_link_libraries(asb PRIVATE 
+    logging
```
AhmedBM, line 16: Same here, can we make `logging`, `commonutils` PUBLIC so that the downstream target doesn't have to include the libraries?

AhmedBM (approved):
Just some minor CMake comments, but otherwise, looks good

src/modules/securitybaseline/src/lib/CMakeLists.txt
AhmedBM: understood, but `PUBLIC` in this context simply means that those dependencies are transitive, so upstream targets (so's) inherit those targets. If your redefining those targets for the so then they should be made PUBLIC in the downstream target

MariusNi: That is not the case here. Why would you suggest PUBLIC at all?

robertschaedler3 (approved):
👍

PR #683, opened 2024-04-29
Removed push triggers from e2e tests

AhmedBM:
## Description

* Removed push triggers from E2E tests

## Checklist

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,11 +1,6 @@
 name: E2E
 
 on:
-  push:
-    branches:
-      - main
-    paths:
-      - src/**
   workflow_dispatch:
 
 jobs:


.github/workflows/e2e.yml
```diff
@@ -1,11 +1,6 @@
 name: E2E
 
 on:
-  push:
```
robertschaedler3, line 4: Since the `push` trigger is being removed, should we also add a `cron` trigger to periodically run the E2E tests on main?

robertschaedler3 (approved):

.github/workflows/e2e.yml
MariusNi: Not now, the E2E tests are currently broken and need to be refactored to do not depend on AIS or IoT Hub at all and instead test at MPI level.

MariusNi (approved):
:shipit:

.github/workflows/e2e.yml
AhmedBM: Eventually yes, the e2e are not working because of the broken AIS dependency

PR #687, opened 2024-05-09
Fixing audit and adding remediation for file system mounting options for Security Baseline

MariusNi:
## Description

Making the audits to pass when no such mount entries are found (and fail only when the desired security options do not exist on existing entries) and adding automatic remediation that adds desired options to existing entries that don't have them. Remediation looks first in /etc/fstab and then next in /etc/mtab and picks from there entries that match to remediate with.

## Checklist


src/common/asb/Asb.c
```diff
@@ -604,6 +608,14 @@ void AsbInitialize(void* log)
     {
         OsConfigLogError(log, "AsbInitialize: failed to allocate memory");
     }
+
+    if (false == FileExists(g_etcFstabCopy))
```
simonjaeger, line 612: What is the reason for not doing `!FileExists(g_etcFstabCopy) && !MakeFileBackupCopy(g_etcFstab, g_etcFstabCopy, log)`? Is this style consistent all throughout?

src/common/commonutils/FileUtils.c
```diff
@@ -67,31 +67,79 @@ bool SavePayloadToFile(const char* fileName, const char* payload, const int payl
...
                 UnlockFile(file, log);
             }
             else
             {
-                OsConfigLogError(log, "SavePayloadToFile: cannot lock '%s' for exclusive access while writing (%d)", fileName, errno);
+                OsConfigLogError(log, "SaveToFile: cannot lock '%s' for exclusive access while writing (%d)", fileName, errno);
             }
-            
+
             fclose(file);
```
simonjaeger, line 81: Is it worth to check error code?

src/common/commonutils/FileUtils.c
```diff
@@ -514,6 +563,288 @@ int CheckFileSystemMountingOption(const char* mountFileName, const char* mountDi
...
+            if (0 == (status = errno))
+            {
+                status = ENOENT;
+            }
+            
+            OsConfigLogError(log, "CopyMountFile: could not open source file '%s', setmntent() failed (%d)", source, status);
+        }
+        
+        fflush(targetHandle);
+        endmntent(targetHandle);
```
simonjaeger, line 615: Same here, is it worth to check error codes?

src/common/asb/Asb.c
MariusNi: To be fully explicit here, as rest of this function. For Boolean values this is the best way, followed by no comparison for true and then ! for false.

src/common/commonutils/FileUtils.c
```diff
@@ -514,6 +563,288 @@ int CheckFileSystemMountingOption(const char* mountFileName, const char* mountDi
...
+
+    if (((NULL == mountDirectory) && (NULL == mountType)) || (NULL == desiredOption))
+    {
+        OsConfigLogError(log, "SetFileSystemMountingOption called with invalid argument(s)");
+        return EINVAL;
+    }
+
+    if (!FileExists(fsMountTable))
+    {
+        OsConfigLogInfo(log, "SetFileSystemMountingOption: '%s' not found, no place to set mounting options", fsMountTable);
```
simonjaeger, line 657: I assume that not finding the mount means it's a no-op and that is why it is OK to return zero?

src/common/commonutils/FileUtils.c
MariusNi: Arguably, not. Unlikely to fail and if it fails, there is nothing else to do. Worst case scenario who gets this file next will find it empty/not flushed correctly.

src/common/commonutils/FileUtils.c
MariusNi: This is similar to fclose. Same reasons. Typically, these are not checked for errors as there is nothing else to follow up after.

src/common/commonutils/FileUtils.c
MariusNi: See this trace portion: " no place to set mounting options". Cannot add mount options to inexistent mount points.

simonjaeger (approved):

src/common/commonutils/FileUtils.c
```diff
@@ -514,6 +563,288 @@ int CheckFileSystemMountingOption(const char* mountFileName, const char* mountDi
...
+    const char* newLineAddNewTemplate = "\n%s %s %s %s,%s %d %d";
+
+    FILE* fsMountHandle = NULL;
+    FILE* mountHandle = NULL;
+    char* newLine = NULL;
+    char* tempFileNameOne = NULL;
+    char* tempFileNameTwo = NULL;
+    struct mntent* mountStruct = NULL;
+    bool matchFound = false;
+    int lineNumber = 0;
```
robertschaedler3, line 646: nit: Should this start at 1?

robertschaedler3 (approved):

src/common/commonutils/FileUtils.c
MariusNi: Could. Prefer it as a 0 based index. Line 0 is first, etc.

PR #700, opened 2024-05-28
Adding a new ASBv2 policy parameter for EnsureDefaultDenyFirewallPolicyIsSet and fixing AuditEnsureIpv6ProtocolIsEnabled

MariusNi:
## Description

Adding a new initialization object and new policy parameter for ensuring default firewall policy is set to drop all to allow forcing remediation with first taking a snapshot and allowing all current channels and then blocking all new channels from there on. AuditEnsureDefaultDenyFirewallPolicyIsSet remains the same, RemediateEnsureDefaultDenyFirewallPolicyIsSet either does the same no-op or acts depending on this policy parameter which is currently set to disabled (0). Also fixing AuditEnsureIpv6ProtocolIsEnabled to not look for a system file that may not be correctly updated.

## Checklist


simonjaeger (approved):

AhmedBM (approved):
👌

PR #717, opened 2024-06-10
Continued stabilization for ASB v2 audit and remediation and the CI test automation

MariusNi:
## Description

In this PR, the following improvements are done:

- For the SSH part of the ASB (and also going towards SSH Posture Control RTM), fixing one instance where sshd_config is written directly, replacing that with a secure write where changes are written to a temporary file, then in one atomic step renamed to the system file, preventing file corruption in case of unexpected power down, for example.
- Tuning down the error chatter that happens at the first initialization of the NRP on a machine, when the backup copy of fstab is made.
- Continuing to stabilize the automation CI for the NRP MC PS tests.

## Checklist

### DIFF
--- a/.github/workflows/universalnrp-test.yml
+++ b/.github/workflows/universalnrp-test.yml
@@ -45,3 +45,24 @@ jobs:
       skip-remediation: ${{ matrix.skip-remediation }}
       install-osconfig: ${{ matrix.install-osconfig }}
       tag: ${{ matrix.target.tag }}
+
+report:
+    name: Report
+    needs: test
+    runs-on: ubuntu-latest
+    permissions:
+      checks: write
+      pull-requests: write
+    if: always()
+    steps:
+      - name: Download Test Report Artifacts
+        uses: actions/download-artifact@v4
+        with:
+          path: universal-nrp-test
+          pattern: '*-report'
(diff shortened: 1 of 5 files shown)


robertschaedler3 (approved):

.github/workflows/universalnrp-test.yml
```diff
@@ -44,25 +44,4 @@ jobs:
       resource-count: ${{ matrix.policy-package.resource-count }}
       skip-remediation: ${{ matrix.skip-remediation }}
       install-osconfig: ${{ matrix.install-osconfig }}
-      tag: ${{ matrix.target.tag }}
-
-  report:
```
AhmedBM, line 49: Why is this being removed? This is what consumes the test report

AhmedBM: It is necessary, notice the hash for the commit is not updated with the latest hash
image

robertschaedler3: We should probably keep some form of a test report, odd this one is not updating to the latest commit though. I see the edits made by the bot but latest is not reflected.

robertschaedler3 (approved):

AhmedBM (approved):
👌

PR #718, opened 2024-06-13
Making the MIM test recipe for Security Baseline module test in full ASB v2, including confirming audits passing

MariusNi:
## Description

Making the module tester to be able to run the test recipe for Security Baseline and validate results in full for audit, except the 4 special case checks missing automatic remediation, alongside existing remediation verification.

Also fixing couple of memory leaks in the test and improving test logging for ASB.

Example of last line for an audit reported MIM object check pass:

[2024-06-13 14:26:19] [main.c:467] Assertion passed with reason: 'None of the requested user accounts ('games,test') is present'

## Checklist

### DIFF
--- a/src/modules/test/main.c
+++ b/src/modules/test/main.c
@@ -25,6 +25,7 @@
 #define RECIPE_JSON "Json"
 #define RECIPE_STATUS "ExpectedResult"
 #define RECIPE_WAIT_SECONDS "WaitSeconds"
+#define SECURITY_BASELINE "SecurityBaseline"
 
 #define RECIPE_RUN_COMMAND "RunCommand"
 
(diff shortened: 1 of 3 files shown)


src/modules/test/main.c
```diff
@@ -418,21 +431,53 @@ int RunTestStep(const TEST_STEP* test, const MANAGEMENT_MODULE* module)
...
+                    LOG_INFO("Assertion passed with reason: '%s'", reason + strlen(SECURITY_AUDIT_PASS));
+                }
+            }
+            else if (actualJsonValue != NULL)
             {
                 if (NULL == (expectedJsonValue = json_parse_string(test->payload)))
                 {
                     LOG_ERROR("Failed to parse expected JSON payload: %s", test->payload);
-                    result = EINVAL;
+                    result = -1;
```
AhmedBM, line 475: `EINVAL` seems like a better choice here? Why change?

MariusNi: For consistency. This test uses -1 for test pass failures and in the end does not matter. If rest of this test code would use real errno errors, sure. As it is this looked like a sore thumb.

MariusNi: Replaced all -1 returns in this function with EFAULT

robertschaedler3 (approved):

AhmedBM (approved):
👌

PR #719, opened 2024-06-13
Publishing test policy packages for ASB v2 and SSH Posture Control

MariusNi:
## Description

Validated with these packages stored in private branch that they work fine being referenced from here by Azure Policy. Now adding most current packages to be merged to main and then update the test policies to reference them from there. Next PR will also add copy of test policy definitions with these references.

## Checklist


robertschaedler3 (commented):
We should avoid checking in binary files. `git` cannot get the diff of a binary file, so the entire file content will always be stored in our history (which can hurt git performance over time). For testing, I recommend using the raw file URI from your branch:

https://github.com/Azure/azure-osconfig/raw/MariusNi/OsConfig_ZIP_Jun13/src/adapters/mc/asb/LinuxSecurityBaselinePolicy.zip

https://github.com/Azure/azure-osconfig/raw/MariusNi/OsConfig_ZIP_Jun13/src/adapters/mc/ssh/LinuxSshServerSecurityBaselinePolicy.zip

AhmedBM (changes requested):
We should be checking in these as Artifacts, not in the repo itself. See https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts#uploading-build-and-test-artifacts

Essentially to publish these, we need a new workflow and use https://github.com/actions/upload-artifact to upload the latest custom policy packages into the artifacts.

MariusNi:
We found a better place to publish from (releases)

PR #761, opened 2024-09-06
Re-adding the SSH port policy parameter for the SSH and ASB v2 test policies

MariusNi:
## Description

Re-adding the SSH port policy parameter for the SSH and ASB v2 test policies. This parameter was previously removed from the test policies in order to pass the maximum 20 parameters per policy limit from AzPolicy. The port is very important to be configurable for SSH though. So it's re-added here in two variants for each test definition: one with 19 SSH parameters as reference that cannot be published (gets to 21 with the 2 params for MC/AzP) and another with 18 SSH parameters that has port but excludes another SSH parameter of less significance. 

This change is being tested with newly published test policies for both SSH and ASB v2.

## Checklist


AhmedBM (approved):
👌

robertschaedler3 (approved):

PR #762, opened 2024-09-09
Prevent CheckFileContents cause an invalid reason to fail json_value_init_string and cascade into a false passing audit

MariusNi:
## Description

Prevent CheckFileContents cause an invalid reason to fail json_value_init_string and cascade into a false passing audit. Random contents from checked system files can cause breaks into the JSON envelope for transport if included in the reason, skip those. Also, if still an invalid reason arrives, switch it to a generic fail reason.

Also, write directly to '/proc/sys/kernel/randomize_va_space' rather than staged via a temp file to avoid case where on SUSE the staged write may fail due to parent directory access.

## Checklist


AhmedBM (approved):
👌

robertschaedler3 (approved):

PR #768, opened 2024-10-11
Fix various issues reported by address sanitizer reported from the unit tests

robertwoj-microsoft:
## Description

The following changes fix miscellaneous issues detected by the address sanitizer in the unit tests.

## Checklist

### DIFF
--- a/src/common/commonutils/PassUtils.c
+++ b/src/common/commonutils/PassUtils.c
@@ -81,9 +81,14 @@ int CheckEnsurePasswordReuseIsLimited(int remember, char** reason, void* log)
             g_etcPamdCommonPassword, g_etcPamdSystemAuth, g_remember);
     }
 
-    if (status && (false == FindPamModule(g_pamUnixSo, log)))
+    if (status)
     {
-        OsConfigCaptureReason(reason, "The PAM module '%s' is not available. Automatic remediation is not possible", g_pamUnixSo);
+        char* pamModule = FindPamModule(g_pamUnixSo, log);
+        if(NULL == pamModule)
+        {
+            OsConfigCaptureReason(reason, "The PAM module '%s' is not available. Automatic remediation is not possible", g_pamUnixSo);
+        }
+        free(pamModule);
     }
 
     return status;
@@ -219,7 +224,7 @@ int CheckLockoutForFailedPasswordAttempts(const char* fileName, const char* pamS
                 continue;
             }
             else if ((NULL != strstr(line, auth)) && (NULL != strstr(line, pamSo)) && 
-                (NULL != (authValue = GetStringOptionFromBuffer(line, auth, ' ', log))) && (0 == strcmp(authValue, required)) && FreeAndReturnTrue(authValue) &&
+                (NULL != (authValue = GetStringOptionFromBuffer(line, auth, ' ', log))) && (0 == strcmp(authValue, required)) &&
                 (0 <= (deny = GetIntegerOptionFromBuffer(line, "deny", '=', log))) && (deny <= 5) &&
                 (0 < (unlockTime = GetIntegerOptionFromBuffer(line, "unlock_time", '=', log))))
             {
@@ -229,10 +234,12 @@ int CheckLockoutForFailedPasswordAttempts(const char* fileName, const char* pamS
                     auth, required, pamSo, deny, unlockTime, fileName);
                 
                 status = 0;
+                FREE_MEMORY(authValue);
                 break;
             }
             else
             {
+                FREE_MEMORY(authValue);
                 status = ENOENT;
             }
 
--- a/src/common/commonutils/PassUtils.c
+++ b/src/common/commonutils/PassUtils.c
@@ -62,6 +62,7 @@ static char* FindPamModule(const char* pamModule, void* log)
 int CheckEnsurePasswordReuseIsLimited(int remember, char** reason, void* log)
 {
     int status = ENOENT;
+    char* pamModule = NULL;
 
     if (0 == CheckFileExists(g_etcPamdCommonPassword, NULL, log))
     {
@@ -83,12 +84,11 @@ int CheckEnsurePasswordReuseIsLimited(int remember, char** reason, void* log)
 
     if (status)
     {
-        char* pamModule = FindPamModule(g_pamUnixSo, log);
-        if(NULL == pamModule)
+        if (NULL == (pamModule = FindPamModule(g_pamUnixSo, log)))
         {
             OsConfigCaptureReason(reason, "The PAM module '%s' is not available. Automatic remediation is not possible", g_pamUnixSo);
         }
-        free(pamModule);
+        FREE_MEMORY(pamModule);
     }
 
     return status;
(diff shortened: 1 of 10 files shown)


robertwoj-microsoft:
@microsoft-github-policy-service agree company="Microsoft"

src/common/commonutils/PassUtils.c
```diff
@@ -81,9 +81,14 @@ int CheckEnsurePasswordReuseIsLimited(int remember, char** reason, void* log)
             g_etcPamdCommonPassword, g_etcPamdSystemAuth, g_remember);
     }
 
-    if (status && (false == FindPamModule(g_pamUnixSo, log)))
+    if (status)
     {
-        OsConfigCaptureReason(reason, "The PAM module '%s' is not available. Automatic remediation is not possible", g_pamUnixSo);
+        char* pamModule = FindPamModule(g_pamUnixSo, log);
```
MariusNi, line 86: >char* pamModule [](http://example.com/codeflow?start=8&length=15)

Move declaration of local variable 'pamModule' at the beginning of the function, just above line 64.

Here combine the two lines in one to keep consistent with existing code:

if (NULL == (pamModule = FindPamModule(g_pamUnixSo, log)))

src/common/commonutils/SshUtils.c
```diff
@@ -292,6 +292,7 @@ static int CheckOnlyApprovedMacAlgorithmsAreUsed(const char* macs, char** reason
         return status;
     }
 
+    sshMacs = DuplicateStringToLowercase(g_sshMacs);
```
MariusNi, line 295: Add another 'else if' case after line 293, check there for successful allocation for sshMacs and if failed, log and return ENOMEM. Like this:

else if (NULL == (sshMacs = DuplicateStringToLowercase(g_sshMacs)))
    {
        OsConfigLogError(log, "CheckOnlyApprovedMacAlgorithmsAreUsed: failed to duplicate string to lowercase");
        return ENOMEM;
    }

src/common/commonutils/UrlUtils.c
```diff
@@ -12,7 +12,7 @@ char* UrlEncode(char* target)
 
     int i = 0, j = 0;
     int targetLength = (int)strlen(target);
-    int encodedLength = 3 * targetLength;
+    int encodedLength = 3 * targetLength + 1;
```
MariusNi, line 15: >3 * targetLength + 1 [](http://example.com/codeflow?start=24&length=20)

Add parenthesis like below (we try to never rely on just operator precedence):

 int encodedLength = (3 * targetLength) + 1;

src/modules/commandrunner/tests/CommandRunnerTests.cpp
```diff
@@ -127,6 +127,7 @@ namespace Tests
 
         EXPECT_EQ(MMI_OK, m_commandRunner->Get(m_component, m_reportedObject, &reportedPayload, &payloadSizeBytes));
         EXPECT_TRUE(IsJsonEq(Command::Status::Serialize(status), std::string(reportedPayload, payloadSizeBytes)));
+        delete[] reportedPayload;
```
MariusNi, line 130: >delete[] reportedPayload; [](http://example.com/codeflow?start=8&length=25)

reportedPayload is a std::string which when going out of scope, its destructor automatically frees memory. Can you please double-check on this? Same for the other similar instances in this file.

src/modules/configuration/src/lib/Configuration.c
```diff
@@ -90,6 +90,7 @@ static char* LoadConfigurationFromFile(const char* fileName)
         g_iotHubManagementEnabled = IsIotHubManagementEnabledInJsonConfig(jsonConfiguration);
         g_iotHubProtocol = GetIotHubProtocolFromJsonConfig(jsonConfiguration, ConfigurationGetLog());
         g_gitManagementEnabled = GetGitManagementFromJsonConfig(jsonConfiguration, ConfigurationGetLog());
+        FREE_MEMORY(g_gitBranch);
```
MariusNi, line 93: >FREE_MEMORY(g_gitBranch); [](http://example.com/codeflow?start=8&length=25)

Add one empty line as separator before the FREE_MEMORY line to split this into two distinct blocks.

src/platform/tests/PlatformTests.cpp
```diff
@@ -26,12 +26,7 @@ namespace Tests
         }
         else
         {
-            char* handle = new (std::nothrow) char[strlen(g_mockHandle) + 1];
-            if (handle != nullptr)
-            {
-                strcpy(handle, g_mockHandle);
-            }
-            return (MPI_HANDLE)handle;
+            return (MPI_HANDLE)strdup(g_mockHandle);
```
MariusNi, line 29: >strdup [](http://example.com/codeflow?start=31&length=6)

Even better, use our own DuplicateString (which today is same as strdup for a constant string, but in the future may gain more abilities).

MariusNi:
>         UNUSED(handle);

Need to free the handle that was allocated by MockCallMpiOpen here, by calling FREE_MEMORY

---
Refers to: src/platform/tests/PlatformTests.cpp:35 in 8fecf4d. [](commit_id = 8fecf4dfaa583c6935e74d51d095ee7fa5032475, deletion_comment = False)

MariusNi:
@robertwoj-microsoft, please add a description to the PR (in place of the guidance: 'Describe your changes in as much detail as possible. Provide a link/reference to the issue solved with this request if any.'). 

Also, since this PR does not add any unit-tests, update that template check to only say: 'All unit tests are passing.' (cut: 'I added unit-tests to validate my changes. '). Because of the nature of change, not adding new unit-tests is acceptable.

src/platform/tests/PlatformTests.cpp
robertwoj-microsoft: I'm aware of DuplicateString, but used the builtin here because it's only a tests and I don't see much point in using wrapper here.

MariusNi: Thanks for your explanation! Our test code should make use of our internal common APIs as much as possible, to ensure greatest coverage. It's not a big deal, but please consider replacing this. If you decide to keep built-in strdup, that's fine with me. Please proceed whichever way you find more appropriate.

src/common/commonutils/SshUtils.c
```diff
@@ -358,6 +363,10 @@ static int CheckAppropriateCiphersForSsh(const char* ciphers, char** reason, voi
     else if (0 != IsSshServerActive(log))
     {
         return status;
+    } else if(NULL == (sshCiphers = DuplicateStringToLowercase(g_sshCiphers)))
```
MariusNi, line 366: >else if(NULL == (sshCiphers = DuplicateStringToLowercase(g_sshCiphers))) [](http://example.com/codeflow?start=6&length=72)

Please break this line like this to be consistent with existing code:

}
else if(NULL == (sshCiphers = DuplicateStringToLowercase(g_sshCiphers)))
{

Closing curly brace to be like on line 364, for example.

src/modules/commandrunner/tests/CommandRunnerTests.cpp
MariusNi: Good catch! Please go ahead with the proposed change.

src/platform/tests/PlatformTests.cpp
robertwoj-microsoft: Keeping strdup here as discussed.

robertschaedler3 (approved):

PR #770, opened 2024-10-16
Adding Rule Ids for ABS v2 (from ASB v1) via MC reason codes

MariusNi:
## Description

Adding Rule Ids for ABS v2 (from ASB v1) via MC reason codes. Also fixing couple of security issues:

- Some temporary files created with unrestricted access could be compromised by attackers. Added access restriction.
- Reading from system files such as ones under /proc that do not have size set but contain data needs a refactoring of an old common function since IoT Hub times in order to work with all files, including these special ones.

## Checklist


src/common/commonutils/FileUtils.c
```diff
@@ -38,6 +37,22 @@ char* LoadStringFromFile(const char* fileName, bool stopAtEol, void* log)
...
+                        {
+                            string = temp;
+                            memset(&string[i], 0, currentSize - i);
+                        }
+                        else
+                        {
+                            FREE_MEMORY(string);
+                            break;
+                        }
+                    }
```
jepio, line 55: is there any test that exercises this new code path (ie. reads more than 1024 bytes)?

src/common/tests/CommonUtilsUT.cpp
```diff
@@ -94,6 +94,19 @@ TEST_F(CommonUtilsTest, LoadStringFromSingleByteFile)
     EXPECT_TRUE(Cleanup(m_path));
 }
 
+TEST_F(CommonUtilsTest, LoadStringFromZeroLengthFile)
```
jepio, line 97: Zero length files worked even before this change. We need a new test case that specifically targets "zero length but non-empty files" (like procfs).

src/common/commonutils/FileUtils.c
```diff
@@ -20,15 +22,12 @@ char* LoadStringFromFile(const char* fileName, bool stopAtEol, void* log)
     {
         if (LockFile(file, log))
         {
-            fseek(file, 0, SEEK_END);
-            fileSize = ftell(file);
-            fseek(file, 0, SEEK_SET);
-
-            string = (char*)malloc(fileSize + 1);
-            if (string)
+            if (NULL != (string = (char*)malloc(initialSize)))
```
jepio, line 25: The `(char *)` cast is present here but not on line 45 with the `realloc` call. I'd say drop it here for consistency (it's not necessary).

jepio (commented):
Can you separate the RuleId change from the bugfixes, they are not related to this change.

Can you cleanup the commits - group related changes, add descriptive titles and descriptions - it helps people trying to get familiar with the project.

This changes look significant - won't this trigger a need to perform DI re-validation?

src/common/commonutils/FileUtils.c
```diff
@@ -58,6 +73,8 @@ static bool SaveToFile(const char* fileName, const char* mode, const char* paylo
 
     if (fileName && mode && payload && (0 < payloadSizeBytes))
     {
+        RestrictFileAccessToCurrentAccountOnly(fileName);
```
jepio, line 76: Is `SaveToFile` only used for existing files, or can the file be created by the `fopen` call that happens later? Looking at it from a logical point of view:
- if the file already exists you're changing the permissions (who says they should be restricted?)
- if the file doesn't exist then the `Restrict` call happens too early and will silently fail

src/common/commonutils/FileUtils.c
```diff
@@ -932,6 +949,8 @@ int ReplaceMarkedLinesInFile(const char* fileName, const char* marker, const cha
         {
             if (NULL != (tempHandle = fopen(tempFileName, "w")))
             {
+                RestrictFileAccessToCurrentAccountOnly(tempFileName);
```
jepio, line 952: This is racy - the file is first created with potentially wide open permissions and later restricted, so there is a window where it can be manipulated.

src/common/commonutils/FileUtils.c
MariusNi: I've tested in the real end to end scenarios with the Watchers, the Agent, the Platform, and the Security Baseline etc with both 1024 and very short buffer size, suboptimal for run time, but good for spotting regressions, of just 5 bytes, for example. I'm considering adding a new test case that tests specifically for 1024+ bytes buffers. Thanks for the feedback.

MariusNi: There will be more PRs following up that will add more test coverage.

src/common/tests/CommonUtilsUT.cpp
MariusNi: Once this will be merged, more PRs will follow to expand the test coverage. For this test case that you suggested in particular, please feel free to propose one via a new PR. unittests are not functional tests and crafting such a file of 0 length but which contains data needs to be done in the test itself, not rely on profcs, unless it is something that works on all target distros.

MariusNi: Thanks for your feedback!

src/common/commonutils/FileUtils.c
MariusNi: Well spotted, thanks! Will do

src/common/commonutils/FileUtils.c
MariusNi: Added here, see CommonUtilsTest!LoadStringFromBigLengthFile.

src/common/commonutils/FileUtils.c
MariusNi: Only files that already exist. One core principle for our code is that all files that we write to are restricted for access. The system files that we write to as part of the security baseline have their access permissions set separately.

src/common/commonutils/FileUtils.c
MariusNi: Please see last iteration. The file is now emptied after being access restricted, before writing to it.

MariusNi:
> Can you separate the RuleId change from the bugfixes, they are not related to this change.
> 
> Can you cleanup the commits - group related changes, add descriptive titles and descriptions - it helps people trying to get familiar with the project.
> 
> The changes look significant - won't this trigger a need to perform DI re-validation?

As discussed, these fixes are taken together and are validated together with the refactor for rule ids. Yes, we will have a new validation and signoff on this. Let's continue this conversation there.

src/common/commonutils/FileUtils.c
```diff
@@ -5,13 +5,15 @@
...
+    int currentSize = 0;
     FILE* file = NULL;
-    int fileSize = 0;
     int i = 0;
     int next = 0;
     char* string = NULL;
+    char* temp = NULL;
 
-    if ((NULL == fileName) || (-1 == access(fileName, F_OK)))
+    if (false == FileExists(fileName))
```
robertschaedler3, line 16: nit: `if (!FileExists(fileName))`

robertschaedler3 (approved):

src/common/commonutils/FileUtils.c
MariusNi: Returns bool and if you look at other instances in this file, same explicit comparison is used.

src/adapters/mc/OsConfigResource.c
```diff
@@ -484,7 +490,13 @@ void MI_CALL OsConfigResource_Invoke_GetTargetResource(
     MI_UNREFERENCED_PARAMETER(instanceName);
 
     const char* auditPassed = "Audit passed";
-    const char* auditFailed = "Audit failed. See /var/log/osconfig*";
+    const char* auditFailed = "Audit failed. See See /var/log/osconfig_nrp.*";
```
AhmedBM, line 493: Duplicated "See"

MariusNi: Good catch! Will fix.

MariusNi: Done, please see last iteration

AhmedBM (approved):
👌

src/adapters/mc/asb/AzureLinuxBaseline.mof
```diff
@@ -1222,6 +1301,7 @@ instance of OsConfigResource as $OsConfigResource78ref
 instance of OsConfigResource as $OsConfigResource79ref
 {
     ResourceID = "Ensure all bootloaders have password protection enabled";
```
simonjaeger, line 1303: Is it OK that the ResourceID is not mapping exactly with the corresponding Name ("All bootloaders should have password protection enabled" in this instance)? Is that agreed?

src/adapters/mc/asb/AzureLinuxBaseline.mof
```diff
@@ -1838,7 +1958,8 @@ instance of OsConfigResource as $OsConfigResource119ref
 
 instance of OsConfigResource as $OsConfigResource120ref
 {
-    ResourceID = "Ensure the isdnutils-base package is not installed";
+    ResourceID = "Ensure the isdn service is disabled and the isdnutils-base package is not installed";
+    ruleId = "51ebf409-911a-4d92-9d3a-1e331e7e4b27";
```
simonjaeger, line 1962: nit: `ruleId` - should be PascalCase?

src/adapters/mc/ssh/LinuxSshServerSecurityBaseline.mof
```diff
@@ -305,6 +324,7 @@ instance of OsConfigResource as $OsConfigResource18ref
 instance of OsConfigResource as $OsConfigResource19ref
 {
     ResourceID = "Ensure that the SSH port is configured";
+    RuleId = "1310042D-193F-414A-A7E3-B2C4E89AA537";
```
simonjaeger, line 327: Just curious. Is this a new `RuleId`? It looks different from others (also uppercase).

src/common/asb/Asb.c
```diff
@@ -629,6 +629,222 @@ static char* g_desiredEnsureDefaultDenyFirewallPolicyIsSet = NULL;
...
+    { "Ensure the auditd package is installed", "f9fd03d2-75e4-4564-84a9-4e955f1e7c30", "EnsureAuditdInstalled" },
+    { "Ensure the auditd service is running", "c146c4a4-5eb6-4205-88da-5a71a82f2d45", "EnsureAuditdServiceIsRunning" },
+    { "Ensure the SNMP Server is not enabled", "ca1aea32-3969-49ab-abfc-2c5796a9a8bb", "EnsureSnmpServerIsDisabled" },
+    { "Ensure the rsync service is not enabled", "63aba603-b1f8-40df-82c5-38915452ce23", "EnsureRsynServiceIsDisabled" },
+    { "Ensure the NIS server is not enabled", "b4ad3fdd-7b68-4b11-a3ed-84b37a68b995", "EnsureNisServerIsDisabled" },
+    { "Ensure the rsh client is not installed", "6d441f31-f888-4f4f-b1da-7cfc26263e3f", "EnsureRshClientNotInstalled" },
+    { "Ensure SMB V1 with Samba is disabled", "7624efb0-3026-4c72-8920-48d5be78a50e", "EnsureSmbWithSambaIsDisabled" }
+};
+
+int AsbIsValidResourceIdRuleId(const char* resourceId, const char* ruleId, const char* payloadKey, void* log)
```
simonjaeger, line 811: Is this meant to reject any `RuleId` which is not known? Is there a concern with hard-coding that?

src/common/asb/Asb.c
```diff
@@ -629,6 +629,222 @@ static char* g_desiredEnsureDefaultDenyFirewallPolicyIsSet = NULL;
...
+    for (i = 0; i < numRules; i++)
+    {
+        if (0 == strcmp(payloadKey, (g_rules[i]).payloadKey))
+        {
+            if ((NULL != resourceId) && (0 != strcmp(resourceId, (g_rules[i]).resourceId)))
+            {
+                OsConfigLogError(log, "AsbIsValidRuleIdAndName: resourceId for rule '%s' of '%s' (instead of '%s') is invalid", payloadKey, resourceId, (g_rules[i]).resourceId);
+                result = ENOENT;
+            }
+            else if ((NULL != ruleId) && (0 != strcmp(ruleId, (g_rules[i]).ruleId)))
```
simonjaeger, line 832: Is it possible to use `strncmp` here as `RuleId` length is known?

src/tests/universal-nrp-e2e/UniversalNRP.Tests.ps1
```diff
@@ -32,13 +32,11 @@ Describe 'Validate Universal NRP' {
 
         It 'Ensure resons are properly populated' {
```
simonjaeger, line 33: Should there also be a test to make sure that `RuleId` is included? Seems like the below test is checking for `BaselineSettingCompliant` and `BaselineSettingNotCompliant` only.

simonjaeger (approved):

src/tests/universal-nrp-e2e/UniversalNRP.Tests.ps1
MariusNi: The actual test in in the new function in Asb.c, AsbIsValidResourceIdRuleId. If a rule id or rule name is changed from expected, the reason code will state that and fail this test in this PS script.  The NRP itself validates that the input from MOF from MC and AzP is correct and that is invoked for this test. It works quite well actually.

MariusNi: Please reactivate if you still have questions. Thanks

src/common/asb/Asb.c
MariusNi: Any rule id and rule name that does not match the expected values for that rule. We need the reference, and the choice here is to put this into the code so the NRP can also use it at run time. The audits will still work, but the reason codes will  signal the problems. Please reactivate this if you still have questions.

src/common/asb/Asb.c
MariusNi: Good catch. Will do change.

src/adapters/mc/ssh/LinuxSshServerSecurityBaseline.mof
MariusNi: The new rule ids are all in clean all uppercase characters. There are total of 6 such rules here. This is one subtle way to tell which ones are, for example.

MariusNi: Please reactivate if you still have questions or comments

src/adapters/mc/asb/AzureLinuxBaseline.mof
MariusNi: Good catch! This works as is but yes, it should be PascalCase, this is a typo. Will fix.

src/adapters/mc/asb/AzureLinuxBaseline.mof
MariusNi: All names in here follow a similar consistent pattern. There are some slight changes from v1 etc. Names will be one last time reviewed before release.

src/adapters/mc/asb/AzureLinuxBaseline.mof
MariusNi: Fixed. Please see last iteration. Thanks for the feedback!

src/common/asb/Asb.c
MariusNi: Fixed. Please see last iteration. Thanks for the feedback!

simonjaeger (approved):

simonjaeger (approved):

PR #771, opened 2024-10-18
Fixing ASB v2 temporary file creation race condition that may lead to /etc/shadow file exposure to unprivileged users

MariusNi:
## Description

Fixing ASB v2 temporary file creation race condition that may lead to /etc/shadow file exposure to unprivileged users. There is a window of time during which an unprivileged user may get a file descriptor of our temp file, wait for us to finish writing to it, and then read. That file content we just wrote may be of protected files such as /etc/shadow, that are not normally allowed to be read by normal users.

Many thanks to [robertwoj-microsoft](https://github.com/robertwoj-microsoft) who spotted this bug and proposed the fix.

The fix is being tested with the CI, plus manual running of unit-tests, functional test recipe tests, and also building and deploying a custom policy package containing the fix and validating results of that.

## Checklist

### DIFF
--- a/src/common/commonutils/FileUtils.c
+++ b/src/common/commonutils/FileUtils.c
@@ -1003,6 +1003,8 @@ int ReplaceMarkedLinesInFile(const char* fileName, const char* marker, const cha
                     OsConfigLogError(log, "ReplaceMarkedLinesInFile: failed to open temporary file '%s', fdopen() failed (%d)", tempFileName, errno);
                     status = EACCES;
                 }
+
+                close(tempDescriptor);
             }
             else
             {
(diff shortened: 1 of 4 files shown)


AhmedBM (approved):
👌

robertschaedler3 (approved):

robertwoj-microsoft (approved):

src/common/commonutils/FileUtils.c
```diff
@@ -998,13 +1000,15 @@ int ReplaceMarkedLinesInFile(const char* fileName, const char* marker, const cha
                 }
```
robertwoj-microsoft, line 1000: In the original fix proposal, the flose(tempHandlee) was removed. This line introduces doble close, because the file descriptor is shared by the file stream pointer returned by the fdopen(tempDescriptor) call.

MariusNi: Taking a look now, fix will be in a new PR.

MariusNi: [https://github.com/Azure/azure-osconfig/pull/773](https://github.com/Azure/azure-osconfig/pull/773)

PR #772, opened 2024-10-21
Add sanitizers build github workflow

robertwoj-microsoft:
## Description

Create a new github actions workflow which builds the source code with address sanitizer and performs unit tests. The new workflow is currently enabled only as workflow_dispatch, because it needs to be merged first to be picked up by github actions.
There should come a second stage when pull_request and schedule jobs are enabled when the new workflow is picked up and is confirmed to work.

## Checklist


.github/workflows/ci-sanitizers.yml
```diff
@@ -0,0 +1,128 @@
+name: CI with sanitizers enabled
+
+on:
+  # TODO: Enable pull request and cron triggers later, when this workflow is tested and not failing
```
MariusNi, line 4: >TODO:  [](http://example.com/codeflow?start=4&length=6)

We do not allow code with TODO comments to be checked in main. Please complete this part now, or remove the comment and create a feature task for the team.

MariusNi:
> 

New line  33 is blank, please remove.

---
Refers to: devops/docker/ubuntu-22.04-amd64/Dockerfile:33 in 48c7a95. [](commit_id = 48c7a95480945176d8d471b6fac1e63adcf34be5, deletion_comment = False)

.github/workflows/ci-sanitizers.yml
```diff
@@ -0,0 +1,128 @@
...
+          tag: ${{ matrix.target.tag }}
+
+      - name: Generate build
+        uses: ./.github/actions/container-exec
+        with:
+          container: ${{ steps.container.outputs.id }}
+          cmd: |
+            mkdir build && cd build
+            cmake ../src -DCMAKE_C_FLAGS="${{ env.SANITIZER_FLAGS }}" -DCMAKE_CXX_FLAGS="${{ env.SANITIZER_FLAGS }}" -DCMAKE_build-type=${{ env.BUILD_TYPE }} -Duse_prov_client=ON -Dhsm_type_symm_key=ON -DCOMPILE_WITH_STRICTNESS=ON -DBUILD_TESTS=ON -DBUILD_SAMPLES=ON -DBUILD_ADAPTERS=ON -Duse_default_uuid=ON
+
```
MariusNi, line 79: This build line is for everything, including for source that is not ours and that we build here. Such as the Azure PnP C SDK, or Parson. We do not want sanitizer to be run on that code and block a PR/fail CI for problems found there.

How are we going to avoid this kind of action?

.github/workflows/ci-sanitizers.yml
```diff
@@ -0,0 +1,128 @@
...
+          container: ${{ steps.container.outputs.id }}
+          working-directory: ${{ env.MOUNT }}/build
+          cmd: cmake --build . --config ${{ env.BUILD_TYPE }} --parallel
+
+      - name: Run ctest
+        uses: ./.github/actions/container-exec
+        with:
+          container: ${{ steps.container.outputs.id }}
+          working-directory: ${{ env.MOUNT }}/build
+          cmd: ctest --verbose > ../${{ matrix.target.name }}-${{ matrix.target.arch }}-sanitizer.log
```
MariusNi, line 92: >cmd: ctest --verbose [](http://example.com/codeflow?start=10&length=20)

This invokes the unit-tests only, correct? Would be possible to also invoke for the module test recipe functional tests? Not necessarily in this PR, possibly later.

.github/workflows/ci-sanitizers.yml
```diff
@@ -0,0 +1,128 @@
...
+    outputs:
+      matrix: ${{ steps.get-matrix.outputs.matrix }}
+    steps:
+      - name: Get matrix
+        id: get-matrix
+        # Skip centos-7, oraclelinux-7, rhel-7 as they don't currently build properly from Dockerfile
+        run: |
+          matrix="$(cat <<'EOL'
+          [
+            { "name": "ubuntu-22.04", "arch": "amd64", "tag": "latest" }
```
MariusNi, line 29: >{ "name": "ubuntu-22.04", "arch": "amd64", "tag": "latest" } [](http://example.com/codeflow?start=12&length=60)

Not for this PR now, but to consider for the future. We build OSConfig for lots more distros. There are differences among them for the code that is executed on each, esp. for functional tests. Also, AMD64-only is only for the MC scenarios. For OSConfig in general we also plan to re-add ARM64 in the future and in the past we also supported x86 and ARMHF/32 bit.

AhmedBM (commented):
Can we just change our current CI pipeline to do this?
Our current CI pipeline does include many other distros, are these sanitizers supported across as far back as GCC 4.8?

Marius: as Robert described, we need debug build for sanitizer, so for this reason alone, separate workflow makes sense.

.github/workflows/ci-sanitizers.yml
robertwoj-microsoft: In my opinion, we should avoid such approach. If there is anything that's caught by sanitizers in 3rd party, it means the bug is reachable in our codebase and should be fixed before merging a particular PR. In addition to that, the first enablement of sanitizers should not trigger any bugs at current main, so I don't expect anything to show up within 3rd parties.

.github/workflows/ci-sanitizers.yml
robertwoj-microsoft: OK, as in the comment, this cannot be done now since the workflow must first appear in the main branch. pull_request and schedule will be done in a separate PR and I'll drop this comment for now.

.github/workflows/ci-sanitizers.yml
robertwoj-microsoft: I initially intended to include it on all platforms. but not all the images build now and fixing that is beyound this PR scope.

.github/workflows/ci-sanitizers.yml
robertwoj-microsoft: Yes, module tests are not yet run by this workflow, planned for later when we see how this one works for us.

.github/workflows/ci-sanitizers.yml
MariusNi: We cannot easily fix code in the PnP C SDK, for example. The team that used to own that no longer exists. For us, that code is now disabled by default and only optionally enabled, however is still included in builds. Best is to see what happens, wanted to rase awareness of this however.

.github/workflows/ci-sanitizers.yml
MariusNi: Sounds good.

.github/workflows/ci-sanitizers.yml
```diff
@@ -0,0 +1,128 @@
+name: CI with sanitizers enabled
```
MariusNi, line 1: >CI with sanitizers enabled [](http://example.com/codeflow?start=6&length=26)

As commented by Ahmed in discussion, can we instead include this into the existing CI workflow? Rather than creating a separate workflow?

MariusNi: Build in debug mode for sanitizer is a good argument to keep this a separate workflow. Thanks for explaining!

devops/docker/ubuntu-22.04-amd64/Dockerfile
```diff
@@ -21,10 +21,12 @@ RUN apt -y update && apt-get -y install \
...
+    libasan8 \
+    libubsan1
 
 WORKDIR /git
 
 # CMake
 RUN git clone https://github.com/Kitware/CMake --recursive -b v3.21.7
-RUN cd CMake && ./bootstrap && make -j$(nproc) && make install && hash -r && rm -rf /git/CMake
\ No newline at end of file
+RUN cd CMake && ./bootstrap && make -j$(nproc) && make install && hash -r && rm -rf /git/CMake
```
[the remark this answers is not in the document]
MariusNi: Why is this unnecessary?

MariusNi (approved):
:shipit:

robertschaedler3 (approved):

AhmedBM (approved):

PR #773, opened 2024-10-21
Avoiding double close on temp file handle

MariusNi:
## Description

If we call both close() and fclose() on the same file descriptor and file stream, it can lead to undefined behavior:
- fclose(): This function flushes the stream, writes any buffered output data, and closes the underlying file descriptor. After calling fclose(), any further access to the stream results in undefined behavior.
- close(): This function closes the file descriptor. If we call close() after fclose(), we are attempting to close an already closed file descriptor, which can lead to errors or unexpected behavior.

## Checklist

### DIFF
--- a/src/common/commonutils/FileUtils.c
+++ b/src/common/commonutils/FileUtils.c
@@ -1000,11 +1000,11 @@ int ReplaceMarkedLinesInFile(const char* fileName, const char* marker, const cha
                 }
                 else
                 {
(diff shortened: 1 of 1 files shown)


robertwoj-microsoft (approved):

AhmedBM (approved):
👌

robertschaedler3 (approved):

PR #781, opened 2024-10-28
Fix possible NULL dereference when lscpu command fails

robertwoj-microsoft:
## Description

In case the lscpu command fails or is not present on a system, the rule called auditEnsureKernelSupportForCpuNx will trigger 
GetCpuFlags function call which returns NULL in this case. This commit adds additional NULL check for the cpuFlags variable to avoid passing NULL pointer to strstr function.

## Checklist

### DIFF
--- a/src/common/commonutils/DeviceInfoUtils.c
+++ b/src/common/commonutils/DeviceInfoUtils.c
@@ -328,7 +328,7 @@ bool CheckCpuFlagSupported(const char* cpuFlag, char** reason, void* log)
     bool result = false;
     char* cpuFlags = GetCpuFlags(log);
 
-    if ((NULL != cpuFlag) && (NULL != strstr(cpuFlags, cpuFlag)))
+    if ((NULL != cpuFlag) && (NULL != cpuFlags) && (NULL != strstr(cpuFlags, cpuFlag)))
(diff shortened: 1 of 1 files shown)


robertschaedler3 (approved):

src/common/commonutils/DeviceInfoUtils.c
```diff
@@ -328,7 +328,7 @@ bool CheckCpuFlagSupported(const char* cpuFlag, char** reason, void* log)
     bool result = false;
     char* cpuFlags = GetCpuFlags(log);
 
-    if ((NULL != cpuFlag) && (NULL != strstr(cpuFlags, cpuFlag)))
+    if ((NULL != cpuFlag) && (NULL != cpuFlags) && (NULL != strstr(cpuFlags, cpuFlag)))
```
MariusNi, line 331: >(NULL != cpuFlags) && (NULL !=  [](http://example.com/codeflow?start=29&length=31)

On what distro did you find the lscpu command to not be present or to fail? Was that a default original configuration of that distro, or was something done to break it? Can you please give more details about how this bug was found.

MariusNi (approved):
:shipit:

PR #785, opened 2024-10-30
Initial fuzzer implementation

robertwoj-microsoft:
## Description

This branch contains an initial fuzzer implementation. It's been made extensible, so further targets will be added later.

## Checklist


src/CMakeLists.txt
```diff
@@ -115,6 +116,19 @@ endif()
 if (BUILD_MODULES)
     add_subdirectory(modules)
 endif()
+if (BUILD_FUZZER)
```
jepio, line 119: Move this whole block before line 108 (before the `add_subdirectory` calls) otherwise the compile options are not applied the way you intended.

jepio:
Hint: **don't** run this unless you are in a container. My suggestion is to create one that matches your host distro, install llvm and then fuzz, for me that's:
```
docker run -it --rm -v /home/jeremi/git:/home/jeremi/git -w /home/jeremi/git/azure/azure-osconfig/build/fuzzer debian:12
```

src/fuzzer/seed_corpus/SecurityBaselineMmiGet.auditEnsureAcceptingSourceRoutedPacketsIsDisabled.target
```diff
@@ -0,0 +1 @@
+SecurityBaselineMmiGet.auditEnsureAcceptingSourceRoutedPacketsIsDisabled
```
MariusNi, line 1: >ecurityBaselineMmiGet.auditEnsureAcceptingSourceRoutedPacketsIsDisabled [](http://example.com/codeflow?start=1&length=71)

There is no need to fuzz at this level, per individual audit* and remediate* call in asb lib. It is 100% the same to audit the exports from asb lib, the AsbMmiGet and AsbMmiSet, which directly translate to these. 

Both have same pros and cons, byt doing it this way complicates things for no reason.

Please consider:

1. Fuzz at commonutils, not asb, level.
2. If at asb lib level, fuzz the top exports, not the individual rules.

src/fuzzer/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
+cmake_minimum_required(VERSION 3.2)
+
+# Set the project name and version
+project(osconfig-fuzzer VERSION 1.0)
+
+# Specify the C++ standard
+set(CMAKE_CXX_STANDARD 17)
```
MariusNi, line 7: Is this C++ version required for fuzzing?

Also, can we make this here conditional on clang so it won't execute for gcc?

src/fuzzer/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
...
+set(CMAKE_CXX_STANDARD_REQUIRED True)
+
+find_package(LLVM REQUIRED)
+
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=fuzzer")
+
+add_executable(osconfig-fuzzer target.cpp)
```
MariusNi, line 14: >target.cpp [](http://example.com/codeflow?start=31&length=10)

Can we do the target in C, like the fuzzed code is?

src/fuzzer/seed_corpus/SecurityBaselineMmiGet.auditEnsureAcceptingSourceRoutedPacketsIsDisabled.target
robertwoj-microsoft: As discussed, the list of targets is going to change in future, I am going to go over your email suggestions and add them up accordingly. Can you clarify what do you mean by 'top exports'?

src/fuzzer/CMakeLists.txt
robertwoj-microsoft: It already is conditional, see the src/CMakeLists.txt. C++ 17 is not required for fuzzing, but unless there is a need to go down, let's keep it reasonably fresh.

src/fuzzer/CMakeLists.txt
robertwoj-microsoft: Why would we do that? This is testing code and we benefit from some goodness of the language like STL containers.

src/fuzzer/seed_corpus/SecurityBaselineMmiGet.auditEnsureAcceptingSourceRoutedPacketsIsDisabled.target
MariusNi: By top exports for asb library mean AsbIsValidResourceIdRuleId and AsbMmiGet. Originally in my comment I stated, quote: 'exports from asb lib, the AsbMmiGet and AsbMmiSet' but AsbMmiSet is too dangerous to fuzz inputs on. Think we all agreed on the e-mail exchange on this and now this conversation can be closed.

src/fuzzer/CMakeLists.txt
MariusNi: Do we need STL containers and more such complex types for the fuzzing test instrumentation? Maybe I could see some in a next commit as an example of your reply and understand better.

src/fuzzer/CMakeLists.txt
MariusNi: Please see requirements for overall project:

set (CMAKE_C_STANDARD 11)
set (CMAKE_CXX_STANDARD 11)

If we need C++ 17 for fuzzing, let's isolate it so it won't be set unless building with clang (so we won't enter here when building with gcc and C++ 11)

MariusNi: (even if the fuzzer code will be moved elsewhere in project, let's not depend on another condition and make a condition here, to say it explicit)

robertwoj-microsoft: I changed to C++ 11

src/fuzzer/CMakeLists.txt
robertwoj-microsoft: For simplicity and to get this project up quick, I decided to go with C++ which is fine to be used in testing code. I use standard library primitives and feel more confident and robust using them. I don't see a reason for going with plain C here and there is no such requirement that I would be aware. Therefore, I don't intend to use C here.

src/common/commonutils/FileUtils.c
```diff
@@ -120,25 +120,52 @@ bool SavePayloadToFile(const char* fileName, const char* payload, const int payl
...
+    {
+        OsConfigLogError(log, "AppendPayloadToFile: invalid arguments");
+        return false;
+    }
 
-    // If the file exists and there is no EOL at the end of file, add one before the append
-    if ((NULL != payload) && (payloadSizeBytes > 0) && FileExists(fileName) && 
-        (NULL != (fileContents = LoadStringFromFile(fileName, false, log))) && 
-        (EOL != fileContents[strlen(fileContents) - 1]))
+    /* If the file exists and there is no EOL at the end of file, add one before the append. */
```
MariusNi, line 134: >/* If the f [](http://example.com/codeflow?start=4&length=11)

Keep consistent with commenting style for this project. We use C++ style comments.

src/common/commonutils/FileUtils.c
```diff
@@ -120,25 +120,52 @@ bool SavePayloadToFile(const char* fileName, const char* payload, const int payl
...
+    if (stat(fileName, &st) == 0 && st.st_size > 0)
```
MariusNi, line 135: >stat(fileName, &st) == 0 && st.st_size > 0) [](http://example.com/codeflow?start=8&length=43)

Inner parenthesis and move constant to the left in comparisons to keep with existing code style

src/common/commonutils/FileUtils.c
```diff
@@ -120,25 +120,52 @@ bool SavePayloadToFile(const char* fileName, const char* payload, const int payl
...
+    if(NULL == fileName || NULL == payload || 0 == payloadSizeBytes)
+    {
+        OsConfigLogError(log, "AppendPayloadToFile: invalid arguments");
+        return false;
+    }
 
-    // If the file exists and there is no EOL at the end of file, add one before the append
-    if ((NULL != payload) && (payloadSizeBytes > 0) && FileExists(fileName) && 
-        (NULL != (fileContents = LoadStringFromFile(fileName, false, log))) && 
-        (EOL != fileContents[strlen(fileContents) - 1]))
```
MariusNi, line 129: What exactly was wrong with this approach? Please elaborate. This function is used for many ASB checks which have been fully validated. Fixing one security bug found by fuzzing should be as minimal as possible. Do not refactor an entire function without strong reason and state what that is.

src/common/commonutils/FileUtils.c
```diff
@@ -81,7 +81,7 @@ static bool SaveToFile(const char* fileName, const char* mode, const char* paylo
             {
                 for (i = 0; i < payloadSizeBytes; i++)
                 {
-                    if (payload[i] != fputc(payload[i], file))
+                    if ((unsigned char)payload[i] != fputc(payload[i], file))
```
MariusNi, line 84: >fputc [](http://example.com/codeflow?start=53&length=5)

fputc returns int. Signed. Why cats to unsigned the payload byte being compared? Please explain or revert.

MariusNi (dismissed):
:clock1:

src/common/commonutils/FileUtils.c
```diff
@@ -120,25 +120,52 @@ bool SavePayloadToFile(const char* fileName, const char* payload, const int payl
 
 bool AppendPayloadToFile(const char* fileName, const char* payload, const int payloadSizeBytes, void* log)
 {
-    char* fileContents = NULL;
+    FILE* existingFile = NULL;
+    struct stat st;
     bool result = false;
+    char lastCharacter = EOF;
+
+    if(NULL == fileName || NULL == payload || 0 == payloadSizeBytes)
```
MariusNi, line 128: >if(NULL == fileName || NULL == payload || 0 == payloadSizeBytes) [](http://example.com/codeflow?start=4&length=64)

Here and on line 135, add inner parenthesis. Never rely on just operator precedence.

src/common/commonutils/FileUtils.c
robertwoj-microsoft: > Unclear here what exactly was wrong with this approach? Can you elaborate?
There are couple of bugs and issues in this single statement:
- we load the whole file into memory just to read one byte,
- we don't check whether length of the fileContents is actually greater than 0,
- we perform a complex operation like allocating memory to fileContents and we have no other choice than freeing it in different place, because it's bundled together,
- why do we check for file existence first? Is the LoadStringFromFile gonna crash on non-existent file or it will just return NULL? We have a NULL check already,
- why don't we actually do a NULL check for the payload and payload size in the very beginning of the function? Is a NULL or empty payload a valid input to the function? If providing a NULL payload is invalid, in this construct we have no way of properly logging it,
- by bundling together all these operations code becomes much less clear. For example, one needs to look into the LoadStringFromFile to check if it returns a non-empty string. Lack of clarity is a very serious issue.

>  Fixing one security bug found by fuzzing should be as minimal as possible.

If the code is complex enough I think some minor upgrades should be done as we go, otherwise we'll only be adding complexity and there will never be good time to do refactoring.

> This function is used for many ASB checks which have been fully validated.

We should put into question the quality of validation then. Maybe we ask the validation team to actually do some code review and raise questions if something is unclear.

> We shall not refactor entire functions without strong reasons and then we need to state as detailed as possible what reasons are and why each part of the respective functions needs change.

I strongly believe we have reasons to do that.

src/common/commonutils/FileUtils.c
robertwoj-microsoft: https://man7.org/linux/man-pages/man3/fgetc.3.html
> fgetc() reads the next character from stream and returns it as an
       unsigned char cast to an int, or EOF on end of file or error.

kkanas: Then it shouldn't it be like 
```
  if (payload[i] != (unsigned char)fputc(payload[i], file))
```

src/common/commonutils/FileUtils.c
MariusNi: That MI header is ingested from Windows as-is. This and the auto-generated NRP infra code use a different coding style from rest of our project.

src/common/commonutils/FileUtils.c
MariusNi: https://github.com/Azure/azure-osconfig/pull/790

src/common/commonutils/FileUtils.c
```diff
@@ -120,25 +120,52 @@ bool SavePayloadToFile(const char* fileName, const char* payload, const int payl
...
+             * Potentially we could just write this byte here using existingFile handle, but the function does some other magic.
+             * Leaving to reduce logic changes.
+             */
+            if (false == SaveToFile(fileName, "a", "\n", 1, log))
+            {
+                OsConfigLogError(log, "AppendPayloadToFile: failed to append EOL to '%s'", fileName);
+                return false;
+            }
+        }
+    }
```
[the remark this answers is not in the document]
MariusNi: https://github.com/Azure/azure-osconfig/pull/790

MariusNi (approved):
:shipit:

robertschaedler3 (approved):

PR #786, opened 2024-10-31
First set of bug fixes resulting from fuzzing (work in progress)

jepio:
## Description

First set of bugfixes found through fuzzing:
* fix leaks in some remediation paths. These remediation functions call audit functions that return an allocated reason string, but the string was not being free'd.
* fix leak of pamModulePath3 in SetPasswordCreationRequirements()

## Checklist

### DIFF
--- a/src/common/asb/Asb.c
+++ b/src/common/asb/Asb.c
@@ -3773,12 +3773,19 @@ static int RemediateEnsurePostfixNetworkListeningIsDisabled(char* value, void* l
     return DisablePostfixNetworkListening(log);
 }
 
+static int CheckAndFreeReason(char *reason)
+{
+    int ret = (0 == strncmp(g_pass, reason, strlen(g_pass))) ? 0 : ENOENT;
+    FREE_MEMORY(reason);
+    return ret;
+}
+
 static int RemediateEnsureRpcgssdServiceIsDisabled(char* value, void* log)
 {
     UNUSED(value);
     StopAndDisableDaemon(g_rpcgssd, log);
     StopAndDisableDaemon(g_rpcGssd, log);
-    return (0 == strncmp(g_pass, AuditEnsureRpcgssdServiceIsDisabled(log), strlen(g_pass))) ? 0 : ENOENT;
+    return CheckAndFreeReason(AuditEnsureRpcgssdServiceIsDisabled(log));
 }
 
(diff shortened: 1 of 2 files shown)


src/common/commonutils/SshUtils.c
```diff
@@ -205,13 +205,11 @@ static int IsSshServerActive(void* log)
 // See https://www.openssh.com/txt/release-8.2, quote: "add an Include sshd_config keyword that allows including additional configuration files"
 static int IsSshConfigIncludeSupported(void* log)
```
MariusNi, line 206: >IsSshConfigIncludeSupported [](http://example.com/codeflow?start=11&length=27)

Please see the comment on lines 226-228. 

This function was tested on all target distros, both supporting -V and not and it was found to work well. We also do not have any customer reports telling us they hit a problem with this, which is specific to the 'SSH security posture for Linux' policy which is now shipped in GA, Unless there is a clear, reproducible bug, of high priority, sorry, but I won't agree with this change.

Please elaborate on the bug repro steps so that we could repro in house and investigate. 

Until then, the other fixes look good, but won't signoff until we agree on this change here for IsSshConfigIncludeSupported.

MariusNi: If you could split this in 2 PRs, the other one will be accepted right away and then we can keep this other one for further discussion.

MariusNi (changes requested):
:clock1:

src/common/commonutils/SshUtils.c
jepio: I'm dropping this commit for now and will revisit again next week.

jepio:
Dropped the commit like you asked.

MariusNi (approved):
:shipit:

MariusNi:
> Dropped the commit like you asked.

Thank you. To be clear, this is not any personal preference. That function is super critical for SSH remediation. Its purpose is to detect if the current distro supports include for sshd_config. From our target distros some do and some don't. This check is done in a distro agnostic manner, just based on sshd server response. This function was validated through and through. Any change to it now comes with a significant risk. Unless we discover a security bug in it, on that basis alone, we should not change it. If you believe there is a problem how the sshd server's include support detection happens (what we have came out from collaboration with the SSH team but who knows, maybe we missed something?) please let me know.

robertschaedler3 (approved):

PR #791, opened 2024-11-06
Ubuntu 22.04 prerequisite for fuzzer CI

robertwoj-microsoft:
## Description

Add clang package installation which is required for implementing CI using fuzzer.

## Checklist

### DIFF
--- a/devops/docker/ubuntu-22.04-amd64/Dockerfile
+++ b/devops/docker/ubuntu-22.04-amd64/Dockerfile
@@ -23,7 +23,8 @@ RUN apt -y update && apt-get -y install \
     bc \
     file \
     libasan8 \
-    libubsan1
+    libubsan1 \
+    clang
 
 WORKDIR /git
 


devops/docker/ubuntu-22.04-amd64/Dockerfile
```diff
@@ -23,7 +23,8 @@ RUN apt -y update && apt-get -y install \
     bc \
     file \
     libasan8 \
-    libubsan1
+    libubsan1 \
+    clang
```
AhmedBM, line 27: Adding `clang` will affect the CMake usage for this ubuntu-22.04, have you ensured building the project still works once you've added `clang`?

AhmedBM: im ok with adding it but i feel like we need to add a cmake toolchain just to ensure we don't affect the builds with the newly introduced compiler

robertwoj-microsoft:
```
root@f0ae2883d87e:/git# cmake --version
cmake version 3.21.7

CMake suite maintained and supported by Kitware (kitware.com/cmake).
root@f0ae2883d87e:/git# gcc --version
gcc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
Copyright (C) 2021 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

root@f0ae2883d87e:/git# clang --version
Ubuntu clang version 14.0.0-1ubuntu1.1
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
root@f0ae2883d87e:/git# cat CMakeLists.txt 
cmake_minimum_required(VERSION 3.2)
project(test)
set(CMAKE_C_STANDARD 11)
set(CMAKE_C_STANDARD_REQUIRED True)
add_executable(test main.c)
root@f0ae2883d87e:/git# cat main.c 
int main()
{
        print("Hello world!\n");
}
root@f0ae2883d87e:/git# rm -fr CMakeFiles; rm -f CMakeCache.txt 
root@f0ae2883d87e:/git# cmake . | grep "C compiler"
-- The C compiler identification is GNU 11.4.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
root@f0ae2883d87e:/git# ls -l /usr/bin/cc
lrwxrwxrwx 1 root root 20 Nov  5 09:38 /usr/bin/cc -> /etc/alternatives/cc
root@f0ae2883d87e:/git# ls -l /etc/alternatives/cc 
lrwxrwxrwx 1 root root 12 Nov  5 09:38 /etc/alternatives/cc -> /usr/bin/gcc
```

The following commands have been executed on the container built from the Dockerfile I have modified.
See that we have both compilers installed, cmake picks up gcc by default.

robertwoj-microsoft: @AhmedBM I've added explicit flags for selecting GCC. Let me know if you think that's fine.

robertwoj-microsoft: @AhmedBM On a side note: We specify bulid commands in more than one place: package-build.yml, ci.yml and ci-sanitizers.yml (last one I added recently and is based on ci.yml). Is this intended to have it in more than one place? Would it be possible to just use package-build.yml and base unit tests on that?

MariusNi (approved):
:shipit:

devops/docker/ubuntu-22.04-amd64/Dockerfile
AhmedBM: it is intentional yes, it would have been better as a composite workflow action so we can scope that to all builds, but yes, for now its a little spread out :)

PR #792, opened 2024-11-06
CI workflow for fuzzing

robertwoj-microsoft:
## Description

CI workflow for fuzzing.

**PLASE DO NOT MERGE YET**

## Checklist


.github/workflows/fuzzing-run.yml
```diff
@@ -0,0 +1,61 @@
...
+
+      - uses: actions/download-artifact@v4
+        id: download
+        with:
+          name: ${{ inputs.target }}-fuzzer
+
+      - name: Install prerequisites
+        run: |
+          sudo apt update
+          sudo apt install -y clang
```
AhmedBM, line 45: this won't be needed once the container changes are in

robertwoj-microsoft: That's needed at runtime. I think the VM image is different than the container image we use for building. If I'm wrong, I'll drop it later when the prerequisite PR is merged.

jepio: For sure the llvm-symbolizer is needed at runtime, that's part of the llvm package which is a dependency of clang.

AhmedBM: ahh yes indeed, good point! We should also add this into our managed image!

jepio: is that controlled by `devops/e2e/cloudtest/` or in some other place?

.github/workflows/fuzzing-run.yml
```diff
@@ -0,0 +1,61 @@
...
+          sudo chmod +x ${{ steps.download.outputs.download-path }}/fuzzer/osconfig-fuzzer
+          mkdir /tmp/corpus
+          mkdir /tmp/artifacts
+          cp -r ${{ steps.download.outputs.download-path }}/src/fuzzer/seed_corpus/* /tmp/corpus
```
MariusNi, line 49: >/src/fuzzer/ [](http://example.com/codeflow?start=59&length=12)

Can we move fuzzer from under src root under src/test/? This is in order to respect our original repo structure. Is there any hard requirement for fuzzer to sit right under src?

robertwoj-microsoft: Good point, I'll move it in the next PR when variant B is done.

MariusNi (approved):
:shipit:

MariusNi:
@ robertwoj-microsoft, https://github.com/Azure/azure-osconfig/pull/791 was merged last week. Can this PR be completed please? It blocks other PRs as well if I am not mistaken.

PR #343, opened 2022-08-22
copy AIS toml at build-time

AhmedBM:
## Description

* Binplaces the AIS `osconfig.toml` at build-time to `/etc/aziot/identityd/config.d/` similar to our CMake post-install scripts

## Checklist

- [X] I have read the [contribution guidelines](https://github.com/Azure/azure-osconfig/blob/main/CONTRIBUTING.md).
- [X] I added unit-tests to validate my changes. All unit tests are passing.
- [X] I have merged the latest `main` branch prior to this PR submission.
- [X] I submitted this PR against the `main` branch.

PR #793, opened 2024-11-07
Only install aziot config if the directory exists

jepio:
## Description

Check if `/etc/aziot` exists before trying to install it during build. This change fixes a build failure on machines when the build is running as non-root. We should never be writing files to /etc/ during building and no one should be building as root.

Ideally this custom command would be removed or moved behind a cmake `option()`.

## Checklist

### DIFF
--- a/src/adapters/pnp/CMakeLists.txt
+++ b/src/adapters/pnp/CMakeLists.txt
@@ -96,9 +96,4 @@ install(FILES daemon/${target_name}.json DESTINATION ${CMAKE_INSTALL_SYSCONFDIR}
 install(FILES daemon/${target_name}.conn DESTINATION ${CMAKE_INSTALL_SYSCONFDIR}/osconfig)
 install(FILES daemon/${target_name}.toml DESTINATION ${CMAKE_INSTALL_SYSCONFDIR}/osconfig)
 install(FILES daemon/${target_name}.service DESTINATION ${CMAKE_INSTALL_SYSCONFDIR}/systemd/system)
-
-add_custom_command(
-    TARGET ${target_name}
-    POST_BUILD
-    COMMAND ${CMAKE_COMMAND} -E copy_if_different ${CMAKE_CURRENT_SOURCE_DIR}/daemon/${target_name}.toml /etc/aziot/identityd/config.d/${target_name}.toml
-)
+install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/daemon/${target_name}.toml DESTINATION /etc/aziot/identityd/config.d/)


src/adapters/pnp/CMakeLists.txt
```diff
@@ -97,8 +97,10 @@ install(FILES daemon/${target_name}.conn DESTINATION ${CMAKE_INSTALL_SYSCONFDIR}
...
-    COMMAND ${CMAKE_COMMAND} -E copy_if_different ${CMAKE_CURRENT_SOURCE_DIR}/daemon/${target_name}.toml /etc/aziot/identityd/config.d/${target_name}.toml
-)
\ No newline at end of file
+if (EXISTS /etc/aziot)
+    add_custom_command(
+        TARGET ${target_name}
+        POST_BUILD
+        COMMAND ${CMAKE_COMMAND} -E copy_if_different ${CMAKE_CURRENT_SOURCE_DIR}/daemon/${target_name}.toml /etc/aziot/identityd/config.d/${target_name}.toml
+    )
+endif()
```
MariusNi, line 106: If I remember correctly, we need this even when AIS (aka aziot services) is not yet installed. What problem did you encounter with this to make you open this PR?

Speaking of this description in PR, quote: "This change fixes a build failure on machines when the build is running as non-root. We should never be writing files to /etc/ during building and no one should be building as root."

We do in fact require the 2nd cmake build command to be ran with sudo as root when the install option is selected. We had (and still have if I am not mistaken) scenarios that ask for this. 

Let's please follow-up in team discussion before proceeding with this, to make sure we are aligned.

Also, if we were to disable the install (under /etc, /bin, /usr, etc) during build, we need to also disable such things in other CMake files. 

Thanks!

jepio: The 2nd cmake command is "install" and it's fine to expect to be running that as root/with sudo. Right now the copy command runs together with compilation which is bad.

jepio: > Maybe instead of conditioning on AIS being present make this conditional of the '--target install' option? So when that is not used, neither would be this file installed?

I've update the PR to run together with the `--target install` phase, which is better than what is currently in main but that will still fail with local cpack invocation (`cpack -G DEB`) which I regularly use and it doesn't require sudo.

> Let's please follow-up in team discussion before proceeding with this to make sure we are aligned.

I'll reach out through other channels but let's keep this kind of discussion in PR for the future. The context is useful to outsiders too.

robertschaedler3 (approved):

src/adapters/pnp/CMakeLists.txt
AhmedBM: Trying to remember why we did this this way (#343), and i believe it due to AIS picking up changes if we just copied over the toml unnecessarily. Personally, i think it should just behind a regular install target as it makes things simpler eg.
```
install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/daemon/${target_name}.toml DESTINATION /etc/aziot/identityd/config.d/${target_name}.toml)
```

MariusNi: Jeremi, we are waiting for your reply and decision how to proceed. AIS is in past for us and we only have some backwards compatibility for IoT Hub with it, only on Ubuntu/Debian basically. Please review our comments and decide which way works best for you and for our old AIS story. Thanks!

jepio: I'm happy with the suggestion from @AhmedBM and I've pushed an updated commit that does that. Please take another look.

> Thanks for the explanation. So can you build with the second command as: 'cmake --build . --config Release --target all' (no install target), without sudo, and do not hit the break in your local environment?

Only with the changes I proposed in this PR (either the original, or the one from Ahmed), not with what is currently in main.

MariusNi (approved):
:shipit:

AhmedBM (approved):
👌

PR #795, opened 2024-11-08
Fuzzing targets: variant B

robertwoj-microsoft:
## Description

Variant B of fuzzing targets, some are skipped if not suitable.

## Checklist


MariusNi (approved):

robertschaedler3 (approved):

src/fuzzer/seed_corpus/crash-057fa54d958afa317498474299db202467ec95d1
AhmedBM, line 1: what are the purpose of these files?

jepio: These are initial inputs for the fuzzer, the fuzzer draws inspiration from these sample inputs and then mutates/randomizes the input to reach more and more code paths. When the fuzzer finds an input that triggers a crash or leak it saves back the file.

This implementation uses the convention that the files contain a prefix indicating which function to call and then input data follows. This makes it possible to fuzz multiple functions in one invocation without having to write and start multiple fuzzer processes.

jepio: The crash- and leak- files indicate an input that the fuzzer found. You can reproduce the same crash by providing that same input to the fuzzer.

PR #798, opened 2024-11-08
Various hardening for commonutils library internal calls for fuzzing

MariusNi:
## Description

Various hardening for commonutils library internal calls for fuzzing and follow-up to bugs found this way. Work in progress still. We may need more fixes than these, to be seen. For now let's review these.

## Checklist

### DIFF
--- a/src/common/commonutils/DaemonUtils.c
+++ b/src/common/commonutils/DaemonUtils.c
@@ -3,6 +3,32 @@
 
 #include "Internal.h"
 
+#define MAX_DAEMON_NAME_LENGTH 256
+
+bool IsValidDaemonName(const char *name) 
+{
+    size_t length = 0, i = 0;
+    bool result = true;
+    
+    if ((NULL != name) && (0 < (length = strnlen(name, MAX_DAEMON_NAME_LENGTH))))
+    {
+        for (i = 0; i < length; i++) 
+        {
+            if ((0 == isalnum(name[i])) && ('_' != name[i]) && ('-' != name[i]))
+            {
+                result = false;
+                break;
+            }
+        }
+    }
+    else
+    {
+        result = false;
+    }
+
+    return result;
+}
+
 static int ExecuteSystemctlCommand(const char* command, const char* daemonName, void* log)
 {
     const char* commandTemplate = "systemctl %s %s";
@@ -12,6 +38,11 @@ static int ExecuteSystemctlCommand(const char* command, const char* daemonName, 
     if ((NULL == command) || (NULL == daemonName))
     {
         OsConfigLogError(log, "ExecuteSystemctlCommand: invalid arguments");
+        return EINVAL;
+    }
+    else if (false == IsValidDaemonName(daemonName))
+    {
+        OsConfigLogError(log, "ExecuteSystemctlCommand: invalid daemon name '%s'", daemonName);
         return EINVAL;
     }
     else if (NULL == (formattedCommand = FormatAllocateString(commandTemplate, command, daemonName)))
@@ -27,7 +58,7 @@ static int ExecuteSystemctlCommand(const char* command, const char* daemonName, 
 
 bool IsDaemonActive(const char* daemonName, void* log)
 {
-    return (0 == ExecuteSystemctlCommand("is-active", daemonName, log)) ? true : false;
+    return (IsValidDaemonName(daemonName) && (0 == ExecuteSystemctlCommand("is-active", daemonName, log))) ? true : false;
 }
 
 bool CheckDaemonActive(const char* daemonName, char** reason, void* log)
@@ -73,6 +104,12 @@ static bool CommandDaemon(const char* command, const char* daemonName, void* log
     int result = 0;
     bool status = true;
 
+    if (false == IsValidDaemonName(daemonName))
+    {
+        OsConfigLogError(log, "CommandDaemon: invalid daemon name '%s'", daemonName);
+        return false;
+    }
+
     if (0 == (result = ExecuteSystemctlCommand(command, daemonName, log)))
     {
         OsConfigLogInfo(log, "Succeeded to %s service '%s'", command, daemonName);
@@ -99,6 +136,12 @@ bool StartDaemon(const char* daemonName, void* log)
 bool EnableAndStartDaemon(const char* daemonName, void* log)
 {
     bool status = true;
+
+    if (false == IsValidDaemonName(daemonName))
+    {
+        OsConfigLogError(log, "EnableAndStartDaemon: invalid daemon name '%s'", daemonName);
+        return false;
+    }
 
     if (false == IsDaemonActive(daemonName, log))
     {
--- a/src/common/commonutils/DeviceInfoUtils.c
+++ b/src/common/commonutils/DeviceInfoUtils.c
@@ -32,17 +32,17 @@ void RemovePrefixBlanks(char* target)
 
 void RemovePrefixUpTo(char* target, char marker)
 {
-    if (NULL == target)
+    size_t targetLength = 0;
+    char* equalSign = NULL;
+    
+    if ((NULL == target) || (0 == (length = strnlen(target, MAX_STRING_LENGTH))))
     {
         return;
     }
-
-    int targetLength =(int)strlen(target);
-    char* equalSign = strchr(target, marker);
-
-    if (equalSign)
+    
+    if (NULL != (equalSign = strchr(target, marker)))
     {
-        targetLength = strlen(equalSign + 1);
+        targetLength = strnlen(equalSign + 1, MAX_STRING_LENGTH);
         memmove(target, equalSign + 1, targetLength);
         target[targetLength] = 0;
     }
@@ -50,17 +50,17 @@ void RemovePrefixUpTo(char* target, char marker)
 
 void RemovePrefixUpToString(char* target, const char* marker)
 {
+    size_t targetLength = 0;
+    char* equalSign = NULL;
+    
     if ((NULL == target) || (NULL == marker))
     {
         return;
     }
 
-    int targetLength = 0;
-    char* equalSign = strstr(target, marker);
-
-    if (equalSign)
+    if ((NULL != (equalSign = strstr(target, marker))) && 
+        (0 < (targetLength = strnlen(equalSign, MAX_STRING_LENGTH))))
     {
-        targetLength = (int)strlen(equalSign);
         memmove(target, equalSign, targetLength);
         target[targetLength] = 0;
     }
@@ -68,13 +68,14 @@ void RemovePrefixUpToString(char* target, const char* marker)
 
 void RemoveTrailingBlanks(char* target)
 {
+    int i = 0;
+
     if (NULL == target)
     {
         return;
     }
 
-    int targetLength = (int)strlen(target);
-    int i = targetLength;
+    i = (int)strnlen(target, MAX_STRING_LENGTH);
 
     while ((i > 0) && (' ' == target[i - 1]))
     {
--- a/src/common/commonutils/DeviceInfoUtils.c
+++ b/src/common/commonutils/DeviceInfoUtils.c
@@ -35,14 +35,14 @@ void RemovePrefixUpTo(char* target, char marker)
     size_t targetLength = 0;
     char* equalSign = NULL;
     
-    if ((NULL == target) || (0 == (length = strnlen(target, MAX_STRING_LENGTH))))
+    if ((NULL == target) || (0 == (targetLength = strnlen(target, MAX_STRING_LENGTH))))
     {
         return;
     }
     
-    if (NULL != (equalSign = strchr(target, marker)))
+    if ((NULL != (equalSign = strchr(target, marker))) && 
+        (0 < (targetLength = strnlen(equalSign + 1, MAX_STRING_LENGTH))))
     {
-        targetLength = strnlen(equalSign + 1, MAX_STRING_LENGTH);
         memmove(target, equalSign + 1, targetLength);
         target[targetLength] = 0;
     }
--- a/src/common/commonutils/DeviceInfoUtils.c
+++ b/src/common/commonutils/DeviceInfoUtils.c
@@ -53,13 +53,13 @@ void RemovePrefixUpToString(char* target, const char* marker)
     size_t targetLength = 0;
     char* equalSign = NULL;
     
-    if ((NULL == target) || (NULL == marker))
+    if ((NULL == target) || (0 >= (targetLength = strnlen(target, MAX_STRING_LENGTH))) || 
+        (NULL == marker) || (0 >= strnlen(marker, MAX_STRING_LENGTH)))
     {
         return;
     }
 
-    if ((NULL != (equalSign = strstr(target, marker))) && 
-        (0 < (targetLength = strnlen(equalSign, MAX_STRING_LENGTH))))
+    if (NULL != (equalSign = strstr(target, marker)))
     {
         memmove(target, equalSign, targetLength);
         target[targetLength] = 0;
--- a/src/common/commonutils/FileUtils.c
+++ b/src/common/commonutils/FileUtils.c
@@ -1297,7 +1297,7 @@ int CheckTextNotFoundInEnvironmentVariable(const char* variableName, const char*
     bool foundText = false;
     int status = 0;
 
-    if ((NULL == variableName) || (NULL == text) || (0 == strlen(variableName)) || (0 == strlen(text)))
+    if ((NULL == variableName) || (NULL == text) || (0 == strlen(variableName)) || (0 == strlen(text) || (false == IsValidDaemonName(variableName))))
     {
         OsConfigLogError(log, "CheckTextNotFoundInEnvironmentVariable called with invalid arguments");
         return EINVAL;
--- a/src/common/commonutils/FileUtils.c
+++ b/src/common/commonutils/FileUtils.c
@@ -1198,6 +1198,11 @@ int CheckTextIsNotFoundInFile(const char* fileName, const char* text, char** rea
     return result;
 }
 
+static bool IsValidFileNameCharacter(char text)
+{
+    return ((0 == isalnum(text)) && ('_' != text) && ('-' != text) && ('.' != text) && ('~' != text) && ('/' != text)) ? false : true;
+}
+
 static bool IsValidGrepCharacter(char text)
 {
     return ((0 == isalnum(text)) && ('_' != text) && ('-' != text) && ('.' != text) && ('~' != text) && ('#' != text) && ('!' != text)) ? false : true;
@@ -1224,6 +1229,27 @@ static bool IsValidGrepArgument(const char* text)
     return result;
 }
 
+static bool IsValidFileName(const char* text)
+{
+    size_t length = 0, i = 0;
+    bool result = true;
+
+    if ((NULL == text) || (0 >= (length = strnlen(text, MAX_STRING_LENGTH))))
+    {
+        return false;
+    }
+
+    for (i = 0; i < length; i++)
+    {
+        if (false == (result = IsValidFileNameCharacter(text[i])))
+        {
+            break;
+        }
+    }
+
+    return result;
+}
+
 int CheckMarkedTextNotFoundInFile(const char* fileName, const char* text, const char* marker, char commentCharacter, char** reason, void* log)
 {
     const char* commandTemplate = "grep -v '^%c' %s | grep %s";
@@ -1235,7 +1261,7 @@ int CheckMarkedTextNotFoundInFile(const char* fileName, const char* text, const 
     int status = 0;
 
     if ((!FileExists(fileName)) || (NULL == text) || (NULL == marker) || (0 == strlen(text)) || (0 == strlen(marker)) || 
-        (false == IsValidGrepArgument(text)) || (false == IsValidGrepArgument(fileName)) || 
+        (false == IsValidGrepArgument(text)) || (false == IsValidFileNam(fileName)) || 
         (false == IsValidGrepArgument(marker)) || (false == IsValidGrepCharacter(commentCharacter)))
     {
         OsConfigLogError(log, "CheckMarkedTextNotFoundInFile called with invalid arguments");
--- a/src/common/commonutils/FileUtils.c
+++ b/src/common/commonutils/FileUtils.c
@@ -1198,6 +1198,32 @@ int CheckTextIsNotFoundInFile(const char* fileName, const char* text, char** rea
     return result;
 }
 
+static bool IsValidGrepCharacter(char text)
+{
+    return ((0 == isalnum(text)) && ('_' != text) && ('-' != text) && ('.' != text)) ? false : true;
+}
+
+static bool IsValidGrepArgument(const char* text)
+{
+    size_t length = 0, i = 0;
+    bool result = true;
+
+    if ((NULL == test) || (0 >= (length = strnlen(text, MAX_STRING_LENGTH))))
+    {
+        return false;
+    }
+    
+    for (i = 0; i < length; i++)
+    {
+        if (false == (result = IsValidGrepCharacter(text[i])))
+        {
+            break;
+        }
+    }
+    
+    return result;
+}
+
 int CheckMarkedTextNotFoundInFile(const char* fileName, const char* text, const char* marker, char commentCharacter, char** reason, void* log)
 {
     const char* commandTemplate = "grep -v '^%c' %s | grep %s";
@@ -1208,7 +1234,8 @@ int CheckMarkedTextNotFoundInFile(const char* fileName, const char* text, const 
     bool foundMarker = false;
     int status = 0;
 
-    if ((!FileExists(fileName)) || (NULL == text) || (NULL == marker) || (0 == strlen(text)) || (0 == strlen(marker)))
+    if ((!FileExists(fileName)) || (NULL == text) || (NULL == marker) || (0 == strlen(text)) || (0 == strlen(marker)) || 
+        (false == IsValidGrepArgument(text)) || (false == IsValidGrepArgument(fileName)))
     {
         OsConfigLogError(log, "CheckMarkedTextNotFoundInFile called with invalid arguments");
         return EINVAL;
--- a/src/common/commonutils/FileUtils.c
+++ b/src/common/commonutils/FileUtils.c
@@ -1261,7 +1261,7 @@ int CheckMarkedTextNotFoundInFile(const char* fileName, const char* text, const 
     int status = 0;
 
     if ((!FileExists(fileName)) || (NULL == text) || (NULL == marker) || (0 == strlen(text)) || (0 == strlen(marker)) || 
-        (false == IsValidGrepArgument(text)) || (false == IsValidFileNam(fileName)) || 
+        (false == IsValidGrepArgument(text)) || (false == IsValidFileName(fileName)) || 
         (false == IsValidGrepArgument(marker)) || (false == IsValidGrepCharacter(commentCharacter)))
     {
         OsConfigLogError(log, "CheckMarkedTextNotFoundInFile called with invalid arguments");
--- a/src/common/commonutils/FileUtils.c
+++ b/src/common/commonutils/FileUtils.c
@@ -1235,7 +1235,7 @@ int CheckMarkedTextNotFoundInFile(const char* fileName, const char* text, const 
     int status = 0;
 
     if ((!FileExists(fileName)) || (NULL == text) || (NULL == marker) || (0 == strlen(text)) || (0 == strlen(marker)) || 
-        (false == IsValidGrepArgument(text)) || (false == IsValidGrepCharacter(commentCharacter)))
+        (false == IsValidGrepArgument(text)) || ('#' != commentCharacter)))
     {
         OsConfigLogError(log, "CheckMarkedTextNotFoundInFile called with invalid arguments");
         return EINVAL;
--- a/src/common/commonutils/FileUtils.c
+++ b/src/common/commonutils/FileUtils.c
@@ -1198,14 +1198,9 @@ int CheckTextIsNotFoundInFile(const char* fileName, const char* text, char** rea
     return result;
 }
 
-static bool IsValidFileNameCharacter(char text)
-{
-    return ((0 == isalnum(text)) && ('_' != text) && ('-' != text) && ('.' != text) && ('~' != text) && ('/' != text)) ? false : true;
-}
-
 static bool IsValidGrepCharacter(char text)
 {
-    return ((0 == isalnum(text)) && ('_' != text) && ('-' != text) && ('.' != text) && ('~' != text) && ('#' != text) && ('!' != text)) ? false : true;
+    return ((0 == isalnum(text)) && ('_' != text) && ('-' != text) && ('.' != text)) ? false : true;
 }
 
 static bool IsValidGrepArgument(const char* text)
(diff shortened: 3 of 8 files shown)


src/common/commonutils/DaemonUtils.c
```diff
@@ -3,6 +3,32 @@
 
 #include "Internal.h"
 
+#define MAX_DAEMON_NAME_LENGTH 256
+
+bool IsValidDaemonName(const char *name)
```
AhmedBM, line 8: What IS a valid daemon name? Just by looking at the code its no numbers, underscores and dashes? Could you include that as function comments?

src/common/commonutils/DaemonUtils.c
```diff
@@ -14,6 +40,11 @@ static int ExecuteSystemctlCommand(const char* command, const char* daemonName,
         OsConfigLogError(log, "ExecuteSystemctlCommand: invalid arguments");
         return EINVAL;
     }
+    else if (false == IsValidDaemonName(daemonName))
+    {
+        OsConfigLogError(log, "ExecuteSystemctlCommand: invalid daemon name '%s'", daemonName);
```
AhmedBM, line 45: same comment as above, might be useful to print that out here as well?

src/common/commonutils/FileUtils.c
```diff
@@ -1199,6 +1198,32 @@ int CheckTextIsNotFoundInFile(const char* fileName, const char* text, char** rea
     return result;
 }
 
+static bool IsValidGrepCharacter(char text)
+{
+    return ((0 == isalnum(text)) && ('_' != text) && ('-' != text) && ('.' != text)) ? false : true;
```
AhmedBM, line 1203: Function comments as to what is a valid grep character would be helpful.

src/common/commonutils/UrlUtils.c
```diff
@@ -41,26 +71,23 @@ char* UrlEncode(char* target)
...
+    if ((NULL == target) || (0 >= (targetLength = strnlen(target, MAX_STRING_LENGTH))))
     {
         return NULL;
     }
 
-    targetLength = (int)strlen(target);
-    
     // The length of the decoded string is the same as the length of the encoded string or smaller
-    decodedTarget = (char*)malloc(targetLength + 1);
+    decodedTarget = (char*)malloc(targetLength + 3);
```
AhmedBM, line 87: why 2 more bytes?

src/common/commonutils/DaemonUtils.c
MariusNi: Correct. Those names cannot be anything. Only these characters.

src/common/commonutils/UrlUtils.c
MariusNi: To allow room for movement towards the end of the buffer. See loop on line 92.

src/common/commonutils/FileUtils.c
MariusNi: That's not a general grep thing, Is just for the grep command in CheckMarkedTextNotFoundInFile. Why a static here. Line 1203 says all that's needed.

src/common/commonutils/DaemonUtils.c
MariusNi: Duplicate, closing.

AhmedBM (approved):
👌

robertschaedler3 (approved):

jepio (commented):
Quick review, not meant to replace a thorough one by Robert.

Can you split this into separate PRs for the different kinds of changes: file handling, string handling, url handling, daemon name.

Can you write down in the PR description which bugs/fuzzing cases/scenarios these are meant to fix?

src/common/commonutils/DeviceInfoUtils.c
```diff
@@ -32,49 +32,52 @@ void RemovePrefixBlanks(char* target)
...
 }
 
 void RemovePrefixUpToString(char* target, const char* marker)
 {
-    if ((NULL == target) || (NULL == marker))
+    size_t targetLength = 0, markerLength = 0;
+    char* equalSign = NULL;
+    
+    if ((NULL == target) || (0 >= (targetLength = strnlen(target, MAX_STRING_LENGTH))) || 
+        (NULL == marker) || (0 >= (markerLength = strnlen(marker, MAX_STRING_LENGTH))) ||
```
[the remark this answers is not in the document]
MariusNi: These functions  are not meant to return anything else except the modified string itself. In this case string remaining unmodified. This being not an error case, but a no-op case (target or markers are too long). At the same time I agree we need to catch and report cases where the strings are too long. Considering.

src/common/commonutils/DeviceInfoUtils.c
```diff
@@ -32,49 +32,52 @@ void RemovePrefixBlanks(char* target)
...
         return;
     }
-
-    int targetLength =(int)strlen(target);
-    char* equalSign = strchr(target, marker);
-
-    if (equalSign)
+    
+    if ((NULL != (equalSign = strchr(target, marker))) && 
+        (0 < (targetLength = strnlen(equalSign + 1, MAX_STRING_LENGTH))))
```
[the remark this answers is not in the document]
MariusNi: RemovePrefixUpTo("option=", '=') should return "" (empty string). I'll add more test cases and investigate this.

MariusNi:
Keeping all these hardening fixes together as this PR will be a long duration one, to allow Robert to CR, and also to allow me see the fuzzer PRs ingested into our main, to allow me complete these fixes with local fuzzing working in this branch. I will expand on the description to enumerate all fixes one by one as the PR will advance further into completion. For now, this PR is just start of the work, to capture early feedback. Thanks for your feedback, it's very useful to me!

---
In reply to: [2427614917](https://github.com/Azure/azure-osconfig/pull/798#issuecomment-2427614917) [](http://example.com/codeflow?ancestors=2427614917)

src/common/commonutils/FileUtils.c
```diff
@@ -1209,7 +1213,8 @@ int CheckMarkedTextNotFoundInFile(const char* fileName, const char* text, const
     bool foundMarker = false;
     int status = 0;
 
-    if ((!FileExists(fileName)) || (NULL == text) || (NULL == marker) || (0 == strlen(text)) || (0 == strlen(marker)))
+    if ((false == FileExists(fileName)) || (NULL == text) || (NULL == marker) || (0 == strlen(text)) || (0 == strlen(marker)) || 
+        (false == IsValidGrepArgument(text)) || ('#' != commentCharacter))
```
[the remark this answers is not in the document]
MariusNi: Just so happens all system files the ASB touches use '#' as comment character. In the future however more comment characters could be added. 

Can you please help me here and say, from your experience, what other comment characters may be common? To add those as possible options and harden for them.

src/common/commonutils/DeviceInfoUtils.c
```diff
@@ -32,49 +32,52 @@ void RemovePrefixBlanks(char* target)
...
+    int i = 0;
+
     if (NULL == target)
     {
         return;
     }
 
-    int targetLength = (int)strlen(target);
-    int i = targetLength;
+    i = (int)strnlen(target, MAX_STRING_LENGTH);
```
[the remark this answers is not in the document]
MariusNi: Thanks for the feedback! I agree that it would be useful to catch and report strings that are too long. May simply remove this restriction here and allow strings for any length, same as it was before. TBD.

src/common/commonutils/DeviceInfoUtils.c
```diff
@@ -32,49 +32,52 @@ void RemovePrefixBlanks(char* target)
 
 void RemovePrefixUpTo(char* target, char marker)
```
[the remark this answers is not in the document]
MariusNi: Great idea. The problem is, they are used like this already in many places, one skipping over marker and the other one returning it. The skip over marker is the likely culprit for the fuzzer induced segfault now I suspect. Investigating to make then behave the same and unifying as you suggested. Thanks

src/common/commonutils/DeviceInfoUtils.c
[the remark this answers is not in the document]
MariusNi: Reverted to allow strings of any length to be handled.

src/common/commonutils/DeviceInfoUtils.c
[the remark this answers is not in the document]
MariusNi: Reverted by going back to allow strings of any length to be processed.

src/common/commonutils/FileUtils.c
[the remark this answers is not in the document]
MariusNi: Adding for now for just alpha-numerical chars plus '#'.

src/common/commonutils/OtherUtils.c
```diff
@@ -5,17 +5,15 @@
 
 char* DuplicateString(const char* source)
```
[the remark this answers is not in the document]
MariusNi: Agree. Please see last iteration

src/common/commonutils/DeviceInfoUtils.c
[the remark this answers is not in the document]
MariusNi: Solved by making both functions return including the marker and completed the existing uses of the one skipping the marker with a new utility function that skips prefix instances of a character, to skip over the marker. Will add more unit-tests but the changes are already in, pls. check them out.

src/common/commonutils/OtherUtils.c
```diff
@@ -41,13 +30,12 @@ char* DuplicateStringToLowercase(const char* source)
...
 }
 
-#define DEFAULT_FORMAT_ALLOCATE_STRING_LENGTH 512
 char* FormatAllocateString(const char* format, ...)
 {
     char* buffer = NULL;
     char* stringToReturn = NULL;
     int formatResult = 0;
-    int sizeOfBuffer = DEFAULT_FORMAT_ALLOCATE_STRING_LENGTH;
+    int sizeOfBuffer = MAX_STRING_LENGTH;
```
simonjaeger, line 38: Is there any chance this is used for banners, etc. that would require a larger buffer than specified? I assume not.

src/common/commonutils/OtherUtils.c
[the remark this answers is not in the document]
robertschaedler3: Can the wrapper be removed (assuming `NULL` is checked) everywhere else?

simonjaeger (approved):

robertschaedler3 (approved):

src/common/commonutils/OtherUtils.c
AhmedBM: good point SSH login screens tend to be quite large

AhmedBM (approved):
👌

src/common/commonutils/UrlUtils.c
```diff
@@ -3,24 +3,54 @@
 
 #include "Internal.h"
 
-char* UrlEncode(char* target)
+#define MAX_URL_LENGTH 2048
+
+static bool IsValidUrlCharacter(char c)
 {
-    if (NULL == target)
+    return (isalnum(c) || ('-' == c) || ('_' == c) || ('.' == c) || ('~' == c) || ('%' == c)) ? true : false;
```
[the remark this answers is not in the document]
MariusNi: These are used in the PnP Agent code, not in ASB.  It's the product info capturing data for our own IoT Hub telemetry, for example. They are in target for fuzzing as well as the rest.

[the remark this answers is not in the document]
MariusNi: These are the only characters we need for now.

src/common/commonutils/OtherUtils.c
MariusNi: If you mean the SSH banner, that one is s static string, coming as an Azure Policy parameter. Not dynamic. We may increase this size in the future if/when we will need to format larger strings.

src/common/commonutils/OtherUtils.c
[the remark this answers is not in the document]
MariusNi: Don't thing strdup plays well called for null. Also, do not want to increase churn code by replacing the wrapper thoroughly. Thanks for the feedback though!

src/common/commonutils/DeviceInfoUtils.c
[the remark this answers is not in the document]
robertwoj-microsoft: How about passing additional parameter, for example includePrefix or skipPrefix and take that into account? My problem with additional removePrefix calls is that we call memmove one more time and we know we'll need the string past the prefix. Just a suggestion tho, I'm fine with what is now too.

src/common/commonutils/DeviceInfoUtils.c
```diff
@@ -141,6 +136,7 @@ char* GetOsName(void* log)
             RemovePrefixBlanks(textResult);
             RemoveTrailingBlanks(textResult);
             RemovePrefixUpTo(textResult, '=');
+            RemovePrefix(textResult, '=');
```
robertwoj-microsoft, line 139: See my comment under unifying RemovePrefixUpTo with RemovePrefixUpToString (this is resolved now).

src/common/commonutils/FileUtils.c
```diff
@@ -170,7 +169,7 @@ bool AppendPayloadToFile(const char* fileName, const char* payload, const int pa
         return result;
     }
 
-    // If the file exists and there is no EOL at the end of file, add one before the append
+    // If the file exists and there is no EOL at the end of file, try to add one before the append
     if (FileExists(fileName) && (false == FileEndsInEol(fileName, log)))
     {
         if (false == SaveToFile(fileName, "a", "\n", 1, log))
```
robertwoj-microsoft, line 175: While at it, please consider failing the function in case we fail here.

src/common/commonutils/FileUtils.c
```diff
@@ -1199,6 +1198,16 @@ int CheckTextIsNotFoundInFile(const char* fileName, const char* text, char** rea
...
 }
 
+static bool IsValidGrepArgument(const char* text)
+{
+    return IsValidDaemonName(text);
+}
+
+static bool IsValidCommentCharacter(char c)
+{
+    return ((0 == isalnum(c)) && ('#' != c)) ? false : true;
```
robertwoj-microsoft, line 1208: This is unclear to me. Do we use alnum characters as comments? If we do, can we maybe add a comment similarly to what we have for IsValidDaemonName?

robertwoj-microsoft, line 1208: Could we do sth like this maybe?
`return isalnum(c) || (c == '#');`
This looks more clear to me.

src/common/commonutils/FileUtils.c
```diff
@@ -1199,6 +1198,16 @@ int CheckTextIsNotFoundInFile(const char* fileName, const char* text, char** rea
...
+}
+
 int CheckMarkedTextNotFoundInFile(const char* fileName, const char* text, const char* marker, char commentCharacter, char** reason, void* log)
 {
     const char* commandTemplate = "grep -v '^%c' %s | grep %s";
```
robertwoj-microsoft, line 1213: Do we need to call grep twice here? The first one just tests the first character in line, the latter is seeking for a text which is not a regex, so this could be achieved by reading a file line by line and calling strstr() on these if I'm correct. We could save the need for 'text' and 'filename' arguments validation, forking a process and relying on timeouts. If you believe this optimisation is worth considering, please do in a separate PR so it's easier to review.

robertwoj-microsoft, line 1213: How about files that start with spaces and then a comment character? Is this a possibility?

src/common/commonutils/FileUtils.c
```diff
@@ -1343,31 +1353,36 @@ int CheckTextNotFoundInEnvironmentVariable(const char* variableName, const char*
...
+    else if ((0 == stat(fileName, &statStruct)) && ((statStruct.st_size > MAX_STRING_LENGTH)))
+    {
+        OsConfigLogError(log, "CheckSmallFileContainsText: file is too large (%lu bytes, maximum supported: %d bytes)", statStruct.st_size, MAX_STRING_LENGTH);
         return EINVAL;
     }
 
     if (NULL != (contents = LoadStringFromFile(fileName, false, log)))
     {
-        textLength = strlen(text);
         contentsLength = strlen(text);
```
robertwoj-microsoft, line 1376: Why not `strlen(contents)`?

src/common/commonutils/FileUtils.c
```diff
@@ -1608,15 +1623,16 @@ static int FindTextInCommandOutput(const char* command, const char* text, void*
...
 
-    if ((NULL == command) || (NULL == text))
+    if ((NULL == command) || (NULL == text) || (0 == strlen(command)) || (0 == strlen(text)))
     {
         OsConfigLogError(log, "FindTextInCommandOutput called with invalid argument");
         return EINVAL;
     }
 
-    if (0 == (status = ExecuteCommand(NULL, command, true, false, 0, 0, &results, NULL, log)))
+    // Execute this command with a 60 seconds timeout
```
robertwoj-microsoft, line 1632: Please verify the SystemCommand function. I'm OK with explicitly specifying 60s here. But my problem is that in the mentioned function we have:
`int timeout = (timeoutSeconds > 0) ? timeoutSeconds : defaultCommandTimeout;`
which suggests when 0 is passed, the default timeout is used, but later on the function only validates the `timeoutSeconds` variable, which means no timeout is used which is misleading to me. If that's intended, please consider adding some comments there.

src/common/commonutils/FileUtils.c
```diff
@@ -1608,15 +1623,16 @@ static int FindTextInCommandOutput(const char* command, const char* text, void*
...
+    if (0 == (status = ExecuteCommand(NULL, command, true, false, 0, 60, &results, NULL, log)))
     {
-        if (NULL != strstr(results, text))
+        if ((NULL != results) && (0 < strlen(results)) && (NULL != strstr(results, text)))
```
robertwoj-microsoft, line 1635: Is the strlen() call needed here? Looks like strstr() should be enough.

src/common/commonutils/UrlUtils.c
```diff
@@ -3,24 +3,26 @@
 
 #include "Internal.h"
 
-char* UrlEncode(char* target)
+char* UrlEncode(const char* target, size_t targetSize)
```
robertwoj-microsoft, line 6: Is there a possibility that we pass a non-terminated string here? I'm not sure why we add the size here. If that's really need then OK.

src/common/commonutils/UrlUtils.c
```diff
@@ -3,24 +3,26 @@
...
     {
         return NULL;
     }
 
-    int i = 0, j = 0;
-    int targetLength = (int)strlen(target);
-    int encodedLength = (3 * targetLength) + 1;
-    char* encodedTarget = (char*)malloc(encodedLength);
-    if (NULL != encodedTarget)
+    encodedLength = (3 * targetSize) + 3;
```
robertwoj-microsoft, line 17: Why +3?

src/common/commonutils/UrlUtils.c
```diff
@@ -32,7 +34,7 @@ char* UrlEncode(char* target)
             }
             else
             {
-                sprintf(&encodedTarget[j], "%%%02X", target[i]);
+                snprintf(&encodedTarget[j], encodedLength - j, "%%%02X", target[i]);
                 j += strlen(&encodedTarget[j]);
```
robertwoj-microsoft, line 38: Will it be different than 3 if we allocated proper size for the buffer?

src/common/commonutils/UrlUtils.c
```diff
@@ -32,7 +34,7 @@ char* UrlEncode(char* target)
             }
             else
             {
-                sprintf(&encodedTarget[j], "%%%02X", target[i]);
+                snprintf(&encodedTarget[j], encodedLength - j, "%%%02X", target[i]);
```
robertwoj-microsoft, line 37: If we need the length parameter here, it means we've allocated a wrong buffer size. This in fact may be worse than crashing, because we'll silently return an ill-formed url.

src/common/commonutils/UrlUtils.c
robertwoj-microsoft, line 6: Do we need special handling for a newline character?

src/common/commonutils/UrlUtils.c
```diff
@@ -41,28 +43,23 @@ char* UrlEncode(char* target)
...
     {
         return NULL;
     }
 
-    targetLength = (int)strlen(target);
-    
-    // The length of the decoded string is the same as the length of the encoded string or smaller
-    decodedTarget = (char*)malloc(targetLength + 1);
-    if (NULL != decodedTarget)
+    if (NULL != (decodedTarget = (char*)malloc(targetSize + 3)))
```
robertwoj-microsoft, line 58: Why +3?

src/common/commonutils/UrlUtils.c
```diff
@@ -71,7 +68,7 @@ char* UrlDecode(char* target)
             }
             else if ('%' == target[j])
             {
-                if (((j + 2) < targetLength) && ('0' == target[j + 1]) && ('A' == toupper(target[j + 2])))
+                if (((j + 2) < targetSize) && ('0' == target[j + 1]) && ('A' == toupper(target[j + 2])))
```
robertwoj-microsoft, line 71: Do we need special handling for the newline character?

src/common/commonutils/UrlUtils.c
```diff
@@ -81,7 +78,7 @@ char* UrlDecode(char* target)
                     buffer[2] = 0;
                     
                     sscanf(buffer, "%x", &value);
-                    sprintf(&decodedTarget[i], "%c", value);
+                    snprintf(&decodedTarget[i], targetSize - i, "%c", value);
```
robertwoj-microsoft, line 81: Similar comment to the same in UrlEncode. This looks suspicious.

robertwoj-microsoft (changes requested):
For next bugfixes, please use separate PRs if possible, this helps reviewing and releases dependencies between changes. You can also pin the bug numbers in the description for easier tracking.

src/common/commonutils/DaemonUtils.c
```diff
@@ -3,6 +3,37 @@
...
+{
+    return ((0 == isalnum(c)) && ('_' != c) && ('-' != c) && ('.' != c)) ? false : true;
+}
+
+bool IsValidDaemonName(const char *name) 
+{
+    size_t length = 0, i = 0;
+    bool result = true;
+    
+    if ((NULL != name) && (0 < (length = strnlen(name, MAX_DAEMON_NAME_LENGTH))))
```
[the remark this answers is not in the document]
MariusNi: Agree, good idea

src/common/commonutils/DaemonUtils.c
```diff
@@ -14,6 +45,11 @@ static int ExecuteSystemctlCommand(const char* command, const char* daemonName,
         OsConfigLogError(log, "ExecuteSystemctlCommand: invalid arguments");
         return EINVAL;
     }
+    else if (false == IsValidDaemonName(daemonName))
+    {
+        OsConfigLogError(log, "ExecuteSystemctlCommand: invalid daemon name '%s'", daemonName);
+        return EINVAL;
+    }
```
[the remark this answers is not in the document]
MariusNi: Static function, only functions from this file can call it, the commands are limited by the public functions that call it:

bool StartDaemon(const char* daemonName, void* log);
bool StopDaemon(const char* daemonName, void* log);
bool EnableDaemon(const char* daemonName, void* log);
bool DisableDaemon(const char* daemonName, void* log);
bool EnableAndStartDaemon(const char* daemonName, void* log);
void StopAndDisableDaemon(const char* daemonName, void* log);
bool RestartDaemon(const char* daemonName, void* log);
bool MaskDaemon(const char* daemonName, void* log);

etc.

No external callers can call into this.

src/common/commonutils/DeviceInfoUtils.c
```diff
@@ -30,51 +29,46 @@ void RemovePrefixBlanks(char* target)
...
+    RemovePrefixUpToString(target, markerString);
 }
 
 void RemovePrefixUpToString(char* target, const char* marker)
 {
-    if ((NULL == target) || (NULL == marker))
+    size_t targetLength = 0, markerLength = 0;
+    char* equalSign = NULL;
+    
+    if ((NULL == target) || (0 == (targetLength = strlen(target))) ||
```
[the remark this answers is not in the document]
MariusNi: Correct, if the needle is longer than the haystack, it cannot be fully contained within the haystack, and thus strstr() will not find a match and will return NULL. We can remove comparison on line 50. We have to check for non-empty strings though.

src/common/commonutils/DeviceInfoUtils.c
```diff
@@ -30,51 +29,46 @@ void RemovePrefixBlanks(char* target)
...
+{
+    char markerString[2] = {marker, 0};
+    RemovePrefixUpToString(target, markerString);
 }
 
 void RemovePrefixUpToString(char* target, const char* marker)
 {
-    if ((NULL == target) || (NULL == marker))
+    size_t targetLength = 0, markerLength = 0;
+    char* equalSign = NULL;
```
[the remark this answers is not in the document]
MariusNi: Renaming to 'found'

src/common/commonutils/DaemonUtils.c
```diff
@@ -73,6 +109,12 @@ static bool CommandDaemon(const char* command, const char* daemonName, void* log
     int result = 0;
     bool status = true;
 
+    if (false == IsValidDaemonName(daemonName))
+    {
+        OsConfigLogError(log, "CommandDaemon: invalid daemon name '%s'", daemonName);
+        return false;
+    }
+
```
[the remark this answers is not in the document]
MariusNi: Same answer.

src/common/commonutils/DeviceInfoUtils.c
```diff
@@ -30,51 +29,46 @@ void RemovePrefixBlanks(char* target)
...
+        (markerLength >= targetLength))
     {
         return;
     }
 
-    int targetLength = 0;
-    char* equalSign = strstr(target, marker);
-
-    if (equalSign)
+    if ((NULL != (equalSign = strstr(target, marker))) && (0 < (targetLength = strlen(equalSign))))
```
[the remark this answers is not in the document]
MariusNi: That's the purpose of the function, to remove that prefix, so no, we won't memmove target to itself, just bring forward the suffix.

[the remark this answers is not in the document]
MariusNi: strstr only returns non NULL if the substring is fully contained in target. So we do not need this here, will remove.

src/common/commonutils/DeviceInfoUtils.c
MariusNi: This is needed here now, once the RemovePrefixUpTo was changed to return the token.

The alternative would have been to add another function, to RemovePrefixAndSkipOverToken or similar, just prefer this variant as it's more explicit of what happens.

src/common/commonutils/FileUtils.c
MariusNi: Will adjust this

MariusNi: // Some of the common comment characters that can be encountered, add more if necessary
static bool IsValidCommentCharacter(char c)
{
    return (('#' != c) || ('/' != c) || ('*' != c) || (';' != c) || ('!' != c)) ? false : true;
}

MariusNi: Yes!

MariusNi: // Some of the common comment characters that can be encountered, add more if necessary
static bool IsValidCommentCharacter(char c)
{
    return (('#' == c) || ('/' == c) || ('*' == c) || (';' == c) || ('!' == c)) ? true : false;
}

src/common/commonutils/FileUtils.c
MariusNi: Not for us. If it were, we'd have to consider space a comment character and then comment sequences of multiple characters, etc.

MariusNi: Not that I am aware of.

MariusNi: This sounds like an interesting idea (thanks!) however it's nothing that we should block or delay release for now if you follow what I am saying. Thx

src/common/commonutils/FileUtils.c
MariusNi: Good catch!

src/common/commonutils/FileUtils.c
MariusNi: That's a different purpose variable. The actual timeout is in timeoutSeconds. There is no default 60 secs timeout.

src/common/commonutils/FileUtils.c
MariusNi: Don't want to call strstr on empty string as it may falsely return.

src/common/commonutils/UrlUtils.c
MariusNi: The answer in actual use is no. Could fuzzing call this with non null terminated strings?

If you look where these functions are used, only Encode is used in a single place, on a statically allocated and null terminated buffer buffer. Decode is just used in the unit test to validate the Encode one. These are very limited functionality functions which are not used in ASB.

MariusNi: I agree the size looks bad, would only be truly useful if we had SAL annotations here but we don't.

src/common/commonutils/UrlUtils.c
MariusNi: (targetSize + 1) * 3? Don't remember why added, won't remove that now and risk breaking the old IoT Hub telemetry this is used for.

src/common/commonutils/UrlUtils.c
MariusNi: Agree. Back to sprintf we go.

src/common/commonutils/UrlUtils.c
MariusNi: To allow room for movement towards the end of the buffer. See loop on line 92. 

But in the end, see the other comment. These are not used for ASB. Only Encode is used and that's for a a very precise and legacy telemetry for IoT Hub. Any change there can regress that.

src/common/commonutils/UrlUtils.c
MariusNi: See line 73

src/common/commonutils/UrlUtils.c
MariusNi: See line 30

src/common/commonutils/UrlUtils.c
MariusNi: As explained, not changing any of this now, risk of regression is too high.

src/common/commonutils/UrlUtils.c
robertwoj-microsoft: I mean, the latter else statement decodes hex character in lines 77-81. Isn't it yielding the same result?

src/common/commonutils/DeviceInfoUtils.c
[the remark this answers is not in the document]
MariusNi: Removed and now putting it back because of potential segfault found with sanitizer without.

src/common/commonutils/UrlUtils.c
MariusNi: Decided to refactor. Please see last iteration.

src/common/commonutils/UrlUtils.c
MariusNi: Please see new version in last iteration.

simonjaeger (approved):

robertschaedler3 (approved):

src/common/tests/CommonUtilsUT.cpp
```diff
@@ -205,6 +205,28 @@ TEST_F(CommonUtilsTest, FileEndsInEol)
...
+
+    EXPECT_TRUE(CreateTestFile(m_path, noEol));
+    EXPECT_EQ(false, FileEndsInEol(m_path, nullptr));
+    EXPECT_TRUE(Cleanup(m_path));
+}
+
+TEST_F(CommonUtilsTest, ZeroLengthFileEndsInEol)
+{
+    int fd = -1;
+    EXPECT_NE(-1, fd = open(m_path, O_CREAT | O_WRONLY, 0644));
```
robertwoj-microsoft, line 225: I think we're missing close here

robertwoj-microsoft (approved):

robertschaedler3 (approved):

PR #800, opened 2024-11-12
Minor fuzzer bugfixes

robertwoj-microsoft:
## Description

Add missing free calls in the fuzzer implementation. 

## Checklist


MariusNi (approved):
:shipit:

jepio (approved):

jepio:
@robertwoj-microsoft fuzzer has moved to `tests/`, could you rebase the PR?

MariusNi:
>             throw std::runtime_error(std::string{ "failed to create temporary directory: " } + std::strerror(errno));

Usually it's not recommended to throw from inside a constructor. Can the constructor be simplified to not crash and move the initialization to a new Initialize() method, for example?

---
Refers to: src/tests/fuzzer/target.cpp:42 in 7da148e. [](commit_id = 7da148e7425c20788a1e2b996a1238b7571e8528, deletion_comment = False)

MariusNi:
> public:

By default all struct members are public. Did you intend 'class' on line 32? If not, line 36 is not necessary.

---
Refers to: src/tests/fuzzer/target.cpp:36 in 7da148e. [](commit_id = 7da148e7425c20788a1e2b996a1238b7571e8528, deletion_comment = False)

MariusNi:
>         return tempdir + "." + std::to_string(counter++);

Can this counter overflow?

---
Refers to: src/tests/fuzzer/target.cpp:65 in 7da148e. [](commit_id = 7da148e7425c20788a1e2b996a1238b7571e8528, deletion_comment = False)

MariusNi:
>     std::string makeTempfile(const char* data, std::size_t size) const noexcept(false)

We use in this project PascalCasing for function names, and camelCasing for variable and constant names. 

Here correct name to respect this rule would be:

'MakeTemporaryFile' or 'MakeTempFile'

Same applies for all other similar instances in this file.

---
Refers to: src/tests/fuzzer/target.cpp:68 in 7da148e. [](commit_id = 7da148e7425c20788a1e2b996a1238b7571e8528, deletion_comment = False)

MariusNi:
> // }

Consider removing these (we do not have all internal function fuzzed anyway so no need to list the ones that are skipped)?

---
Refers to: src/tests/fuzzer/target.cpp:323 in 7da148e. [](commit_id = 7da148e7425c20788a1e2b996a1238b7571e8528, deletion_comment = False)

MariusNi:
> #include <unistd.h>

Why not adding a precompiled header and moving the includes there?

---
Refers to: src/tests/fuzzer/target.cpp:5 in 7da148e. [](commit_id = 7da148e7425c20788a1e2b996a1238b7571e8528, deletion_comment = False)

MariusNi:
> struct Context

Can this be renamed perhaps 'FuzzerContext'?

Also, can the file be renamed from 'Target.cpp' to 'FuzzingTest' or similar, to better describe its purpose?

---
Refers to: src/tests/fuzzer/target.cpp:32 in 7da148e. [](commit_id = 7da148e7425c20788a1e2b996a1238b7571e8528, deletion_comment = False)

robertwoj-microsoft:
@MariusNi I cannot reply to your comments since your review is not targeting any code in this PR. These should be added while the fuzzer initial implementation was submitted for review.

robertwoj-microsoft:
Regarding precompiled headers: please refer me to place in the project where we already do precompile headers and I'll stick to it.

robertwoj-microsoft:
Regarding commented out code: this was left intentionally and was waiting for your fixes. Since they landed, I'll review them and uncomment these that do not come with risks (i.e. arbitrary command execution potential). Unapplicable code will be dropped.

robertwoj-microsoft:
Regarding global variable: this is intentional as we need a single point of initialization of the library.  
Having a global seemed easiest and is widely used within the project which is why I picked this solution. I don't think globals are C-specific, but up to you really.
This can be achieved in various ways, if this one bothers you, I can switch. I can make this class a singleton and use ::instance() method in targets, or declare it inside the LLVMFuzzerTestOneInput function and pass around as argument. Let me know which one you think is best.

robertwoj-microsoft:
Regarding throwing inside constructors: I believe it's the right way, won't fix. You don't throw inside destructors. Constructors are fine. Having initialize() function is against RAII.

robertwoj-microsoft:
Regarging counter overflow: It can, but it's only to distinguish temporary file names. Overflow should not cause tests to fail, because temporary files are cleaned up.

MariusNi:
> Regarding precompiled headers: please refer me to place in the project where we already do precompile headers and I'll stick to it.

You are right, we do not use precompiled header in the other tests. For example, just checked and we do not have it for the UTs. Also, precompiled header is mostly useful when compiling multiple source files while here we have just one. It's fine to not have one, thanks for checking.

MariusNi (approved):
Looks good to me! 👍

AhmedBM (approved):
👌

PR #804, opened 2024-11-14
Disabling atomic reference counting for Security Baseline module only when compiler is gcc 4.8

MariusNi:
… or earlier

## Description

Disabling atomic reference counting for Security Baseline module only when compiler is gcc 4.8 or earlier, in order to not disable in production builds but only in test builds done with gcc 4.8 (and not with clang or newer gcc).

## Checklist


robertschaedler3 (approved):

AhmedBM (approved):
👌

PR #807, opened 2024-11-15
Add pull request fuzzer workflow trigger with shorter timeout

robertwoj-microsoft:
## Description

Include PR to trigger fuzzer workflows. Make the timeout shorter for quicker validation.

## Checklist

- [ ] I have merged the latest `main` branch prior to this PR submission.

### DIFF
--- a/.github/workflows/fuzzing-run.yml
+++ b/.github/workflows/fuzzing-run.yml
@@ -17,6 +17,14 @@ on:
         required: false
         type: number
         default: 600
+      container-tag:
+        type: string
+        required: false
+        default: latest
+
+env:
+  MOUNT: /azure-osconfig
+  REGISTRY: ghcr.io
 
 jobs:
   package:
@@ -30,7 +38,7 @@ jobs:
 
   fuzzer-execution:
     needs: package
-    runs-on: [self-hosted, 1ES.Pool=e2e-pool, '1ES.ImageOverride=${{ inputs.target }}']
+    runs-on: 'ubuntu-latest'
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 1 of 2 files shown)


.github/workflows/fuzzing-run.yml
```diff
@@ -39,25 +48,40 @@ jobs:
...
+          working-directory: ${{ env.MOUNT }}/tests/fuzzer
+          cmd: mkdir -p ${{ env.MOUNT }}/output/artifacts && mkdir /tmp/corpus && cp -r ${{ env.MOUNT }}/src/tests/fuzzer/seed_corpus/* /tmp/corpus && chmod u+x ./osconfig-fuzzer
+
+      - if: ${{ inputs.trigger == 'workflow_dispatch' }}
+        name: Run osconfig-fuzzer
+        uses: ./.github/actions/container-exec
+        with:
+          container: ${{ steps.container.outputs.id }}
+          working-directory: ${{ env.MOUNT }}/tests/fuzzer
+          cmd: set -o pipefail && ./osconfig-fuzzer -artifact_prefix=${{ env.MOUNT }}/output/artifacts/ -max_total_time=${{ env.WF_DISPATCH_TIMEOUT }} /tmp/corpus 2>&1 >/dev/null | tee ${{ env.MOUNT }}/output/osconfig-fuzzer.log
```
jepio, line 73: Can you collapse it into one step?

Probably smth like:
```
-max_total_time=${{ inputs.trigger == 'pull_request' && env.PR_TIMEOUT || env.WF_DISPATCH_TIMEOUT }}
```

.github/workflows/fuzzing.yml
```diff
@@ -2,6 +2,7 @@ name: Fuzzer execution
 
 on:
   workflow_dispatch:
+  pull_request:
```
MariusNi, line 5: >pull_request: [](http://example.com/codeflow?start=2&length=13)

Most project CI runs complete in seconds. Ours is already very long. Adding 5 minutes for fuzzing makes it longer. 

Why cannot we run the fuzzer just nightly and for longer duration rather than fuzz at every PR?

SethBe: I think nightly and longer fuzzing is probably going to give us more bang for our buck. As i don't know how much benefit the 5 minutes will bring, but i do no the 5 minutes more in a pipelines can be quite annoying to dev's want to build.

Since this is an area of contention and because for the task at hand we only need to run for the release. I would suggest the nightly run and then revisit this at a later date to fully understand the benefit of the PR path.
I would suggest creating a workitem to make sure we follow up at a later date.

robertwoj-microsoft:
Short fuzzer execution is useful for validating code changes where some new code paths are added and which were not included in unit testing. Because code coverage is used for guidance, it's easy for the fuzzer to get onto non-complicated inputs quickly and yield results before code is merged by blocking the PR. It also serves as regression testing based on the seed corpus. 

But if you feel 5 minuts is too much, then I won't insist. Closing this, feel free to reopen when you feel it'd be beneficial.

PR #808, opened 2024-11-15
Fix FormatAllocateString complexity

jepio:
## Description

FormatAllocateString has multiple issues:
- it always allocates at least 512 bytes even when the input is shorter
- once the length of the formatted output exceeds 512 bytes it starts incrementing the buffer size by 1 byte and goes through a `malloc,snprintf,free` loop until it finds the right size. This has an O(n^2) complexity and is slow for longer strings.
- it has an unnecessary `DuplicateString` at the end

Rewrite FormatAllocateString to call `snprintf` once to calculate the required buffer size, then `malloc` the right buffer and then format the string.

This change is motivated by hitting this edge case through fuzzing `CheckFileSystemMountingOption` where a lot of reason strings get added to the reason buffer. I've added a test case for long strings and a fuzzer case whose runtime goes from ~3s to 10ms.

## Checklist

### DIFF
--- a/src/common/commonutils/OtherUtils.c
+++ b/src/common/commonutils/OtherUtils.c
@@ -32,42 +32,35 @@ char* DuplicateStringToLowercase(const char* source)
 
 char* FormatAllocateString(const char* format, ...)
 {
-    char* buffer = NULL;
     char* stringToReturn = NULL;
     int formatResult = 0;
-    int sizeOfBuffer = MAX_STRING_LENGTH;
+    int sizeOfBuffer = 0;
 
-    if (NULL == format) 
+    if (NULL == format)
     {
         return stringToReturn;
     }
 
-    while (NULL == stringToReturn)
+    va_list arguments;
+    va_start(arguments, format);
+    // snprintf returns the required buffer size, excluding the null terminator
+    sizeOfBuffer = vsnprintf(NULL, 0, format, arguments);
+    va_end(arguments);
+
+    if (sizeOfBuffer >= 0)
     {
-        if (NULL == (buffer = malloc(sizeOfBuffer)))
+        if (NULL != (stringToReturn = malloc((size_t)sizeOfBuffer + 1)))
         {
-            break;
+            va_start(arguments, format);
+            formatResult = vsnprintf(stringToReturn, sizeOfBuffer + 1, format, arguments);
+            va_end(arguments);
+
+            if ((formatResult < 0) || (formatResult > sizeOfBuffer))
+            {
+                FREE_MEMORY(stringToReturn);
+            }
         }
-
-        memset(buffer, 0, sizeOfBuffer);
-
-        va_list arguments;
-        va_start(arguments, format);
-        formatResult = vsnprintf(buffer, sizeOfBuffer, format, arguments);
-        va_end(arguments);
-
-        if ((formatResult > 0) && (formatResult < (int)sizeOfBuffer))
-        {
-            stringToReturn = DuplicateString(buffer);
-            break;
-        }
-
-        FREE_MEMORY(buffer);
-        sizeOfBuffer += 1;
     }
-
-    FREE_MEMORY(buffer);
-
     return stringToReturn;
 }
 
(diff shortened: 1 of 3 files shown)


src/common/commonutils/OtherUtils.c
```diff
@@ -32,42 +32,41 @@ char* DuplicateStringToLowercase(const char* source)
...
+    formatResult = vsnprintf(stringToReturn, sizeOfBuffer + 1, format, arguments);
+    va_end(arguments);
 
+    if ((formatResult < 0) || (formatResult > sizeOfBuffer))
+    {
+        FREE_MEMORY(stringToReturn);
+        return NULL;
+    }
     return stringToReturn;
 }
```
MariusNi, line 71: This is a great idea! Just a nit suggestion, same code, just bit changed to only return once after initial input args validation (for consistence with rest of code), how about this?

char* FormatAllocateString(const char* format, ...)
{
    char* stringToReturn = NULL;
    int formatResult = 0;
    int sizeOfBuffer = 0;

    if (NULL == format)
    {
        return stringToReturn;
    }

    va_list arguments;
    va_start(arguments, format);
    // snprintf returns the required buffer size, excluding the null terminator
    sizeOfBuffer = vsnprintf(NULL, 0, format, arguments);
    va_end(arguments);

    if (sizeOfBuffer > 0)
    {
        if (NULL != (stringToReturn = malloc((size_t)sizeOfBuffer + 1)))
        {
            va_start(arguments, format);
            formatResult = vsnprintf(stringToReturn, sizeOfBuffer + 1, format, arguments);
            va_end(arguments);

            if ((formatResult < 0) || (formatResult > sizeOfBuffer))
            {
                FREE_MEMORY(stringToReturn);
            }
        }
    }

    return stringToReturn;
}

MariusNi (approved):
:shipit:

src/common/tests/CommonUtilsUT.cpp
```diff
@@ -1155,6 +1155,20 @@ TEST_F(CommonUtilsTest, FormatAllocateString)
...
+    std::string a(strLength, 'a');
+    std::string b(strLength, 'b');
+    formatted = FormatAllocateString("%s %s", a.c_str(), b.c_str());
+    EXPECT_NE(nullptr, formatted);
+    EXPECT_EQ(strLength * 2 + 1, strlen(formatted));
+    EXPECT_EQ(0, memcmp(formatted, a.c_str(), strLength));
+    EXPECT_EQ(0, memcmp(formatted + strLength + 1, b.c_str(), strLength));
+    FREE_MEMORY(formatted);
+}
+
```
MariusNi, line 1171: You may have noticed that although the wrapper of this test is C++, the actual test cases use C style for things such as allocating memory (with malloc), plain char* strings (not std::string), etc., in order to mimic real use of these functions under test as real as possible. 

It's not a big deal if not, but would you please consider:

1. Writing this test case in similar fashion.
2. Consider merging it with the existing FormatAllocateString test case.

We only need added one case for strings longer than 512 bytes, correct? (meaning, the test could be also simplified a bit)

jepio: Ok, merged with `FormatAllocateString` and rewrote+simplified the test case.
One test case with a string longer than 512 bytes is enough.

src/common/commonutils/OtherUtils.c
jepio: Restructured to match the structure you suggested, with one difference: `if (sizeOfBuffer >= 0)` instead of `if (sizeOfBuffer > 0)`. This is so that we don't return NULL on an invocation with valid inputs (for example `FormatAllocateString("%s", "")`). There is an added test case for this.

MariusNi: Thanks!

src/common/tests/CommonUtilsUT.cpp
MariusNi: Awesome. Thank you!

MariusNi (approved):
Looks good to me! 👍

robertschaedler3 (approved):

PR #811, opened 2024-11-15
moduletest: add --config option

kkanas:
## Description

This allows testing with various osconfig.json and will remove need for file copy into /etc/

## Checklist

- [x]  All unit tests are passing.

### DIFF
--- a/.github/workflows/module-test-run.yml
+++ b/.github/workflows/module-test-run.yml
@@ -49,6 +49,10 @@ jobs:
           for recipe in $recipes; do
             name=$(basename $recipe | tr '[:upper:]' '[:lower:]' | sed 's/\.[^.]*$//' | sed 's/\(test\|tests\)$//')
 
+            ls -al ../../src/adapters/pnp/daemon/osconfig.json
+            find ../.. -name osconfig.json
+            echo Second find
+            find ../../.. -name osconfig.json
             echo -n "testing $name ... "
 
             if output=$(sudo ./moduletest $recipe --bin ../bin --config ../../src/adapters/pnp/daemon/osconfig.json ); then
--- a/.github/workflows/module-test-run.yml
+++ b/.github/workflows/module-test-run.yml
@@ -38,11 +38,6 @@ jobs:
         with:
           name: ${{ inputs.target }}
 
-      - name: Create osconfig.json
-        run: |
-          sudo mkdir -p /etc/osconfig
-          sudo cp -r ./src/adapters/pnp/daemon/osconfig.json /etc/osconfig/osconfig.json
-
       - name: Run moduletest
         working-directory: ${{ steps.download.outputs.download-path }}/modules/test
         run: |
@@ -56,7 +51,7 @@ jobs:
 
             echo -n "testing $name ... "
 
-            if output=$(sudo ./moduletest $recipe --bin ../bin); then
+            if output=$(sudo ./moduletest $recipe --bin ../bin --config ../../src/adapters/pnp/daemon/osconfig.json ); then
               echo passed
             else
               echo failed
--- a/src/modules/test/main.c
+++ b/src/modules/test/main.c
@@ -6,7 +6,7 @@
 #include <math.h>
 
 #define DEFAULT_BIN_PATH "/usr/lib/osconfig"
-#define OSCONFIG_CONFIG_FILE "/etc/osconfig/osconfig.json"
+#define DEFAULT_OSCONFIG_CONFIG_FILE "/etc/osconfig/osconfig.json"
 
 #define AZURE_OSCONFIG "Azure OSConfig"
 
@@ -92,6 +92,7 @@ typedef struct FAILURE
 } FAILURE;
 
 static bool g_verbose = false;
+static const char* g_osconfig_config_file = DEFAULT_OSCONFIG_CONFIG_FILE;
 
 void FreeStep(STEP* step)
 {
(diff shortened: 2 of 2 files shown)


src/modules/test/main.c
```diff
@@ -6,7 +6,7 @@
 #include <math.h>
 
 #define DEFAULT_BIN_PATH "/usr/lib/osconfig"
-#define OSCONFIG_CONFIG_FILE "/etc/osconfig/osconfig.json"
+#define DEFAULT_OSCONFIG_CONFIG_FILE "/etc/osconfig/osconfig.json"
```
MariusNi, line 9: Why changing this?

.github/workflows/module-test-run.yml
```diff
@@ -56,7 +51,7 @@ jobs:
 
             echo -n "testing $name ... "
 
-            if output=$(sudo ./moduletest $recipe --bin ../bin); then
+            if output=$(sudo ./moduletest $recipe --bin ../bin --config ../../src/adapters/pnp/daemon/osconfig.json ); then
```
MariusNi, line 54: >../bin --config ../../src/adapters/pnp/daemon/osconfig.json [](http://example.com/codeflow?start=56&length=59)

Why?

moduletest supports command line arguments, if invoked manually. We do not need more for now.

Can you please explain why do you believe this change is necessary and what needs is going to solve?

Thanks

SethBe:
This would really help me as a reviewer.
in addition to the description, it would be really helpful to include the motivation for the change.
Typically in other projects, its required to have   What the change is?  Why the change is needed? and how it was validated in the description. I feel like we are just missing the why :).

.github/workflows/module-test-run.yml
kkanas: Allowing moduletest to get configuration file as argument we can optimize module-test-run by eliminating duplicate copy.
This allows to:
- Test with different configurations without necessary copy
- Allow to run moduletest without need access to /etc file system

So instead of  flow
cp ./src/adapters/pnp/daemon/osconfig.json  /etc/osconfig/osconfig.json 
$EDIT /etc/osconfig/osconfig.json (may require root)
./moduletest $recipe --bin ../bin

you can just do

$EDIT  ./src/adapters/pnp/daemon/osconfig.json
./moduletest $recipe --bin ../bin --config ../../src/adapters/pnp/daemon/osconfig.json

Moreover you can have multiple  ./src/adapters/pnp/daemon/osconfig.json and you can juggle them using cmdline switch not cp  ./src/adapters/pnp/daemon/osconfig.json.1 /etc.osconfig/osconfig.json  and remember which one you tested with.

src/modules/test/main.c
kkanas: Prior to that change OSCONFIG_CONFIG_FILE ("/etc/osconfig/osconfig.json") was the configuration file used by moduletest.
After that change "/etc/osconfig/osconfig.json" is the default osconfig config file hence name change.

In another words I am chaning the define name to match the usage of the defile

kkanas:
> This would really help me as a reviewer. in addition to the description, it would be really helpful to include the motivation for the change. Typically in other projects, its required to have What the change is? Why the change is needed? and how it was validated in the description. I feel like we are just missing the why :).

Unfortunately I don't have  right to add you to review.
Anything you want to comment in particular. I think the change is small enough that I can understand your comments if you add them as file:Line reference

src/modules/test/main.c
MariusNi: That file is the OSConfig configuration file as described in the documentation.  It shall not be duplicated or forked or else. The test is only using it here to load the PnP model version. That shall not be changed.  This PR should be recalled.

.github/workflows/module-test-run.yml
MariusNi: There is no real motivation for this change other than maybe not have to run elevated the tests. The tests need to be run elevated regardless of this file (which again, is the global OSConfig configuration file, is not test configuration or anything of that sort) because it's functional tests that mimmick OSConfig usage of the modules under test (e.g. the Security Baseline module must be executed from elevated process, etc).

This PR should be recalled.

MariusNi: Even the title of the PR is misleading - that file is NOT the configuration of the test....

src/modules/test/main.c
MariusNi: If there is a desire to simplify this and make moduletest don't read from /etc, simplest way is to remove this dependency on osconfig.json that is not truly necessary for currently active modules (it is necessary for the Firewall module which is currently paused; that module, and there can be others, validate that the client name for MMI calls follows the required template used by OSConfig). This can be done by changing GetClientName in src\modules\test\main.c when a new command line arg is provided to return a hard-coded string that does not contain the OSConfig model and version but instead just 'OSConfigModulesTest' for example. 

Please note however that there are no scenarios asking for such change and than when Firewall or other modules will be readded to test, that will be at risk. 

So best is at this time to just recall this PR, unless there is a need for fuzzing or something else that has impact on next release, in which case we need to discuss this team triage.

Thanks for your cooperation!

MariusNi (commented):
Think that now I understand what happens and what led to this PR. To quote from the description: "This allows testing with various osconfig.json and will remove need for file copy into /etc/"

We OSConfig Core team build OSConfig locally with the option to install it, which also binplaces osconfig.json under /etc. So we never have to manually copy it there.

I get a feeling that you do not like to build with sudo and use install, so that's why you have to manually copy the file and that led to this PR.

Here's an idea how to adapt in the least invasive and risky way:

Add a new command line argument to say for example '-isolated' and when that option is checked, the GetClientName to return a hard-coded string "OSConfigModuleTest". And when is not checked, GetClientName to do what is doing now, read from  /etc.

See, the PR description was confusing as it claimed the file is copied by the test somehow, yet it is not. What led us asking 'why?'

jepio:
@kkanas: the module tests are failing for various reasons related to this change. Do you agree that it would make sense to drop this for now and revisit at a later time?

kkanas:
> @kkanas: the module tests are failing for various reasons related to this change. Do you agree that it would make sense to drop this for now and revisit at a later time?

@jepio - yes lets revisit this at later time.  It started as small improvement that I did for myself to test without /etc/osconfig/osconfig.json which sounded easy to share but  given @MariusNi  proposal to fix correctly the scope is bigger than anticipated, so I am ok with skipping it for now

PR #812, opened 2024-11-18
Fix slow execution of CheckTextNotFoundInEnvironmentVariable function

robertwoj-microsoft:
## Description

Break text search loop when pattern is found. This fix avoids long running executions by repetitively scanning list of environment variables if pattern occurs frequently in the searched string.

## Checklist

### DIFF
--- a/src/common/commonutils/FileUtils.c
+++ b/src/common/commonutils/FileUtils.c
@@ -1331,6 +1331,7 @@ int CheckTextNotFoundInEnvironmentVariable(const char* variableName, const char*
                         OsConfigCaptureReason(reason, "'%s' found in '%s' ('%s')", text, variableName, found);
                         foundText = true;
                         status = EEXIST;
+                        break;
                     }
                 } 
             


src/common/commonutils/FileUtils.c
```diff
@@ -1331,6 +1331,7 @@ int CheckTextNotFoundInEnvironmentVariable(const char* variableName, const char*
                         OsConfigCaptureReason(reason, "'%s' found in '%s' ('%s')", text, variableName, found);
                         foundText = true;
                         status = EEXIST;
+                        break;
```
MariusNi, line 1334: >break; [](http://example.com/codeflow?start=24&length=6)

The existing code's intention is to continue to search all instances in the text. Here it record a match was found but does not break, continues. That's in order for all instances to be logged and captured in the reason (as opposed to bail out at the first found instance). Does this make sense? Please let me know if you need to discuss this further.

MariusNi:
Re: tile of the PR. We do not use here in GitHub internal bug or task ids (which are just that, internal). Instead, we use full descriptive names and link the PR into that internal bug or task, if any. Also, please note that for such internal bugs, there shall be no PRs opened until that bug is approved in triage.

src/common/commonutils/FileUtils.c
robertwoj-microsoft: Do we need this logging? Continuing iteration is waste of resources if there is no specific requirement.

robertwoj-microsoft:
Changed PR title.

src/common/commonutils/FileUtils.c
MariusNi: To follow up on my last reply (can you please answer that question? thx) - if it's just a transient delay in fuzzing, that is fine, is due to how this function works. However, if it ends into a denial of service problem with an infinite loop, then we need a fix to harden the function or we need to skip this function from fuzzing.

The hardening fix would be not to change the internal algorithm of the function, but instead before executing that, check the file size and if is bigger than a certain size, bail out right there, reporting as reason that the file is too large to be processed.

PR #813, opened 2024-11-19
Update test matrix and add module tests to NRP test workflow

robertschaedler3:
## Description

Update test matrix and add module tests to NRP test workflow.

## Checklist

### DIFF
--- a/.github/workflows/universalnrp-test.yml
+++ b/.github/workflows/universalnrp-test.yml
@@ -16,6 +16,7 @@ jobs:
     name: Package
     if: ${{ inputs.policy_packages == '[]' || inputs.policy_packages == '' }}
     uses: ./.github/workflows/package-build.yml
+    needs: [setup-matrix]
     with:
       target: ubuntu-14.04
       arch: amd64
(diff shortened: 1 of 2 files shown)


MariusNi (approved):
:shipit:

.github/workflows/universalnrp-test.yml
```diff
@@ -183,18 +173,94 @@ jobs:
...
-          path: '${{ steps.normalize.outputs.PolicyPackageDir }}/*testResults.xml'
+          path: '${{ steps.normalize.outputs.dir }}/*testResults.xml'
 
       - uses: actions/upload-artifact@v4
         if: success() || failure()
         with:
           name: ${{ steps.normalize.outputs.name }}_logs
           path: osconfig-logs/osconfig*
 
+  module-test:
```
AhmedBM, line 184: Why are we moving `Module Test` into this workflow? Isn't it better to have it broken out to a different workflow given it also has different targets?

.github/workflows/universalnrp-test.yml
```diff
@@ -3,127 +3,120 @@ name: Universal NRP Test
...
         description: 'List of policy packages to test'
         required: true
         default: '[]'
   pull_request:
   schedule:
     - cron: '0 20 * * *' # Every day at 12pm PST (UTC-8)
 
-env:
-  # Default build matrix - this is where additional distros are added, policy-package are defined either through the workflow_dispatch input or the DEFAULT_PACKAGE_LIST
-  BUILD_MATRIX: |
```
AhmedBM, line 16: I prefer having the configurable knobs (distros/packages/arch) available at the top of the workflow as workflow globals, why move them below? Its easier to manage this way, at least that was my intent when i did it :)

AhmedBM:
Does the workflow work with/without policy packages given?

.github/workflows/universalnrp-test.yml
robertschaedler3: This module test run uses the same "universal" shared build artifacts as the MC tests and only tests the security baseline recipe across all the target distros.

.github/workflows/universalnrp-test.yml
robertschaedler3: The policy packages are only relevant to the MC test workflow so I moved them into that env so that they do not show up on the env of every job/step.

I'd prefer if it wasn't JSON but since we have 2 jobs sharing the same set of targets it makes sense to have a shared variable. I'll move some stuff around to make the targets shared between the 2 jobs too. :)

.github/workflows/universalnrp-test.yml
```diff
@@ -3,127 +3,125 @@ name: Universal NRP Test
...
+              { "os": "ubuntu", "version": "22.04" }
+            ]
+          # { "os": "almalinux", "version": "9" },
+          # { "os": "amazonlinux", "version": "2" },
+          # { "os": "centos", "version": "7" },
+          # { "os": "oraclelinux", "version": "7" },
+          # { "os": "rhel", "version": "7" },
+          # { "os": "sles", "version": "12" },
+          # { "os": "ubuntu", "version": "16.04" },
+          # { "os": "ubuntu", "version": "18.04" },
```
MariusNi, line 67: nit: inserting them in the TARGETS to keep the  original order would help longer term, rather than separating them here

.github/workflows/universalnrp-test.yml
MariusNi: Ahmed, we discussed about this and got agreement (you were present).

.github/workflows/universalnrp-test.yml
robertschaedler3: Unfortunately, they need to stay separate since the `TARGETS` env is a JSON string. As I uncomment them and add them back in later PRs I'll make sure to insert them in the correct order.

MariusNi (approved):
Looks good to me 👍

AhmedBM (approved):
Approved as long as this works with/without policy packages provided

AhmedBM:
Looks like this change breaks the case when policy packages are defined. Ill fix in a follow-up PR

PR #814, opened 2024-11-20
Add ASA scans running on CentOS 8

danielszot:
## Description

* ASA Scans running on CentOS 8 Added.
* Base container images locked by SHA256 to improve hermeticity.

## Checklist

### DIFF
--- a/devops/scripts/asa/run_asa.sh
+++ b/devops/scripts/asa/run_asa.sh
@@ -8,16 +8,33 @@ set -e
 # Tested and supported runtimes: podman, docker
 CONTAINER_RUNTIME="podman"
 
+DISTROS=("ubuntu22" "centos8")
+
+function echo_header {
+  echo ""
+  echo "============================================================="
+  echo ${1}
+  echo "============================================================="
+}
+
 if ! dpkg --list ${CONTAINER_RUNTIME} > /dev/null && ! rpm -q ${CONTAINER_RUNTIME} > /dev/null; then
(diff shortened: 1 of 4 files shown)


devops/scripts/asa/run_asa.sh
```diff
@@ -8,16 +8,33 @@ set -e
 # Tested and supported runtimes: podman, docker
 CONTAINER_RUNTIME="podman"
```
AhmedBM, line 9: can we have this resolved at runtime? eg. Use whichever is present podman/docker and report error if none?

AhmedBM (approved):
Just realized the container runtime definition wasn't even part of the change. Would be a great nice-to-have though

devops/scripts/asa/run_asa.sh
AhmedBM: sorry just realized this wasn't even part of this changeset, im just thinking out loud here 😆

danielszot: Hehe, that's totally fine! By that I seen that we have podman set as default and it was supposed to be docker. I'll change that and will ask for re-approval.

I'll do runtime detection with priority to the first on the list in the follow up PR. That's not time consuming and rather straight forward.

MariusNi (approved):
:shipit:

danielszot:
Failing check seems not related to the PR itself.
It was also failing in the last merged PR (https://github.com/Azure/azure-osconfig/pull/813)
I think this one is rather done. @AhmedBM @MariusNi can you merge it?

PR #816, opened 2024-11-21
github: Run fuzzing nightly

jepio:
## Description

Add a trigger to the fuzzing workflow that runs it nightly. The fuzzing runs will need to be periodically reviewed to see if there are any failures.

## Checklist

### DIFF
--- a/.github/workflows/fuzzing.yml
+++ b/.github/workflows/fuzzing.yml
@@ -2,6 +2,8 @@ name: Fuzzer execution
 
 on:
   workflow_dispatch:
+  schedule:
+    - cron: '0 3 * * *' # Every day at 7pm PST/4am CET
 
 jobs:
   fuzzer-execution:


.github/workflows/fuzzing.yml
```diff
@@ -2,6 +2,8 @@ name: Fuzzer execution
 
 on:
   workflow_dispatch:
+  schedule:
+    - cron: '0 3 * * *'
```
AhmedBM, line 6: Can we also leave a comment with the frequency, similar to what we have here https://github.com/Azure/azure-osconfig/blob/d29a9ac7bd7569abef2de58de4c1ebd843370a61/.github/workflows/insiders-fast.yml#L6

AhmedBM:
LGTM just some comments on the frequency of the workflow

MariusNi (approved):
:shipit:

.github/workflows/fuzzing.yml
jepio: Added a comment

robertschaedler3 (approved):

PR #818, opened 2024-11-21
Fix module test failure on Debian 12

jepio:
## Description

Two issues:
### apt-cache
On a freshly configured system (VM/container/server) the apt cache may not be populated initially.  This causes any remediation task that requires package installation to happen to fail. Ensure we update the package cache on the first invocation of a function that attempts to install a package.

With this change `apt-get update` is done once and only on demand. Some remediation functions need to be revised to better interact with this, for example `RemediateEnsureAuditdServiceIsRunning`. `RemediateEnsureAuditdServiceIsRunning` attempts to install 4 different package candidates, some of which may not be available in a distro version at all. Each such probe could trigger an `apt-get update`, even if a later package in the list is already installed. This scenario is not critical because:
- for MC there is a `Test()` call that would never invoke `apt-get upate` and that would suppress the call to `Set()`
- for standalone Osconfig we would have one `apt-get update` per lifetime of `osconfig-platform` daemon which is acceptable.

### cronie

~The order of remediate tasks differs between moduletest and MC policy. The remediation tasks that modify permissions of cron files in `/etc/` do nothing until these files are created by installing crond. Subsequent audit tasks related to these cron files can fail. Move the `remediateEnsureCronServiceIsEnabled` task in moduletest earlier, so that the files are created before the permission remediation tasks run.~
dropped, because it's superseded by #821 which has the correct ordering.

## Checklist

### DIFF
--- a/src/common/commonutils/PackageUtils.c
+++ b/src/common/commonutils/PackageUtils.c
@@ -17,6 +17,7 @@ static bool g_tdnfIsPresent = false;
 static bool g_dnfIsPresent = false;
 static bool g_yumIsPresent = false;
 static bool g_zypperIsPresent = false;
+static bool g_aptGetUpdateExecuted = false;
 
 int IsPresent(const char* what, void* log)
 {
@@ -178,6 +179,26 @@ int CheckPackageNotInstalled(const char* packageName, char** reason, void* log)
     return result;
 }
 
+void AptGetUpdateOnce(void* log)
+{
+    const char* command = "apt-get update";
+    int status = 0;
+    if (g_aptGetUpdateExecuted)
+    {
+        return;
+    }
+
+    if (0 == (status = ExecuteCommand(NULL, command, false, false, 0, 0, NULL, NULL, log)))
+    {
+        OsConfigLogInfo(log, "AptGetUpdateOnce: '%s' was successful", command);
+        g_aptGetUpdateExecuted = true;
+    }
+    else
+    {
+        OsConfigLogError(log, "AptGetUpdateOnce: '%s' failed with %d", command, status);
+    }
+}
+
 int InstallOrUpdatePackage(const char* packageName, void* log)
 {
     const char* commandTemplate = "%s install -y %s";
@@ -187,6 +208,7 @@ int InstallOrUpdatePackage(const char* packageName, void* log)
     
     if (g_aptGetIsPresent)
     {
+        AptGetUpdateOnce(log);
         status = CheckOrInstallPackage(commandTemplate, g_aptGet, packageName, log);
     }
     else if (g_tdnfIsPresent)


jepio:
> Two issues:

Happy to split the PR and or delay this (since it only affects remediation), but combined here for narrowing down moduletest failures.

src/common/commonutils/PackageUtils.c
```diff
@@ -178,6 +179,28 @@ int CheckPackageNotInstalled(const char* packageName, char** reason, void* log)
...
+    if (0 == status)
+    {
+        OsConfigLogInfo(log, "AptGetUpdateOnce: apt-get update was successful");
+    }
+    else
+    {
+        OsConfigLogError(log, "AptGetUpdateOnce: apt-get update failed with %d", status);
+    }
+
+    g_aptGetUpdateExecuted = true;
```
AhmedBM, line 201: Should we set to true even though it fails on L198 so it can try again on a subsequent run?

src/common/commonutils/PackageUtils.c
```diff
@@ -178,6 +179,28 @@ int CheckPackageNotInstalled(const char* packageName, char** reason, void* log)
...
+{
+    const char* command = "apt-get update";
+    int status = 0;
+    if (g_aptGetUpdateExecuted)
+    {
+        return;
+    }
+
+    status = ExecuteCommand(NULL, command, false, false, 0, 0, NULL, NULL, log);
+    if (0 == status)
```
MariusNi, line 192: nit: combine in one line to match existing code?

Like this:

```
if (0 == (status = ExecuteCommand(NULL, command, false, false, 0, 0, NULL, NULL, log)))
```

MariusNi:
Yes, second that, this is an awesome find. Also same question (do we need to retry if the command fails?)

---
In reply to: [2452224606](https://github.com/Azure/azure-osconfig/pull/818#issuecomment-2452224606) [](http://example.com/codeflow?ancestors=2452224606)

MariusNi:
No need to split in my opinion. If you can add the fix for the ModuleTest (the SecueityBaseline module test recipe) in this PR, it would be great. Make sure that you will fix the ordering there of the MIM objects under test to match the policy MOF for all sequences, Get/Audit, Init, and Set/Remediate. The test recipe is at [https://github.com/Azure/azure-osconfig/blob/main/src/modules/test/recipes/SecurityBaselineTests.json](https://github.com/Azure/azure-osconfig/blob/main/src/modules/test/recipes/SecurityBaselineTests.json). Please validate this locally before submitting to PR, with: ```sudo modules/test/moduletest ../src/modules/test/recipes/SecurityBaselineTests.json```

---
In reply to: [2491873158](https://github.com/Azure/azure-osconfig/pull/818#issuecomment-2491873158) [](http://example.com/codeflow?ancestors=2491873158)

MariusNi:
Another thing to consider, we cannot rely on the exact order in which the MC calls will arrive. So we may need also a change in Asb.c where in case one check is executed before the other, the necessary update to happen (with a new global there like here). 

---
In reply to: [2492095717](https://github.com/Azure/azure-osconfig/pull/818#issuecomment-2492095717) [](http://example.com/codeflow?ancestors=2492095717,2491873158)

src/common/commonutils/PackageUtils.c
jepio: Sure, let's try that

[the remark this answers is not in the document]
jepio: Sure, will do

src/common/commonutils/PackageUtils.c
jepio: done

src/common/commonutils/PackageUtils.c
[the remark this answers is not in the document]
jepio: done

jepio:
> No need to split in my opinion. If you can add the fix for the ModuleTest (the SecurityBaseline module test recipe) in this PR, it would be great. Make sure that you will fix the ordering there of the MIM objects under test to match the policy MOF for all sequences, Get/Audit, Init, and Set/Remediate. The test recipe is at https://github.com/Azure/azure-osconfig/blob/main/src/modules/test/recipes/SecurityBaselineTests.json. Please validate this locally before submitting to PR, with: sudo modules/test/moduletest ../src/modules/test/recipes/SecurityBaselineTests.json

I've pushed this part out to a separate PR because of the size of the changes: https://github.com/Azure/azure-osconfig/pull/821. The PR is draft for now. We also need to make sure all CI jobs use the same set of test recipes.

> Another thing to consider, we cannot rely on the exact order in which the MC calls will arrive. So we may need also a change in Asb.c where in case one check is executed before the other, the necessary update to happen (with a new global there like here).

Could you explain what you mean by "we cannot rely on the exact order in which the MC calls will arrive"? I thought the order of operations is defined by the MOF file.

jepio:
A different fix is necessary for RHEL9: #822

AhmedBM:
Should we abandon this PR given its covered in #823 ?

src/common/commonutils/PackageUtils.c
```diff
@@ -178,6 +179,27 @@ int CheckPackageNotInstalled(const char* packageName, char** reason, void* log)
...
+    }
+
+    if (0 == (status = ExecuteCommand(NULL, command, false, false, 0, 0, NULL, NULL, log)))
+    {
+        OsConfigLogInfo(log, "AptGetUpdateOnce: apt-get update was successful");
+        g_aptGetUpdateExecuted = true;
+    }
+    else
+    {
+        OsConfigLogError(log, "AptGetUpdateOnce: apt-get update failed with %d", status);
```
MariusNi, line 198: nit: wrap the command in ' ' in traces, and reuse the constant on line 184 (e.g.: "....'%s' was successful", command).

src/common/commonutils/PackageUtils.c
```diff
@@ -178,6 +179,27 @@ int CheckPackageNotInstalled(const char* packageName, char** reason, void* log)
...
+    }
```
MariusNi, line 199: >  } [](http://example.com/codeflow?start=2&length=3)

nit: line 200 is blank, please remove

MariusNi:
About the order. MC makes Get, Test and Set calls on individual checks (or rules, each represented by an instance of the resource class). The order in which these calls arrive is not guaranteed to be the order in MOF, although usually it is we cannot hard rely on this and expect any programmed order of calls. On the other hand, when security baselines are audited via OSConfig (not via MC, but via another entity and RC/DC, for example) where each rule is represented by a MIM object payload, those MIM objects can be set in any order. Please let me know if this is still unclear and reach to discuss if necessary.

---
In reply to: [2498786828](https://github.com/Azure/azure-osconfig/pull/818#issuecomment-2498786828) [](http://example.com/codeflow?ancestors=2498786828)

MariusNi:
IMHO, would be better to do these PRs separately. Beginning of this week starting on 12/2 we are going to 'cut' a securely built policy package and begin the validation with vendors. That package, once signed off in this validation, will then go for publication for Private Preview, etc. We need to triage the existing PRs and determine which ones must be in this release and which ones can wait. Reach to me to discuss more if you need.

---
In reply to: [2504230163](https://github.com/Azure/azure-osconfig/pull/818#issuecomment-2504230163) [](http://example.com/codeflow?ancestors=2504230163)

src/common/commonutils/PackageUtils.c
jepio: done (also on line 193)

src/common/commonutils/PackageUtils.c
jepio: done

MariusNi (approved):
:shipit:

robertschaedler3 (approved):

PR #821, opened 2024-11-25
Update e2e test recipe for ASB

jepio:
## Description

Several related changes to the ModuleTests/E2E test recipes:

1. Rely on single set of ModuleTest recipes. Previously there were two sets of recipes: in `src/module/tests/recipes` (used by universal-nrp based ModuleTests) and `src/tests/e2e-test-recipes` (used by distro-package ModuleTests). Now both CI jobs rely on the set in `src/module/tests/recipes`.
2. Generate the `SecurityBaselineTests.json` from SSH and ASB MOF files. Expanded below.
3. Fix the ASB MOF file to include 2 missing `InitObjectName` entries (this was found by comparing the the existing test recipe with the newly generated on

### Test Recipe generation

There is a new make target `generate-asb-test-json` that runs a script to generate the file, it is meant to be run by developers when modifying the MOF files. The generated recipe contains rules from SSH and ASB MOF file, in that order. The generated recipe has the following structure:
1. load module
2. remediate phase
3. unload module
4. load module
5. audit phase
6. unload module

In each phase the rules match the order of the MOF files.
If the rule has a payload (`DesiredObjectValue`), then we run the `InitObjectName` in both the remediate and audit phases, before `DesiredObjectName`/`ReportedObjectName`. This is meant to miror the order of operations executed by MC, where `GetTargetResource` consists of (Init+Audit) and which runs `TestTargetResource` (Init+Audit) before `SetTargetResource` (Remediate). We must skip the `Audit` function before `Remediate`, not doing so would cause rules to fail depending on the initial state of the test vm.

## Checklist


jepio:
@AhmedBM, @robertschaedler3 : there are two sets of moduletest recipes: `src/modulest/test/recipes` and `src/test/e2e-test-recipes`. Can we get rid of one?

src/modules/test/recipes/create-asb-json.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
```
MariusNi, line 1: We need the copyright notice even here:

src/modules/test/recipes/mof-to-json.awk
```diff
@@ -0,0 +1,105 @@
+#!/usr/bin/awk -f
```
MariusNi, line 1: Also here:

src/modules/test/recipes/SecurityBaselineTests.json-footer
```diff
@@ -0,0 +1,4 @@
+  {
+    "Action": "UnloadModule"
+  }
+]
```
MariusNi, line 4: Concerns with this approach, we need to discuss more:

1. W3e have several other modules that are temporarily disabled from build but which we want to refactor and re-enable later. They may not work with this pattern.

2. In general, test recipes can contain anything. A module test recipe won't be required to have the footer or the header shown here for ASB one. 

3. Test recipes can unload and reload modules any number of times. It's not just one, or two. Thus, what is now split into header, mid and footer may become -if going by this pattern- header, mid1, mid2..., midN, footer.

Last and not least, are the header, footer and mid intended only for the ASB test recipe? If so, why not keep the test recipe one single JSON as it was (and in sync with other such big files, the policy MOF and the MIM, for example) which has the big benefit of keeping the recipe simple?

MariusNi:
These test recipes are per module, not per MOF/policy. There is a single test recipe for the ASB module that handles both the ASB and the SSH parts. Where the SSH one is part of the ASB. These are not the MC tests. 

About this part in the PR description, quote: 

```
Test Recipe generation
There is a new make target generate-asb-test-json that runs a script to generate the file, it is meant to be run by developers when modifying the MOF files.
```

New security baselines in the future will be serviced by new modules, each with their own test recipe. Not by the Security Baseline module which is ASB only (SSH is just a part of ASB). 

Perhaps the confusion was that this was called "security baseline" test recipe? Is only for ASB, not for other baselines. Wish now it would be called, simply, ASB test recipe.

Let's discuss.

src/modules/test/recipes/create-asb-json.sh
jepio: done

src/modules/test/recipes/SecurityBaselineTests.json-footer
jepio: Let me try to answer quickly:

> 1. We have several other modules that are temporarily disabled from build but which we want to refactor and re-enable later. They may or may not work with this pattern.

I don't aim to switch all modules to this pattern, I think it only makes sense where there is a MOF file(s) that map (1 to 1 or many to 1) to a module. Here ASB+SSH MOF are used to test the ASB module. The goal was to get the order and parameters in the test recipe to match the one in a MOF file describing a policy.

> 2. In general, test recipes can contain anything. A module test recipe won't be required to have the footer or the header shown here for ASB one.

Agree

> 3. Test recipes can unload and reload modules any number of times. It's not just one, or two times. Thus, what is now split into header, mid and footer may become (if going by this pattern) header, mid1, mid2..., midN, footer.

I'm not married to the approach - I noticed that the existing test recipe had a unload/reload in the middle between remediate and audit steps, and the reload is necessary to get the SSH settings to be applied.

> Are the header, footer and mid intended only for the ASB test recipe? If so, why not keep the test recipe one single JSON as it was (and in sync with other such big files, the policy MOF and the MIM, for example) which has the big benefit of keeping the recipe simple and easy to understand and maintain in the future?

The generated ASB file is still present: https://github.com/Azure/azure-osconfig/pull/821/files#diff-5a26e381378cc3525d9236fc6e0b756218ceaf04daf0f5f04c380954d994e751 and used. The header/footer/mid parts are just templates used as part of the generation specific to the ASB recipe. Test recipes for other policy modules (CIS) might not need a mid section.

Hope that helps clear things up

jepio:
> These test recipes are per module, not per MOF/policy. There is a single test recipe for the ASB module that handles both the ASB and the SSH parts. Where the SSH one is part of the ASB. These are not the MC tests.

We're on the same page - the ASB test recipe is generated from both SSH and ASB MOFs together here. This is what the invocation ends up looking like:
```
./create-asb-json.sh <DIR>/src/adapters/mc/ssh/LinuxSshServerSecurityBaseline.mof <DIR>/src/adapters/mc/asb/AzureLinuxBaseline.mof ><DIR>/src/modules/test/recipes/SecurityBaselineTests.json
```

> New security baselines in the future will be serviced by new modules, each with their own test recipe. Not by the Security Baseline module which is ASB only (SSH is just a part of ASB).

Agree, and as long as they have some MOF file corresponding to the module we can implement generation in a similar fashion. If they don't have a MOF file we might be able to generate a test recipe from some sample JSON config file.

MariusNi (approved):
:shipit:

robertschaedler3 (approved):

PR #822, opened 2024-11-26
Restore selinux context when replacing config files

jepio:
## Description

Several ASB ModuleTest steps have started failing after the release of RHEL 9.5. Looking into them, I found that they are failing due to issues with selinux file contexts. The failure comes from `chage` when changing password expiration settings, after a previous rule modified `/etc/shadow`,`/etc/shadow-`,`/etc/passwd`,`/etc/passwd-`. With additional output from `chage` one can see the following:
```
[2024-11-26 09:00:27] [main.c:604] Step 3 of 16
[2024-11-26 09:00:27] [main.c:614] Running desired test 'SecurityBaseline.remediateEnsurePasswordExpiration'
[2024-11-26 09:00:27] [UserUtils.c:2400] EnsureUsersHaveDatesOfLastPasswordChanges: all users who have passwords have dates of last password changes
[2024-11-26 09:00:27] [UserUtils.c:2289] SetMaxDaysBetweenPasswordChanges: user 'cloudtest' (1001, 1001) has maximum time between password changes of -1 days while requested is 365 days
[2024-11-26 09:00:27] [UserUtils.c:2309] [ERROR] SetMaxDaysBetweenPasswordChanges: chage -M 365 cloudtest: failed: chage: failure while writing changes to /etc/shadow
```
Looking at syslog shows:
```
Nov 26 09:00:34 vm setroubleshoot[9443]: SELinux is preventing /usr/bin/chage from write access on the file /etc/shadow-. For complete SELinux messages run: sealert -l 0cb08071-377d-4ba7-958d-7f9ed6c5e08e
Nov 26 09:00:34 vm setroubleshoot[9443]: SELinux is preventing /usr/bin/chage from write access on the file /etc/shadow-.

                                         *****  Plugin restorecon (94.8 confidence) suggests   ************************

                                         If you want to fix the label.
                                         /etc/shadow- default label should be shadow_t.
                                         Then you can run restorecon. The access attempt may have been stopped due to insufficient permissions to access a parent directory in which case try to change the following command accord>
                                         Do
                                         # /sbin/restorecon -v /etc/shadow-

                                         *****  Plugin catchall_labels (5.21 confidence) suggests   *******************

                                         If you want to allow chage to have write access on the shadow- file
                                         Then you need to change the label on /etc/shadow-
                                         Do
                                         # semanage fcontext -a -t FILE_TYPE '/etc/shadow-'
                                         where FILE_TYPE is one of the following: afs_cache_t, faillog_t, krb5_host_rcache_t, lastlog_t, passwd_file_t, puppet_tmp_t, security_t, shadow_t, sssd_public_t, sssd_var_lib_t, user_cron>
                                         Then execute:
                                         restorecon -v '/etc/shadow-'


                                         *****  Plugin catchall (1.44 confidence) suggests   **************************

                                         If you believe that chage should be allowed write access on the shadow- file by default.
                                         Then you should report this as a bug.
                                         You can generate a local policy module to allow this access.
                                         Do
                                         allow this access for now by executing:
                                         # ausearch -c 'chage' --raw | audit2allow -M my-chage
                                         # semodule -X 300 -i my-chage.pp
```

Restoring the selinux context shows that the file has the wrong context after ASB runs:
```
[jeremi@vm ~]$ sudo /sbin/restorecon -Fv /etc/shadow-
Relabeled /etc/shadow- from unconfined_u:object_r:etc_t:s0 to system_u:object_r:shadow_t:s0
```

The selinux context is lost when a temporary file is created to edit the contents of the config file. Restore it after renaming by running `restorecon -F`.

## Checklist

### DIFF
--- a/src/common/asb/Asb.c
+++ b/src/common/asb/Asb.c
@@ -915,6 +915,11 @@ void AsbInitialize(void* log)
         OsConfigLogInfo(log, "AsbInitialize: running on product '%s'", PRODUCT_NAME_AZURE_COMMODORE);
     }
 
+    if (DetectSelinux(log))
+    {
+        OsConfigLogInfo(log, "AsbInitialize: SELinux present");
+    }
+
     OsConfigLogInfo(log, "%s initialized", g_asbName);
 }
 
--- a/src/common/commonutils/DeviceInfoUtils.c
+++ b/src/common/commonutils/DeviceInfoUtils.c
@@ -11,6 +11,8 @@ typedef struct OS_DISTRO_INFO
     char* description;
 } OS_DISTRO_INFO;
 
+static bool g_selinuxPresent = false;
+
 void RemovePrefix(char* target, char marker)
 {
     size_t targetLength = 0, i = 0;
@@ -947,4 +949,15 @@ bool IsCommodore(void* log)
     FREE_MEMORY(textResult);
 
     return status;
+}
+
+bool IsSelinuxPresent(void)
+{
+    return g_selinuxPresent;
+}
+
+bool DetectSelinux(void* log)
+{
+    g_selinuxPresent = (0 == CheckTextIsFoundInFile("/sys/kernel/security/lsm", "selinux", NULL, log));
+    return IsSelinuxPresent();
 }
--- a/src/common/commonutils/FileUtils.c
+++ b/src/common/commonutils/FileUtils.c
@@ -872,6 +872,29 @@ int GetDirectoryAccess(const char* name, unsigned int* ownerId, unsigned int* gr
     return GetAccess(true, name, ownerId, groupId, mode, log);
 }
 
+static int RestoreSelinuxContext(const char* target, void* log)
+{
+    char* restoreCommand = NULL;
+    char* textResult = NULL;
+    int status = 0;
+
+    if (NULL == (restoreCommand = FormatAllocateString("restorecon -F '%s'", target)))
+    {
+        OsConfigLogError(log, "RestoreSelinuxContext: out of memory");
+        return ENOMEM;
+    }
+
+    if (0 != (status = ExecuteCommand(NULL, restoreCommand, false, false, 0, 0, &textResult, NULL, log)))
+    {
+        OsConfigLogError(log, "RestoreSelinuxContext: restorecon failed %d: %s", status, textResult);
+    }
+
+    FREE_MEMORY(textResult);
+    FREE_MEMORY(restoreCommand);
+
+    return status;
+}
+
 int RenameFile(const char* original, const char* target, void* log)
 {
     int status = 0;
@@ -887,7 +910,14 @@ int RenameFile(const char* original, const char* target, void* log)
         return EINVAL;
     }
 
-    if (0 != (status = rename(original, target)))
+    if (0 == (status = rename(original, target)))
+    {
+        if (IsSelinuxPresent())
+        {
+            RestoreSelinuxContext(target, log);
+        }
+    }
+    else
     {
         OsConfigLogError(log, "RenameFile: rename('%s' to '%s') failed with %d", original, target, errno);
         status = (0 == errno) ? ENOENT : errno;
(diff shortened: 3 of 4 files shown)


src/common/commonutils/FileUtils.c
```diff
@@ -893,6 +916,11 @@ int RenameFile(const char* original, const char* target, void* log)
         status = (0 == errno) ? ENOENT : errno;
     }
 
+    if (DetectSelinux(log))
```
MariusNi, line 919: >(DetectSelinux(log)) [](http://example.com/codeflow?start=7&length=20)

Should we also limit RestoreSelinuxContext() for just the affected system files? 

Also, should we condition this on the successful rename (basically to not do RestoreSelinuxContext when the target is not modified)?

src/common/asb/Asb.c
```diff
@@ -910,6 +910,11 @@ void AsbInitialize(void* log)
     FREE_MEMORY(prettyName);
     FREE_MEMORY(kernelVersion);
 
+    if (DetectSelinux(log))
+    {
+        OsConfigLogInfo(log, "AsbInitialize: SELinux present; keeping file contexts");
+    }
```
MariusNi, line 916: Move after the Commodore detection and change to simply say:

"AsbInitialize: SELinux present"

src/common/commonutils/DeviceInfoUtils.c
```diff
@@ -947,4 +947,22 @@ bool IsCommodore(void* log)
...
 
     return status;
-}
\ No newline at end of file
+}
+
+enum SelinuxState {
+  SelinuxUnknown = 0,
+  SelinuxFound,
+  SelinuxNotFound,
```
MariusNi, line 955: Unknown -1, found 0 and not found 1 would better match overall status style? 0 being true/success, etc.

src/common/commonutils/DeviceInfoUtils.c
```diff
@@ -947,4 +947,22 @@ bool IsCommodore(void* log)
...
+};
+static enum SelinuxState g_selinuxState = SelinuxUnknown;
```
MariusNi, line 957: The SelinuxState can be simplified to a Boolean if we'll initialize the global var to false aka no SELinux.

src/common/commonutils/DeviceInfoUtils.c
```diff
@@ -947,4 +947,22 @@ bool IsCommodore(void* log)
...
+
+bool DetectSelinux(void* log)
```
MariusNi, line 959: >DetectSelinux [](http://example.com/codeflow?start=5&length=13)

Together with making the enum a Boolean, we could also rename the function as 'IsSelinuxPresent' to better describe its purpose. Thoughts?

src/common/asb/Asb.c
jepio: done

src/common/commonutils/DeviceInfoUtils.c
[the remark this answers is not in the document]
jepio: done

src/common/commonutils/DeviceInfoUtils.c
jepio: dropped the enum due to a later comment ("The SelinuxState can be simplified to a Boolean")

src/common/commonutils/DeviceInfoUtils.c
jepio: switched it to a boolean, but there is a caveat - see comment below.

src/common/commonutils/DeviceInfoUtils.c
jepio: That's ok, but we need to avoid calling `CheckTextIsFoundInFile` on every invocation (it spams the log). With only one function we'd need to rely on a function static variable because this alone wouldn't work:
```
bool IsSelinuxPresent(void* log)
{
    if (!g_selinuxPresent)
    {
        g_selinuxPresent = (0 == CheckTextIsFoundInFile("/sys/kernel/security/lsm", "selinux", NULL, log));
    }
    return g_selinuxPresent;
}
```

 So I split the code into a function that does probing `DetectSelinux` and one that returns the status `IsSelinuxPresent`. Let me know if you see a different way?

MariusNi (approved):
:shipit:

robertschaedler3 (approved):

PR #835, opened 2024-12-16
Fix ASB v2's EnsureSourceRoutedPacketsIsDisabled and change all rule UUIDs to lowercase

jepio:
## Description

- the "EnsureAcceptingSourceRoutedPacketsIsDisabled" rule was checking `net.ipv4.conf.all.accept_source_route` instead of `net.ipv4.conf.default.accept_source_route`. Fix this.
- use lowercase rule ids for all rules in MOF files for consistency

## Checklist


src/adapters/mc/asb/AzureLinuxBaseline.mof
```diff
@@ -103,7 +103,7 @@ instance of OsConfigResource as $OsConfigResource5ref
 instance of OsConfigResource as $OsConfigResource6ref
 {
     ResourceID = "Ensure that the denied users for SSH are configured (CIS: L1 - Server - 5.2.18)";
-    RuleId = "1DB7D147-3A75-4637-8CDC-C119E22D0206";
+    RuleId = "1db7d147-3a75-4637-8cdc-c119e22d0206";
```
MariusNi, line 106: >"1db7d147-3a75-4637-8cdc-c119e22d0206"; [](http://example.com/codeflow?start=13&length=39)

This is actually by design. The uppercase UUIDs indicate the new rules that do not have metadata in MDC.

src/common/asb/Asb.c
```diff
@@ -1678,7 +1678,7 @@ static char* AuditEnsureSourceRoutedPacketsIsDisabled(void* log)
 static char* AuditEnsureAcceptingSourceRoutedPacketsIsDisabled(void* log)
 {
     char* reason = 0;
-    RETURN_REASON_IF_NOT_ZERO(CheckLineFoundNotCommentedOut("/proc/sys/net/ipv4/conf/all/accept_source_route", '#', "0", &reason, log));
+    RETURN_REASON_IF_NOT_ZERO(CheckLineFoundNotCommentedOut("/proc/sys/net/ipv4/conf/default/accept_source_route", '#', "0", &reason, log));
```
MariusNi, line 1681: We need both 'all' and 'default' for this rule. This is the code in main:

```
static char* AuditEnsureAcceptingSourceRoutedPacketsIsDisabled(void* log)
{
    char* reason = 0;
    RETURN_REASON_IF_NOT_ZERO(CheckLineFoundNotCommentedOut("/proc/sys/net/ipv4/conf/all/accept_source_route", '#', "0", &reason, log));
    CheckLineFoundNotCommentedOut("/proc/sys/net/ipv6/conf/default/accept_source_route", '#', "0", &reason, log);
    return reason;
}
````

It matches the ASB recipe and the remediation function in same file:

```
static int RemediateEnsureAcceptingSourceRoutedPacketsIsDisabled(char* value, void* log)
{
    UNUSED(value);
    return (SavePayloadToFile("/proc/sys/net/ipv4/conf/all/accept_source_route", "0", 1, log) &&
        SavePayloadToFile("/proc/sys/net/ipv6/conf/default/accept_source_route", "0", 1, log)) ? 0 : ENOENT;
}
```

Please let me know if I did not understood correctly what you are trying to do.

jepio: That is the bug - `net.ipv4.conf.all.accept_source_route` and `net.ipv6.conf.all.accept_source_route` are already checked by
`AuditEnsureSourceRoutedPacketsIsDisabled`/`RemediateEnsureSourceRoutedPacketsIsDisabled` (https://github.com/Azure/azure-osconfig/blob/main/src/common/asb/Asb.c#L1670-L1676, 
https://github.com/Azure/azure-osconfig/blob/main/src/common/asb/Asb.c#L3263-L3268).

`AuditEnsureAcceptingSourceRoutedPacketsIsDisabled`/`RemediateEnsureAcceptingSourceRoutedPacketsIsDisabled` should be checking `net.ipv4.conf.default.accept_source_route` and `net.ipv6.conf.default.accept_source_route`. These are the two functions I am correcting here.

src/adapters/mc/asb/AzureLinuxBaseline.mof
jepio: I see, but we will need to update the metadata for MDC for new rules.

It complicates consumers if we emit mixed lower and upper case uuids. Is there a strong reason not to unify this? We can discover rules not covered by MDC metadata again at any time by iterating and performing a lookup.

src/common/asb/Asb.c
MariusNi: Understood. As per these two entries at https://learn.microsoft.com/en-us/azure/governance/policy/samples/guest-configuration-baseline-linux:

<html>
<body>

The default setting for accepting source routed packets should be disabled for network interfaces. (net.ipv4.conf.default.accept_source_route = 0)(42.1) | Description: An attacker could redirect traffic for malicious purposes. | Run sysctl -w key=value and set to a compliant value.
-- | -- | --
The default setting for accepting source routed packets should be disabled for network interfaces. (net.ipv6.conf.default.accept_source_route = 0)(42.2) | Description: An attacker could redirect traffic for malicious purposes. | Run sysctl -w key=value and set to a compliant value.


</body>
</html>

MariusNi:
- Do we want bugs to be filled and triaged for both issues described in this PR? Or do we want to triage these two issues via this PR, without filled bugs?
- Can the title of the PR be expanded to describe what is changed, for example: `Fix ASB v2's EnsureSourceRoutedPacketsIsDisabled and change all rule UUIDs to lowercase`?

src/adapters/mc/asb/AzureLinuxBaseline.mof
MariusNi: You mean, consumers will see these UUIDs in the MDC UI? These UUIDs are not shown to consumers anywhere else as far as I know. Fine with me to correct all to lowercase, but not sure if this makes the bar for upcoming preview releases in January.

MariusNi (dismissed):
:shipit:

robertschaedler3 (approved):

MariusNi:
> const BASELINE_RULE g_rules[] = 

You also need to update this table here for the lowercase UUIDs. 

---
Refers to: src/common/asb/Asb.c:639 in 3531762. [](commit_id = 3531762dbfa668766dd34ff3f9403a1ceebb52a1, deletion_comment = False)

jepio:
> * Do we want bugs to be filled and triaged for both issues described in this PR? Or do we want to triage these two issues via this PR, without filled bugs?

Sure, will do tomorrow.

jepio:
> > const BASELINE_RULE g_rules[] =
> 
> Also need to update the lowercase UUIDs in the validation/reference table g_rules global table in Asb.c that is used to check that ids and names are not accidentally changed in MOF.

Thanks for the hint, this is great!

MariusNi (approved):
:shipit:

adteran:
It is not clear to me if these changes tied to issues with the audit/remediation functionality or are they good to have? Could you clarify which distros are impacted by the ipv4.* all vs default, and what is the impact when the issue presents?

---
In reply to: [2546585874](https://github.com/Azure/azure-osconfig/pull/835#issuecomment-2546585874) [](http://example.com/codeflow?ancestors=2546585874)

MariusNi:
> > * Do we want bugs to be filled and triaged for both issues described in this PR? Or do we want to triage these two issues via this PR, without filled bugs?
> 
> Sure, will do tomorrow.

We can also triage here. The 3 of us agree this is a good fix and well localized to only one rule, so even if not a P0/P1 it is worth including in preview. Since our manual validation is agile, we can as planned merge in stabilization fixes like this in, update the test policy, and continue to validate. We see results in a matter of hours. I will merge this and use tonight's build to update the test policy tomorrow and continue validating. Thanks for this work, Jeremi!

PR #846, opened 2025-01-23
New simple action to check policy package size

MariusNi:
## Description

New GitHub action to check policy package size added to existing Universal NRP workflow. Today our two packages are ~ 280 kB each. The current limit is 290 kB. We'll see how this goes in CI and adjust (increase) later if justified (for example, for the future CIS baseline).

## Checklist

### DIFF
--- a/.github/actions/check-size/action.yml
+++ b/.github/actions/check-size/action.yml
@@ -19,15 +19,22 @@ runs:
       run: |
         failed=false
 
+        echo "$package"
+        echo "$limit"
+        
         echo "Checking the size of '$package'"
         echo "Size of '$package':"
-        actual = wc -c $package | awk '{print $1}'
 
-        if (actual > limit)
-            echo "Size of '$package' of $actual kB is greater than the limit of $limit kB"
+        # command = f"wc -c $package | awk '{{print $1}}'"
+        # actual = int(subprocess.check_output(command, shell=True).strip())
+
+        actual = $(wc -c < "$package")
+
+        if [ "$actual" -gt "$limit" ]; then
+            echo "Size of '$package' of $actual bytes is greater than the limit of $limit bytes"
             failed=true
         else
-            echo "Size of '$package' of $actual kB is at or under the limit of $limit kB"
+            echo "Size of '$package' of $actual bytes is at or under the limit of $limit bytes"
         fi
 
         if [ "$failed" == true ]
--- a/.github/actions/check-size/action.yml
+++ b/.github/actions/check-size/action.yml
@@ -19,19 +19,13 @@ runs:
       run: |
         failed=false
 
-        echo "Package: $package"
-        echo "Limit: $limit kB"
-        
-        echo "Checking the size of '$package'"
-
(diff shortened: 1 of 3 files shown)


.github/workflows/universalnrp-test.yml
```diff
@@ -184,6 +184,13 @@ jobs:
               Rename-Item -Path $_.Name -NewName "$name.zip"
           }
       
+      - name: Check package size
+        if: success() || failure()
+        uses: ./.github/actions/check-size
+        with:
+          package: '${{ steps.normalize.outputs.path }}'
+          limit: "290"
```
robertschaedler3, line 192: Can `290` be a number?

robertschaedler3 (approved):

.github/actions/check-size/action.yml
```diff
@@ -0,0 +1,36 @@
...
+        actual=$(du -k "$package" | cut -f1)
+
+        if [ "$actual" -gt "$limit" ]; then
+            echo "Size of '$package' of $actual kB is greater than the limit of $limit kB"
+            failed=true
+        else
+            echo "Size of '$package' of $actual kB is at or under the limit of $limit kB"
+        fi
+
+        if [ "$failed" == true ]
```
AhmedBM, line 31: bad syntax here, should be `if [ "$failed" == true ]; then`

.github/workflows/universalnrp-test.yml
MariusNi: Assuming so. It works fine like this and takes a long time to validate and fix if any regression, so for now I am tempted to leave it like this (make it more obvious where the number starts and ends)

.github/actions/check-size/action.yml
MariusNi: Strangely, actually does work like this. I'll keep an eye on it and if bad, will fix.

.github/actions/check-size/action.yml
```diff
@@ -0,0 +1,36 @@
+name: Check size of a policy package
+description: Checks the disk size of a policy package
+
+inputs:
+  package:
+    description: the name of the package
+    required: true
+  limit:
```
AhmedBM, line 8: Can we make this a `type: integer` and pass an integer and not a string to the action?

MariusNi: See earlier answer to same. I guess, sure, but .... such slow runs.

.github/workflows/universalnrp-test.yml
MariusNi: Done!

.github/actions/check-size/action.yml
MariusNi: Done! (although it was working fine like this)

.github/actions/check-size/action.yml
MariusNi: Done!

AhmedBM (approved):
👌

robertschaedler3 (approved):

PR #856, opened 2025-02-04
Add ProcedureObjectName and ProcedureObjectValue to OsConfig Schema

wupeka:
## Description

Add ProcedureObjectName and ProcedureObjectValue fields to OsConfig schema and OsConfigResource
Add support for multiple Components in OsConfigResource.

## Checklist

### DIFF
--- a/src/adapters/mc/OsConfigResource.c
+++ b/src/adapters/mc/OsConfigResource.c
@@ -21,7 +21,9 @@ static char* g_ruleId = NULL;
 static char* g_payloadKey = NULL;
 static char* g_componentName = NULL;
 static char* g_initObjectName = NULL;
-static char* g_reportedObjectName = NULL;
+static char *g_procedureObjectName = NULL;
+static char *g_procedureObjectValue = NULL;
+static char *g_reportedObjectName = NULL;
 static char* g_expectedObjectValue = NULL;
 static char* g_desiredObjectName = NULL;
 static char* g_desiredObjectValue = NULL;
@@ -54,6 +56,8 @@ void __attribute__((constructor)) Initialize()
     g_payloadKey = DuplicateString(g_defaultValue);
     g_componentName = DuplicateString(g_defaultValue);
     g_initObjectName = DuplicateString(g_defaultValue);
+    g_procedureObjectName = DuplicateString(g_defaultValue);
+    g_procedureObjectValue = DuplicateString(g_defaultValue);
     g_reportedObjectName = DuplicateString(g_defaultValue);
     g_expectedObjectValue = DuplicateString(g_passValue);
     g_desiredObjectName = DuplicateString(g_defaultValue);
@@ -69,6 +73,8 @@ void __attribute__((destructor)) Destroy()
     FREE_MEMORY(g_payloadKey);
     FREE_MEMORY(g_componentName);
     FREE_MEMORY(g_initObjectName);
+    FREE_MEMORY(g_procedureObjectName);
+    FREE_MEMORY(g_procedureObjectValue);
     FREE_MEMORY(g_reportedObjectName);
     FREE_MEMORY(g_expectedObjectValue);
     FREE_MEMORY(g_desiredObjectName);
@@ -611,6 +617,25 @@ void MI_CALL OsConfigResource_Invoke_GetTargetResource(
         FREE_MEMORY(g_initObjectName);
     }
 
+    // Read the MIM procedure object name from the input resource values
+    if ((MI_TRUE == in->InputResource.value->ProcedureObjectName.exists) && (NULL != in->InputResource.value->ProcedureObjectName.value))
+    {
+        FREE_MEMORY(g_procedureObjectName);
+        if (NULL == (g_procedureObjectName = DuplicateString(in->InputResource.value->ProcedureObjectName.value)))
+        {
+            LogError(context, miResult, GetLog(), "[OsConfigResource.Get] DuplicateString(%s) failed", in->InputResource.value->ProcedureObjectName.value);
+            g_procedureObjectName = DuplicateString(g_defaultValue);
+            miResult = MI_RESULT_FAILED;
+            goto Exit;
+        }
+    }
+    else
+    {
+        // Not an error
+        LogInfo(context, GetLog(), "[OsConfigResource.Get] No ProcedureObjectName");
+        FREE_MEMORY(g_procedureObjectName);
+    }
+
     // Read the MIM reported object name from the input resource values
     if ((MI_TRUE == in->InputResource.value->ReportedObjectName.exists) && (NULL != in->InputResource.value->ReportedObjectName.value))
     {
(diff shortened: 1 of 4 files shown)


wupeka:
@microsoft-github-policy-service agree company="Microsoft"

src/adapters/mc/OsConfigResource.c
```diff
@@ -328,12 +366,132 @@ static MI_Result SetDesiredObjectValueToDevice(const char* who, char* objectName
...
+            for (component = g_components; NULL != component->Name; ++component)
+            {
+                if (!strcmp(component->Name, g_componentName))
+                {
+                    break;
+                }
+            }
+            if (NULL != component->MmiSet)
+            {
+                mpiResult = component->MmiSet(g_componentName, objectName, payloadString, payloadSize, GetLog());
```
MariusNi, line 383: >component->MmiSet( [](http://example.com/codeflow?start=28&length=18)

Adapters MUST NOT call MMI directly. Adapters can only call the MPI. In this case, the Universal NRP adapter can call OSConfig over the MPI REST API.

src/adapters/mc/Common.h
```diff
@@ -52,4 +51,23 @@
...
+void osConfigRegisterComponent(const OsConfigComponent* component);
+
+#define OSCONFIG_REGISTER_COMPONENT(name, initialize, shutdown, get, set, validresource) static void __attribute__((constructor)) __osconfig_register_component_##name() \
+                                                                                                {\
+                                                                                                  static const OsConfigComponent component = {#name, initialize, shutdown, get, set, validresource};\
+                                                                                                  osConfigRegisterComponent(&component);\
+                                                                                                }
+
+
 #endif
```
MariusNi, line 73: This PR contains a lot more changes than adding the two new (proposed but not yet agreed on) Universal Resource Class Parameters.

MariusNi (changes requested):
:clock1:

src/adapters/mc/OsConfigResource.c
```diff
@@ -378,6 +384,108 @@ static MI_Result SetDesiredObjectValueToDevice(const char* who, char* objectName
...
+        miResult = MI_RESULT_FAILED;
+        mpiResult = ENOMEM;
+        LogError(context, miResult, GetLog(), "[%s] json_serialize_to_string('%s') failed", who, g_procedureObjectValue);
+    }
+    else
+    {
+        payloadSize = (int)strlen(serializedValue);
+        if (NULL != (payloadString = malloc(payloadSize + 1)))
+        {
+            const OsConfigComponent *component = NULL;
```
MariusNi, line 421: >const OsConfigComponent *component  [](http://example.com/codeflow?start=12&length=35)

Pure C style requires all variables locally declared withing a statement of a function to be defined at the beginning/top of the statement. To respect this old style and for consistency with existing code, can you please move the definition on line 421 at top, between lines 389 and 396? Please repeat the same for all other locals that would fall into this case.

MariusNi, line 421: >OsConfigComponent [](http://example.com/codeflow?start=18&length=17)

Where is OsConfigComponent defined?

src/adapters/mc/OsConfigResource.c
```diff
@@ -378,6 +384,108 @@ static MI_Result SetDesiredObjectValueToDevice(const char* who, char* objectName
...
+                    break;
+                }
+            }
+            if (NULL != component->MmiSet)
+            {
+                mpiResult = component->MmiSet(g_componentName, objectName, payloadString, payloadSize, GetLog());
+                LogInfo(context, GetLog(), "[%s] MmiSet(%s, %s, '%.*s', %d) returned %d",
+                    who, g_componentName, objectName, payloadSize, payloadString, payloadSize, mpiResult);
+
+            }
```
MariusNi, line 439: Please do not do this. This is the role of the Modules Manager. Adapters cannot call MMI directly. Adapters can only call the MPI. 

Please remove this entire function and keep the NRP existing functions for get and set that talk to OSConfig over the MPI REST API.

MariusNi:
>             mpiResult = AsbMmiSet(g_componentName, objectName, payloadString, payloadSize, GetLog());

Instead of the new function SetProcedureObjectValueToDevice (that duplicates a large chunk of code) why not add a switch here in SetDesiredObjectValueToDevice (which is not for just one paramater, is for any desired MIM object):

 if (0 == strncmp(objectName, "procedure", strlen(procedure)) then call ComplianceMmiSet(), otherwise call AsbMmiSet()

---
Refers to: src/adapters/mc/OsConfigResource.c:340 in 9237261. [](commit_id = 9237261d8c4dc0e663a2defd53cc98a08cd607a5, deletion_comment = False)

src/adapters/mc/OsConfigResource.c
```diff
@@ -378,6 +384,108 @@ static MI_Result SetDesiredObjectValueToDevice(const char* who, char* objectName
...
+            for (component = g_components; NULL != component->Name; ++component)
+            {
+                if (!strcmp(component->Name, g_componentName))
+                {
+                    break;
+                }
+            }
+            if (NULL != component->MmiSet)
+            {
+                mpiResult = component->MmiSet(g_componentName, objectName, payloadString, payloadSize, GetLog());
```
MariusNi, line 435: >mpiResult = component->MmiSet [](http://example.com/codeflow?start=16&length=29)

Instead of this, let's call the new ComplianceMmiSet in SetDesiredObjectValueToDevice and then add the new static library Compliance under common, with the header and empty body functions, just to allow us build the NRP with it and start testing:

compliance.h:

#ifdef __cplusplus
extern "C"
{
#endif

void ComplianceInitialize(void* log);
void ComplianceShutdown(void* log);

int ComplianceMmiGet(const char* componentName, const char* objectName, char** payload, int* payloadSizeBytes, unsigned int maxPayloadSizeBytes, void* log);
int ComplianceMmiSet(const char* componentName, const char* objectName, const char* payload, const int payloadSizeBytes, void* log);

#ifdef __cplusplus
}
#endif

and then compliamce.c (or compliamce.cpp) with the functions, just enough so the code will compile and run w/o yet invoking the actual module functionality)

MariusNi (approved):
:shipit:

robertschaedler3 (approved):

src/adapters/mc/OsConfigResource.c
```diff
@@ -656,6 +681,17 @@ void MI_CALL OsConfigResource_Invoke_GetTargetResource(
...
     }
 
+    // Read the procedure MIM object value from the input resource values
+    if ((in->InputResource.value->ProcedureObjectValue.exists == MI_TRUE) && (in->InputResource.value->ProcedureObjectValue.value != NULL))
+    {
+        FREE_MEMORY(g_procedureObjectValue);
+        if (NULL == (g_procedureObjectValue = DuplicateString(in->InputResource.value->ProcedureObjectValue.value)))
+        {
+            LogError(context, miResult, GetLog(), "[OsConfigResource.Get] DuplicateString(%s) failed", in->InputResource.value->ProcedureObjectValue.value);
+            g_procedureObjectValue = DuplicateString(g_defaultValue);
```
simonjaeger, line 691: Is no `goto Exit;` here intentional?

simonjaeger (approved):

src/adapters/mc/OsConfigResource.c
MariusNi: Correct. The new 'procedure' parameters are optional, they may not always exist in the MOF, and we must not fail when not.

PR #857, opened 2025-02-06
Completing adding the two 'procedure' resource class parameters for 'OSConfig for MC'

MariusNi:
## Description

Changes in this PR include:
- Further completing the work started to add support for the two new resource class optional parameters.
- Creating a simple way to swap the current baseline library to link to the NRP, with a common interface. 
- Each new policy package will have a new sub-folder under /adapters/mc and there add implementation for their common baseline interface (from adapters/mc/Baseline.h) and in their CMakeList.txt link to their own static libraries. 
- In particular, SSH and ASB have same asb library and interface implementation in their local Baseline.c. Compliance can add new such interface implementation C or C++ files under each new directory created under adapters/mc (and avoid linking to the asb library).

## Checklist


simonjaeger (approved):

wupeka (changes requested):

src/adapters/mc/OsConfigResource.c
```diff
@@ -300,7 +301,7 @@ void MI_CALL OsConfigResource_DeleteInstance(
     MI_Context_PostResult(context, MI_RESULT_NOT_SUPPORTED);
 }
 
-static MI_Result SetDesiredObjectValueToDevice(const char* who, char* objectName, MI_Context* context)
+static MI_Result SetDesiredObjectValueToDevice(const char* who, char* componentName, char* objectName, char* objectValue, MI_Context* context)
```
robertwoj-microsoft, line 304: Can we make componentName and objectName const char*?

src/adapters/mc/OsConfigResource.c
```diff
@@ -628,6 +636,32 @@ void MI_CALL OsConfigResource_Invoke_GetTargetResource(
...
+        else
+        {
+            // We have a procedure object name, next see if we also have a procedure object value (also optional)
+            if ((MI_TRUE == in->InputResource.value->ProcedureObjectValue.exists) && (NULL != in->InputResource.value->ProcedureObjectValue.value))
+            {
+                FREE_MEMORY(g_procedureObjectValue);
+                if (NULL == (g_procedureObjectValue = DuplicateString(in->InputResource.value->ProcedureObjectValue.value)))
+                {
+                    LogError(context, miResult, GetLog(), "[OsConfigResource.Get] DuplicateString(%s) failed", in->InputResource.value->ProcedureObjectValue.value);
+                    g_procedureObjectValue = DuplicateString(g_defaultValue);
```
robertwoj-microsoft, line 648: This is a more general concern: In case we fail to allocate memory, we try to allocate another block. As this is a repeated pattern, please just think about it. It is very likely the second allocation would fail as well, and we should not try to allocate anything in in such cases.

src/adapters/mc/OsConfigResource.c
```diff
@@ -1108,19 +1136,53 @@ void MI_CALL OsConfigResource_Invoke_TestTargetResource(
...
+                {
+                    LogError(context, miResult, GetLog(), "[OsConfigResource.Test] DuplicateString(%s) failed", in->InputResource.value->ProcedureObjectValue.value);
+                    g_procedureObjectValue = DuplicateString(g_defaultValue);
+                    miResult = MI_RESULT_FAILED;
+                    goto Exit;
+                }
+                else
+                {
+                    // We have both a procedure object name and value, we need to set then now to apply context for commpliance
+                    SetDesiredObjectValueToDevice("OsConfigResource.Test", g_componentName, g_procedureObjectName, g_procedureObjectValue, context);
```
robertwoj-microsoft, line 1166: Similar comment to previous: we're not handing error here.

src/adapters/mc/OsConfigResource.c
```diff
@@ -1108,19 +1136,53 @@ void MI_CALL OsConfigResource_Invoke_TestTargetResource(
...
+                }
+            }
+            else
+            {
+                // Not an error
```
robertwoj-microsoft, line 1171: Similar comment to previous: I think it is an error

src/adapters/mc/OsConfigResource.c
```diff
@@ -1308,7 +1370,52 @@ void MI_CALL OsConfigResource_Invoke_SetTargetResource(
...
+                {
+                    LogError(context, miResult, GetLog(), "[OsConfigResource.Set] DuplicateString(%s) failed", in->InputResource.value->ProcedureObjectValue.value);
+                    g_procedureObjectValue = DuplicateString(g_defaultValue);
+                    miResult = MI_RESULT_FAILED;
+                    goto Exit;
+                }
+                else
+                {
+                    // We have both a procedure object name and value, we need to set then now to apply context for commpliance
+                    SetDesiredObjectValueToDevice("OsConfigResource.Set", g_componentName, g_procedureObjectName, g_procedureObjectValue, context);
```
robertwoj-microsoft, line 1400: Similar: we're not handling error here

src/adapters/mc/OsConfigResource.c
```diff
@@ -1308,7 +1370,52 @@ void MI_CALL OsConfigResource_Invoke_SetTargetResource(
...
+                }
+            }
+            else
+            {
+                // Not an error
```
robertwoj-microsoft, line 1405: Similar to above: I think this is an error.

src/adapters/mc/OsConfigResource.c
```diff
@@ -1308,7 +1370,52 @@ void MI_CALL OsConfigResource_Invoke_SetTargetResource(
...
+        }
+    }
+    else
+    {
+        // Not an error
+        LogInfo(context, GetLog(), "[OsConfigResource.Test] No ProcedureObjectName");
+        FREE_MEMORY(g_procedureObjectName);
+    }
+
+    SetDesiredObjectValueToDevice("OsConfigResource.Set", g_componentName, g_desiredObjectName, g_desiredObjectValue, context);
```
robertwoj-microsoft, line 1418: Similar comment: we're not handling error here.

src/modules/test/recipes/SecurityBaselineTests.json
```diff
@@ -1216,16 +1216,6 @@
     "ObjectName": "remediateEnsureUnnecessaryAccountsAreRemoved",
     "Payload": "games,test"
   },
-  {
```
robertwoj-microsoft, line 1219: Please separate this changes into different PR, this is unrelated.

robertwoj-microsoft (changes requested):

src/modules/test/recipes/SecurityBaselineTests.json
jepio: We don't want these tests to be removed - but they are the ones that most often collide with an internal tool that runs on the worker nodes. I'm guessing that's why @MariusNi disabled them temporarily.

src/adapters/mc/OsConfigResource.c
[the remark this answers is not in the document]
jepio: This feels like a step in the right direction - with this change only the MC entrypoints (Get/Set/Test) manipulate the global and all other functions take the component as a parameter.

src/adapters/mc/OsConfigResource.c
```diff
@@ -628,6 +636,32 @@ void MI_CALL OsConfigResource_Invoke_GetTargetResource(
...
+                if (NULL == (g_procedureObjectValue = DuplicateString(in->InputResource.value->ProcedureObjectValue.value)))
+                {
+                    LogError(context, miResult, GetLog(), "[OsConfigResource.Get] DuplicateString(%s) failed", in->InputResource.value->ProcedureObjectValue.value);
+                    g_procedureObjectValue = DuplicateString(g_defaultValue);
+                    miResult = MI_RESULT_FAILED;
+                    goto Exit;
+                }
+                else
+                {
+                    // We have both a procedure object name and value, we need to set then now to apply context for commpliance
```
jepio, line 654:
```suggestion
                    // We have both a procedure object name and value, we need to set then now to apply context for compliance
```

src/modules/test/recipes/SecurityBaselineTests.json
robertwoj-microsoft: Yeah, but this is just a thing that should be put in a different PR is what I'm saying.

src/adapters/mc/OsConfigResource.c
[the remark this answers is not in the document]
robertwoj-microsoft: I agree that this is a good direction, but this function is still internal to the OsConfigResource. I think the intention is to actually move forward avoid the global variable, if possible, but as it is now it doesn't change much.

src/adapters/mc/OsConfigResource.c
```diff
@@ -816,6 +839,30 @@ void MI_CALL OsConfigResource_Invoke_GetTargetResource(
         }
     }
 
+    // Write the desired procedure MIM object name to the output resource values if present in input resource values
+    if ((in->InputResource.value->ProcedureObjectName.exists == MI_TRUE) && (NULL != in->InputResource.value->ProcedureObjectName.value))
+    {
+        memset(&miValue, 0, sizeof(miValue));
+        miValue.string = (MI_Char*)(g_procedureObjectName);
+        if (MI_RESULT_OK != (miResult = MI_Instance_SetElement(resultResourceObject, MI_T("ProcedureObjecttName"), &miValue, MI_STRING, 0)))
```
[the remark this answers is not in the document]
MariusNi: Godo catch, fixed.

src/adapters/mc/OsConfigResource.c
```diff
@@ -311,13 +312,13 @@ static MI_Result SetDesiredObjectValueToDevice(const char* who, char* objectName
     MI_Result miResult = MI_RESULT_OK;
     int mpiResult = MPI_OK;
 
-    if ((NULL == objectName) || (NULL == g_desiredObjectValue))
+    if ((NULL == componentName) || (NULL == objectName) || (NULL == objectValue))
```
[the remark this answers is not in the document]
MariusNi: 'who' is only used in logging. If it were null, the log trace would show '<null>' (no crash).

src/adapters/mc/OsConfigResource.c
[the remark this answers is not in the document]
MariusNi: The component name was added when I was trying Wytek's solution, before finding it too much for what we need here. They remain for now, does not change anything but could become useful in the future.

src/adapters/mc/OsConfigResource.c
```diff
@@ -628,6 +636,32 @@ void MI_CALL OsConfigResource_Invoke_GetTargetResource(
...
+                }
+                else
+                {
+                    // We have both a procedure object name and value, we need to set then now to apply context for commpliance
+                    SetDesiredObjectValueToDevice("OsConfigResource.Get", g_componentName, g_procedureObjectName, g_procedureObjectValue, context);
+                }
+            }
+            else
+            {
+                // Not an error
```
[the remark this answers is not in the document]
MariusNi: Adding this

src/adapters/mc/OsConfigResource.c
```diff
@@ -628,6 +636,32 @@ void MI_CALL OsConfigResource_Invoke_GetTargetResource(
...
+                {
+                    LogError(context, miResult, GetLog(), "[OsConfigResource.Get] DuplicateString(%s) failed", in->InputResource.value->ProcedureObjectValue.value);
+                    g_procedureObjectValue = DuplicateString(g_defaultValue);
+                    miResult = MI_RESULT_FAILED;
+                    goto Exit;
+                }
+                else
+                {
+                    // We have both a procedure object name and value, we need to set then now to apply context for commpliance
+                    SetDesiredObjectValueToDevice("OsConfigResource.Get", g_componentName, g_procedureObjectName, g_procedureObjectValue, context);
```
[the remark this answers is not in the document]
MariusNi: Yes. AzPolicy does not surface remediation results, only audit results. If a remediation does not work, audit fails and then more information (beyond what's in the reason phrase) can be found in the logs. We do not want to fail the NRP interface methods Test/Get/Set unless absolutely necessary - the experience is very bad for customer, the entire policy audit fails. Is much better to allow each rule audit to fail gracefully, with a meaningful reason.

src/adapters/mc/OsConfigResource.c
MariusNi: Adding this. Will treat this way because this signals a bad MOF so is ok to fail the entire audit.

src/adapters/mc/OsConfigResource.c
MariusNi: Same as explained earlier. Thinking that we must fail here for any error is a trap. We only want to fail when there is no way we can work anymore. MOF problems are OK to fail on, so are memory allocation failures. MPI/MMI errors do not fall into this category.

src/modules/test/recipes/SecurityBaselineTests.json
MariusNi: It;'s just this one auditd package, is random and hits our CI 9 our of 10 runs sort off. Let's keep them off Modules Test until we find a solution. Will move to another PR.

src/adapters/mc/OsConfigResource.c
MariusNi: As explained in the other instance.

src/adapters/mc/OsConfigResource.c
MariusNi: Godo catch, fixed

src/adapters/mc/OsConfigResource.c
MariusNi: See the explanation why.

src/adapters/mc/OsConfigResource.c
MariusNi: I agree. A change for these may come in another PR.

robwoj (approved):

robwoj (approved):

src/modules/test/recipes/SecurityBaselineTests.json
```diff
@@ -2403,6 +2403,11 @@
     "ComponentName": "SecurityBaseline",
     "ObjectName": "auditEnsureSnmpServerIsDisabled"
   },
+  {
+    "ObjectType": "Reported",
+    "ComponentName": "SecurityBaseline",
+    "ObjectName": "auditEnsureSnmpServerIsDisabled"
```
MariusNi, line 2409: Duplicate, will remove.

robertwoj-microsoft (approved):

wupeka (approved):

PR #861, opened 2025-02-10
Import Rule Evaluation engine

robertwoj-microsoft:
## Description

This PR imports the Rule Evaluation engine into the project. Part of the POC which maintains evaluation of Boolean logic and execution of built-in functions.
Evaluation engine implementation done by @wupeka. 

## Checklist


src/modules/compliance/src/lib/Evaluator.cpp
```diff
@@ -0,0 +1,256 @@
...
+
+#include <errno.h>
+#include <sys/stat.h>
+#include <unistd.h>
+#include <stdarg.h>
+#include <ctype.h>
+#include <parson.h>
+#include <CommonUtils.h>
+#include <UserUtils.h>
+#include <SshUtils.h>
```
MariusNi, line 17: Why is this necessary to be included here?

src/modules/compliance/src/lib/Evaluator.cpp
```diff
@@ -0,0 +1,256 @@
...
+
+Result<bool> Evaluator::ExecuteAudit(char** payload, int* payloadSizeBytes)
+{
+    if (mJson == nullptr)
+    {
+        OsConfigLogError(mLog, "Invalid argument: rule is null");
+        return Error("Rule is null");
+    }
+
+    if(nullptr == payload || nullptr == payloadSizeBytes)
```
MariusNi, line 34: >if(nullptr == payload || nullptr == payloadSizeBytes) [](http://example.com/codeflow?start=4&length=53)

One general rule that we try to respect in this project is to not rely on just operator precedence. Always add parenthesis.  For example, here:

'
if((nullptr == payload) || (nullptr == payloadSizeBytes))
'

Also for this particular line here, may also want to check for *payloadSizeBytes being <= 0.

src/modules/compliance/src/lib/Evaluator.cpp
```diff
@@ -0,0 +1,256 @@
...
+    }
+
+    auto result = EvaluateProcedure(mJson, Action::Audit);
+    if (!result.has_value())
+    {
+        OsConfigLogError(mLog, "Evaluation failed: %s", result.error().message.c_str());
+        return result.error();
+    }
+
+    std::string vlog = mLogstream.str().substr(0, 4096 - (1 + 4 + 2)); // 4 for "PASS" or "FAIL", 2 for quotes
```
MariusNi, line 47: >1 + 4 + 2 [](http://example.com/codeflow?start=58&length=9)

Try to not use magic numbers. Instead, define "PASS" and "FAIL" as macros (or use existing ones if any) and then do sizeof() or strlen() (single byte chars so either will work). Then the comment won't be necessary.

src/modules/compliance/src/lib/Evaluator.cpp
```diff
@@ -0,0 +1,256 @@
...
+    }
+
+    if(nullptr == payload || nullptr == payloadSizeBytes)
+    {
+        OsConfigLogError(mLog, "Invalid argument: payload or payloadSizeBytes is null");
+        return Error("Payload or payloadSizeBytes is null");
+    }
+
+    auto result = EvaluateProcedure(mJson, Action::Audit);
+    if (!result.has_value())
```
MariusNi, line 41: > (!result.has_value()) [](http://example.com/codeflow?start=6&length=22)

nit: always try to be explicit. Here, is the result boolean or integer? If boolean, is false defined as 0?

Instead, can we do like this here:

'
if (false == result.has_value())
'

?

src/modules/compliance/src/lib/Evaluator.cpp
MariusNi, line 47: Define a constant for this (for 4096)

src/modules/compliance/src/lib/Evaluator.cpp
```diff
@@ -0,0 +1,256 @@
...
+}
+
+void Evaluator::setProcedureMap(std::map<std::string, std::pair<action_func_t, action_func_t>> procedureMap)
+{
+    mProcedureMap = std::move(procedureMap);
+}
+
+Result<bool> Evaluator::EvaluateProcedure(const JSON_Object* json, const Action action)
+{
+    if(nullptr == json)
```
MariusNi, line 73: > if(nullptr == json) [](http://example.com/codeflow?start=3&length=20)

One space between 'if' and '('. Here and on line 79, and everywhere else.

src/modules/compliance/src/lib/Evaluator.cpp
```diff
@@ -0,0 +1,256 @@
...
+            return Error("anyOf value is not an array");
+        }
+        JSON_Array *array = json_value_get_array(value);
+        size_t count = json_array_get_count(array);
+        mLogstream << "{ anyOf: [";
+        for (size_t i = 0; i < count; ++i)
+        {
+            JSON_Object *subObject = json_array_get_object(array, i);
+            auto result = EvaluateProcedure(subObject, action);
+            if (!result.has_value()) {
```
MariusNi, line 104: > { [](http://example.com/codeflow?start=36&length=2)

Please break '{' to the next line. Here and elsewhere

src/modules/compliance/src/lib/Optional.h
```diff
@@ -0,0 +1,135 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#ifndef COMPLIANCE_OPTIONAL_H
+#define COMPLIANCE_OPTIONAL_H
+
+#include "TypeTraits.h"
+#include <memory>
```
MariusNi, line 8: Add a precompiled header to contain all includes such as these  and include just that everywhere?

src/tests/fuzzer/target.cpp
```diff
@@ -981,6 +1108,8 @@ static const std::map<std::string, int (*)(const char*, std::size_t)> g_targets
     { "GetGitBranchFromJsonConfig.", GetGitBranchFromJsonConfig_target },
     { "CheckOrEnsureUsersDontHaveDotFiles.", CheckOrEnsureUsersDontHaveDotFiles_target },
     { "CheckUserAccountsNotFound.", CheckUserAccountsNotFound_target },
+    { "ComplianceEvaluateAudit.", ComplianceEvaluateAudit_target },
+    { "ComplianceEvaluateRemediation.", ComplianceEvaluateRemediation_target },
```
MariusNi, line 1112: >, [](http://example.com/codeflow?start=78&length=1)

bit: ',' not necessary on last list element

src/modules/compliance/src/lib/Evaluator.cpp
wupeka: In C++ the condition in if statement is converted to bool, not compared to 0 as in C, so since has_value() returns a bool this is an explicit statement.

src/modules/compliance/src/lib/Evaluator.cpp
```diff
@@ -0,0 +1,256 @@
...
+    const char* name = json_object_get_name(json, 0);
+    if(nullptr == name)
+    {
+        return Error("Rule name is null");
+    }
+
+    JSON_Value* value = json_object_get_value_at(json, 0);
+    if(nullptr == value)
+    {
+        return Error("Rule value is null");
```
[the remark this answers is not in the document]
robertwoj-microsoft: This code does not log, it handles error and returns it to the caller.

src/modules/compliance/src/lib/JSON.cpp
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include "JSON.h"
```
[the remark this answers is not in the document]
robertwoj-microsoft: It defines a JSON type, so the file name follows. It does not contain any more utilities.

robertwoj-microsoft:
>Can we stick with the globals we have set for the entire project? The root CMakeList.txt has these:
It should follow the setting from the main CMakeLists.txt, but I'll verify that.

src/modules/compliance/src/lib/Evaluator.cpp
robertwoj-microsoft: I'll verify, probably leftover.

src/modules/compliance/src/lib/Optional.h
robertwoj-microsoft: I don't think using a precompiled header would introduce much benefit since these header files contain mainly templates which won't be optimized unless we provide explicit instantiations.

src/adapters/mc/compliance/Baseline.c
```diff
@@ -0,0 +1,68 @@
...
+#include "../Common.h"
+#include <ComplianceInterface.hpp>
+
+static MMI_HANDLE gCompliance = NULL;
+
+static const char gComponentName[] = "Compliance";
+
+int BaselineIsValidResourceIdRuleId(const char* resourceId, const char* ruleId, const char* payloadKey, void* log)
+{
+    (void)resourceId;
```
MariusNi, line 13: >(void)resourceId; [](http://example.com/codeflow?start=4&length=17)

Please use the UNUSED() macro instead (here and elsewhere).

src/modules/compliance/src/lib/Evaluator.cpp
MariusNi: In general, let's please follow this rule to try to be explicit for code clarity. In this instance, it's fine to keep as-is. 

My comment on Boolean was that Boolean values may have different defined values depending on platform. On Windows at least you do not want to assume false/FALSE/False is always 0, for example.

src/modules/compliance/src/lib/Evaluator.cpp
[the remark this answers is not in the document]
MariusNi: Consider logging to capture errors with context, from where they originate. Even for tools like this, could be useful.

src/modules/compliance/src/lib/JSON.cpp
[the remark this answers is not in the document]
MariusNi: Then call it 'JsonType'? Even if keeping the current name, please use correct casing for it (Json.h, JsonDeleter, etc)

src/modules/compliance/src/lib/Optional.h
MariusNi: Main benefit is keeping includes clean and ordered. Up to you.

src/modules/compliance/src/lib/Result.h
```diff
@@ -0,0 +1,249 @@
...
+        Result& operator=(Result&& other) noexcept
+        {
+            if (this == &other)
+            {
+                return *this;
+            }
+
+            if (mTag == Tag::Value)
+            {
+                delete mPointer.value;
```
MariusNi, line 155: > delete mPointer.value; [](http://example.com/codeflow?start=15&length=23)

Could mPointer.value be nullptr? Or can it be not nullptr but pointing to freed memory? How does this code prevent against such conditions. Here and elsewhere.

src/modules/compliance/src/lib/Evaluator.cpp
```diff
@@ -0,0 +1,249 @@
...
+            }
+        }
+        else
+        {
+            fn = procedure->second.first;
+        }
+        if (nullptr == fn)
+        {
+            mLogstream << "ERROR: Function not found";
+            return Error("Function not found");
```
MariusNi, line 227: Please add a log like all other OSConfig modules have, using the common logging lib, that will log to:

/var/log/osconfig_compliance.log

And then trace everywhere it makes sense and log all error cases. In OSConfig we try to have complete logs, such that debugging is rarely or never necessary in order to investigate a problem (also in most cases our customers/partners could not give us debug remotes, and we may also be unable to get the repro in house).

src/modules/compliance/src/lib/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+project(compliancelib)
+
+add_library(compliancelib STATIC
```
MariusNi, line 6: >dd_library(compliancelib STATIC [](http://example.com/codeflow?start=1&length=31)

Possible to break this PR in separate parts for the Rule Evaluation engine, the Compliance module library, the Compliance module tests, etc?

As it is is right now this PR is quite large and this makes it more difficult to review.

src/modules/compliance/src/lib/Evaluator.cpp
robertwoj-microsoft: Yes, but it does not apply to C++ - as Witek said, bool is well-defined here. Obviously the more clarity the better, so in case there may be some implicit conversions, we should follow explicit comparisons, so I think we're on the same page here.

src/modules/compliance/src/lib/JSON.cpp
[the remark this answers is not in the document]
robertwoj-microsoft: I keep JSON as it is an abbreviation, but OK, I'll rename to be cased and append Type.

src/modules/compliance/src/lib/Result.h
robertwoj-microsoft: In general you can delete a null pointer, same as in C. So it may be null, but then it's not causing the code to crash. It can happen to be null for example if it's a moved-from object (see move constructor).

src/tests/fuzzer/target.cpp
jepio: I would suggest we keep it though - that way we don't need to modify two lines when adding an extra entry to the map (one for the new entry and one adding a comma after the last entry).
So we keep the diffs more readable.

src/modules/compliance/src/lib/Evaluator.cpp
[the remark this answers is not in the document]
robertwoj-microsoft: Added additional logging as discussed.

src/modules/compliance/src/lib/JSON.cpp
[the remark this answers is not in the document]
robertwoj-microsoft: Renamed to JsonWrapper which better describes the purpose as it simply is a unique_ptr with custom deleter.

src/modules/compliance/src/lib/Optional.h
robertwoj-microsoft: Ignoring for now. Will get back to putting common headers into a single one when other parts of the POC are merged and see what's needed then.

src/adapters/mc/compliance/Baseline.c
robertwoj-microsoft: Done

src/modules/compliance/src/lib/Evaluator.cpp
robertwoj-microsoft: Added additional logging as discussed.

src/modules/compliance/src/lib/CMakeLists.txt
robertwoj-microsoft: Keeping as is - rebased with dev and should contain less changes than previously.

src/modules/compliance/src/lib/Evaluator.h
```diff
@@ -0,0 +1,54 @@
...
+        Evaluator(const struct json_object_t* json, const std::map<std::string, std::string>& parameters, OSCONFIG_LOG_HANDLE log) : mJson(json), mParameters(parameters), mLog(log) {}
+        Evaluator(const Evaluator&) = delete;
+        Evaluator(Evaluator&&) = delete;
+        Evaluator& operator=(const Evaluator&) = delete;
+        Evaluator& operator=(Evaluator&&) = delete;
+
+        Result<bool> ExecuteAudit(char** payload, int* payloadSizeBytes);
+        Result<bool> ExecuteRemediation();
+
+        void setProcedureMap(std::map<std::string, std::pair<action_func_t, action_func_t>> procedureMap);
```
kkanas, line 36: Maybe typedef helper for std::map<std::string, std::pair<action_func_t, action_func_t>> ? 
Or even maybe using - seems to be available in c++11 https://en.cppreference.com/w/cpp/language/type_alias

src/modules/compliance/src/lib/Evaluator.cpp
```diff
@@ -0,0 +1,259 @@
...
+    else
+    {
+        if (json_value_get_type(value) != JSONObject)
+        {
+            mLogstream << "ERROR: value is not an object";
+            OsConfigLogError(mLog, "value is not an object");
+            return Error("value is not an object");
+        }
+        std::map<std::string, std::string> arguments;
+        auto args_object = json_value_get_object(value);
```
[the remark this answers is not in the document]
robertwoj-microsoft: No, parson requires you to free the top-level object.

MariusNi (approved):
:shipit:

PR #863, opened 2025-02-11
Import Engine class

robertwoj-microsoft:
## Description

Import the Engine class which implements mmiGet and mmiSet functions of the compliance module.
This PR includes work done by @wupeka.

## Checklist

### DIFF
--- a/src/common/logging/Logging.h
+++ b/src/common/logging/Logging.h
@@ -5,7 +5,6 @@
 #define LOGGING_H
 
 #include <string.h>
-#include <stdio.h>
 
 // Maximum log size (1,024,000 is 1MB aka 1024 * 1000), increase or decrease as needed
 #define MAX_LOG_SIZE 1024000
--- a/src/modules/compliance/src/lib/Base64.cpp
+++ b/src/modules/compliance/src/lib/Base64.cpp
@@ -1,0 +1,96 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include <string>
+#include "Result.h"
+#include "Base64.h"
+
+namespace compliance
+{
+    static inline char Base64Char(const unsigned char c)
+    {
+        if (c >= 'A' && c <= 'Z')
+        {
+            return c - 'A';
+        }
+        else if (c >= 'a' && c <= 'z')
+        {
+            return c - 'a' + 26;
+        }
+        else if (c >= '0' && c <= '9')
+        {
+            return c - '0' + 52;
+        }
+        else if (c == '+')
+        {
+            return 62;
+        }
+        else if (c == '/')
+        {
+            return 63;
+        }
+        else
+        {
+            return 0;
+        }
+    }
+
+    static inline bool IsBase64(const unsigned char c)
+    {
+        return (isalnum(c) || (c == '+') || (c == '/') || (c == '='));
+    }
+
+    Result<std::string> Base64Decode(const std::string &input)
+    {
+        if (input.size() % 4 != 0)
+        {
+            return Error("Invalid base64 length", EINVAL);
+        }
+        for (const char c : input)
+        {
+            if (!IsBase64(c))
+            {
+                return Error("Invalid base64 character", EINVAL);
+            }
+        }
+
+        std::string ret;
+        ret.reserve((input.size() * 3) / 4);
+
+        for (size_t i = 0; i < input.size(); i += 4)
+        {
+            unsigned char enc[4];
+
+            int j = 0;
+            for (j = 0; j < 4; j++)
+            {
+                if (input[i+j] == '=')
+                {
+                    break;
+                }
+                enc[j] = Base64Char(input[i + j]);
+            }
+
+            if (j == 4)
+            {
+                ret += (enc[0] << 2) | (enc[1] >> 4);
+                ret += ((enc[1] & 0x0f) << 4) | (enc[2] >> 2);
+                ret += ((enc[2] & 0x03) << 6) | enc[3];
+            }
+            else if (j == 3)
+            {
+                ret += (enc[0] << 2) | (enc[1] >> 4);
+                ret += ((enc[1] & 0x0f) << 4) | (enc[2] >> 2);
+            }
+            else if (j == 2)
+            {
+                ret += (enc[0] << 2) | (enc[1] >> 4);
+            }
+            else
+            {
+                return Error("Invalid base64", EINVAL);
+            }
+        }
+        return ret;
+    }
+}
--- a/src/modules/compliance/src/lib/CMakeLists.txt
+++ b/src/modules/compliance/src/lib/CMakeLists.txt
@@ -8,23 +8,20 @@ add_library(compliancelib STATIC
     Evaluator.cpp
     ProcedureMap.cpp
     JsonWrapper.cpp
+    Base64.cpp
+    Engine.cpp
+    Procedure.cpp
     ${PROCEDURES}
     )
-
-target_link_libraries(compliancelib
-    PRIVATE
-        commonutils
-        logging
-)
 
 set_property(TARGET compliancelib PROPERTY POSITION_INDEPENDENT_CODE ON)
 
 target_link_libraries(compliancelib PRIVATE
     logging
     commonutils
-    parsonlib)
+    parsonlib
+    )
 
 target_include_directories(compliancelib PUBLIC
-    ${logging_SOURCE_DIR}
     ${MODULES_INC_DIR}
     ${CMAKE_CURRENT_SOURCE_DIR})
--- a/src/modules/compliance/src/lib/Engine.cpp
+++ b/src/modules/compliance/src/lib/Engine.cpp
@@ -1,0 +1,329 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include "Engine.h"
+
+#include <CommonUtils.h>
+#include "Evaluator.h"
+#include "Base64.h"
+
+#include "parson.h"
+#include <string>
+#include <vector>
+#include <map>
+#include <sstream>
+
+namespace compliance
+{
+    static constexpr const char* cLogFile = "/var/log/osconfig_compliance.log";
+    static constexpr const char* cRolledLogFile = "/var/log/osconfig_compliance.bak";
+    static constexpr const char* cModuleInfo = "{\"Name\": \"Compliance\","
+        "\"Description\": \"Provides functionality to audit and remediate Security Baseline policies on device\","
+        "\"Manufacturer\": \"Microsoft\","
+        "\"VersionMajor\": 2,"
+        "\"VersionMinor\": 0,"
+        "\"VersionInfo\": \"Dilithium\","
+        "\"Components\": [\"Compliance\"],"
+        "\"Lifetime\": 2,"
+        "\"UserAccount\": 0}";
+
+    Engine::Engine(void* log) noexcept : mLog{ log }
+    {
+    }
+
+    Engine::Engine() noexcept : mLog{ OpenLog(cLogFile, cRolledLogFile) }, mLocalLog{ true }
+    {
+    }
+
+    Engine::~Engine()
+    {
+        if(mLocalLog)
+        {
+            CloseLog(&mLog);
+        }
+    }
+
+    void Engine::setMaxPayloadSize(unsigned int value) noexcept
+    {
+        mMaxPayloadSize = value;
+    }
+
+    unsigned int Engine::getMaxPayloadSize() const noexcept
+    {
+        return mMaxPayloadSize;
+    }
+
+    OSCONFIG_LOG_HANDLE Engine::log() const noexcept
+    {
+        return mLog;
+    }
+
+    const char* Engine::getMoguleInfo() const noexcept
+    {
+        return cModuleInfo;
+    }
+
+    Result<Engine::AuditResult> Engine::mmiGet(const char* objectName)
+    {
+        if (nullptr == objectName)
+        {
+            return Error("Invalid argument", EINVAL);
+        }
+
+        OsConfigLogInfo(log(), "Engine::mmiGet(%s)", objectName);
+        auto result = AuditResult();
+        auto ruleName = std::string(objectName);
+        constexpr const char* auditPrefix = "audit";
+        if (ruleName.find(auditPrefix) != 0)
+        {
+            return Error("Invalid object name", EINVAL);
+        }
+
+        ruleName = ruleName.substr(strlen(auditPrefix));
+        if (ruleName.empty())
+        {
+            return Error("Rule name is empty", EINVAL);
+        }
+
+        auto it = mDatabase.find(ruleName);
+        if (it == mDatabase.end())
+        {
+            return Error("Rule not found", EINVAL);
+        }
+        const auto& procedure = it->second;
+        if (nullptr == procedure.audit())
+        {
+            return Error("Failed to get 'audit' object");
+        }
+
+        Evaluator evaluator(procedure.audit(), procedure.parameters(), log());
+        auto rc = evaluator.ExecuteAudit(&result.payload, &result.payloadSize);
+        if (!rc.has_value())
+        {
+            return rc.error();
+        }
+
+        result.result = rc.value();
+        return Result<AuditResult>(std::move(result));
+    }
+
+    Result<JsonWrapper> Engine::decodeB64JSON(const char* input) const
+    {
+        if (nullptr == input)
+        {
+            return Error("Input is null", EINVAL);
+        }
+        std::string inputStr(input);
+        if ((inputStr.length() > 2) && (inputStr[0] == '"') && (inputStr[inputStr.length() - 1] == '"'))
+        {
+            inputStr = inputStr.substr(1, inputStr.length() - 2);
+        }
+        auto decodedString = Base64Decode(inputStr);
+        if (!decodedString.has_value())
+        {
+            return decodedString.error();
+        }
+        auto result = json_parse_string(decodedString.value().c_str());
+        if (nullptr == result)
+        {
+            return Error("Failed to parse JSON", EINVAL);
+        }
+
+        return JsonWrapper(result);
+    }
+
+    Optional<Error> Engine::setProcedure(const std::string& ruleName, const char* payload, const int payloadSizeBytes)
+    {
+        if (ruleName.empty())
+        {
+            return Error("Rule name is empty", EINVAL);
+        }
+
+        mDatabase.erase(ruleName);
+        auto ruleJSON = decodeB64JSON(std::string(payload, payloadSizeBytes).c_str());
+        if (!ruleJSON.has_value())
+        {
+            return ruleJSON.error();
+        }
+
+        auto object = json_value_get_object(ruleJSON.value().get());
+        if (nullptr == object)
+        {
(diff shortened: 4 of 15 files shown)


src/common/base64/base64.c
```diff
@@ -0,0 +1,314 @@
...
+    uint32_t curPos = 0;
+    uint32_t dstPos = 0;
+
+    // Parameter validation
+    if (!(Input) || !(Output)) {
+        return -1;
+    }
+
+    // Calculate required output buffer length in bytes
+    reqSize = (Length == 0) ? (0) : ((((Length - 1) / 3) + 1) * 4);
```
kkanas, line 250: Can you use Base64Length macro here ?

src/common/base64/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+project(base64)
+
+add_library(base64 STATIC base64.c)
+target_compile_options(base64 PRIVATE -Wno-psabi)
```
kkanas, line 7: Can you explain -Wno-psabi rationale here?

src/common/base64/base64.h
```diff
@@ -0,0 +1,35 @@
...
+#ifndef BASE64_H
+#define BASE64_H
+
+#include <stdint.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#define Base64Length(l) ((l == 0) ? (1) : (((((l - 1) / 3) + 1) * 4) + 1))
```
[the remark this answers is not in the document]
jepio: What are the consequences of that?

src/modules/compliance/src/lib/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
...
+set_property(TARGET compliancelib PROPERTY POSITION_INDEPENDENT_CODE ON)
+
+target_link_libraries(compliancelib PRIVATE
+    logging
+    commonutils
+    parsonlib
+    base64)
+
+target_include_directories(compliancelib PUBLIC
+    ${logging_SOURCE_DIR}
```
jepio, line 25: This is not needed with cmake - logging defines `PUBLIC` include directories that are added to consumers compile commands.

src/modules/compliance/src/lib/Procedure.h
```diff
@@ -0,0 +1,44 @@
...
+struct json_value_t;
+struct json_object_t;
+
+namespace compliance
+{
+    class Procedure
+    {
+        std::map<std::string, std::string> mParameters;
+        json_value_t* mAuditRule = nullptr;
+        json_value_t* mRemediationRule = nullptr;
```
jepio, line 22: Unique_ptrs?

src/modules/compliance/src/lib/Procedure.cpp
```diff
@@ -0,0 +1,139 @@
...
+    Procedure::Procedure(Procedure&& other) noexcept
+        : mParameters(std::move(other.mParameters))
+        , mAuditRule(other.mAuditRule)
+        , mRemediationRule(other.mRemediationRule)
+    {
+        other.mAuditRule = nullptr;
+        other.mRemediationRule = nullptr;
+    }
+
+    Procedure& Procedure::operator=(Procedure&& other) noexcept
```
jepio, line 27: This leaks existing resource of this object.
You don't need to provide this if you turn the members into unique_ptrs OR you need to swap out the members between `*this` and `other`.

src/modules/compliance/src/lib/CMakeLists.txt
robertwoj-microsoft: Fixed. I added this but the problem was in a different place.

src/common/logging/Logging.h
```diff
@@ -5,6 +5,7 @@
 #define LOGGING_H
 
 #include <string.h>
+#include <stdio.h>
```
MariusNi, line 9: Would it work to include stdio.h instead where Logging.h is newly included, before it gets included?

src/modules/compliance/src/lib/Base64.cpp
```diff
@@ -0,0 +1,96 @@
...
+            return 62;
+        }
+        else if (c == '/')
+        {
+            return 63;
+        }
+        else
+        {
+            return 0;
+        }
```
MariusNi, line 35: As you may know, structured programming (which is below object oriented programming, what C++ programming attempts to achieve) does not recommend multiple exists from within a function body, aka fast returns. 

Won't cost much to declare a local variable, set that from if-else, and return that once at the end.

src/modules/compliance/src/lib/Base64.cpp
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    static inline bool is_base64(const unsigned char c)
```
MariusNi, line 38: >is_base64 [](http://example.com/codeflow?start=23&length=9)

Please use correct casing for names Function names should be PascalCased. Thus, this would be: 

IsBase64()

Here and elsewhere.

src/modules/compliance/src/lib/Base64.cpp
```diff
@@ -0,0 +1,96 @@
...
+        for (const char c : input)
+        {
+            if (!is_base64(c))
+            {
+                return Error("Invalid base64 character", EINVAL);
+            }
+        }
+
+        std::string ret;
+        ret.reserve(input.size() * 3 / 4);
```
MariusNi, line 58: >input.size() * 3 / 4) [](http://example.com/codeflow?start=20&length=21)

Please add one set of () for clarity of intention.

src/modules/compliance/src/lib/Base64.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+        for (size_t i = 0; i < input.size(); i += 4)
+        {
+            unsigned char enc[4];
+
+            int j = 0;
+            for (j = 0; j < 4; ++j)
```
MariusNi, line 65: >++j [](http://example.com/codeflow?start=31&length=3)

Just curios, why? (why not j++ here)

src/modules/compliance/src/lib/Engine.cpp
```diff
@@ -0,0 +1,324 @@
...
+
+    Engine::~Engine()
+    {
+        if(mLocalLog)
+        {
+            CloseLog(&mLog);
+        }
+    }
+
+    void Engine::setMaxPayloadSize(unsigned int value) noexcept {
```
MariusNi, line 46: >{ [](http://example.com/codeflow?start=64&length=1)

Please move opening curly brace { on the next line.

In this project, in all C and C++ code, we do:

foo
{
    ...
}

and not:

foo (
  ....
}

Here and elsewhere.

src/modules/compliance/src/lib/Engine.cpp
```diff
@@ -0,0 +1,324 @@
...
+        if (nullptr == procedure.audit())
+        {
+            return Error("Failed to get 'audit' object");
+        }
+
+        Evaluator evaluator(procedure.audit(), procedure.parameters(), log());
+        auto rc = evaluator.ExecuteAudit(&result.payload, &result.payloadSize);
+        if (!rc.has_value())
+        {
+            return rc.error();
```
MariusNi, line 99: >return rc.error(); [](http://example.com/codeflow?start=12&length=18)

Try to avoid fast returns as rest of the project is not doing this. If you want to keep fast returns, please do it consistently, from everywhere.

src/modules/compliance/src/lib/Engine.cpp
```diff
@@ -0,0 +1,324 @@
...
+
+#include <CommonUtils.h>
+#include "Evaluator.h"
+#include "Base64.h"
+
+#include "parson.h"
+#include <string>
+#include <vector>
+#include <map>
+#include <sstream>
```
MariusNi, line 14: This module has a mess of includes, from everywhere. Both headers and implementation file include on their own. We try to avoid this in this project as it may lead to tangled components that cannot be moved or shared.

Instead, please try to:

- Add a precompiled header for each library.
- Include that precompiled header into all impl files.
- Do not include anything in the library header files.
- Ensure that the correct include order is in the precompiled header.

src/modules/compliance/src/lib/Base64.cpp
wupeka: Just a habit, will fix fo consistency.

src/modules/compliance/src/lib/Base64.cpp
robertwoj-microsoft: We have prepared a document regarding C++ usage within osconfig. Please review it again, there is information we plan to use early exits: it improves readability compared to deep-nested if statements.

src/modules/compliance/src/lib/Engine.cpp
robertwoj-microsoft: See my previous comment.
> please do it consistently, from everywhere.
Please point me to places where we're inconsistent.

src/modules/compliance/src/lib/Engine.cpp
robertwoj-microsoft: > This module has a mess of includes, from everywhere

Please clarify this. All these headers are needed as far as I know.
> Add a precompiled header for each library.

Please show me where the project uses precompiled headers and how we benefit from this.
> Do not include anything in the library header files.

I don't follow, could you explain in more detail?
> Ensure that the correct include order is in the precompiled header.

Define what you think is correct order please.

src/modules/compliance/src/lib/Base64.cpp
MariusNi: With classic structured programming, there can be one early exit for function arguments validation, then the only exit is at the end of the function. With immediate returns, the style is returning at every error encountered. Either style is fine for the Compliance C++ code, we are just recommending **consistence** (no one style over the other, just chose one and be consistent). Thanks

src/modules/compliance/src/lib/Engine.cpp
MariusNi: Up to you to decide if the coding style is consistent, you own it. My comment is a recommendation for you to consider and is up to you to decide if you want to change anything for this, or not.

src/modules/compliance/src/lib/Engine.cpp
MariusNi: By "mess of includes, from everywhere' I mean entanglement of header includes from every header and every implementation file. This can grow into a big problem later and make code very hard to move around, or share with other components. Would you consider having a Common.h header for each library that does most of the includes, have all library headers be free of includes, and in the impl files only include Common.h?

robertwoj-microsoft: We have discussed the topic of header file topic today. We plan to stick to the entanglement as we believe it works best for us.

jepio: Just to be clear: we don't want **entaglement**, we just want each header file and implementation file to stay self-contained and include-what-is-used and nothing beyond that.

Precompiled headers are not commonly used in Linux projects and I don't see the need for the added complexity.

src/common/logging/Logging.h
MariusNi: Ideally, this PR should focus on this engine class which is part of the Compliance library module. Without changes in the core and other areas. I would include stdio.h before Logging.h is included.

MariusNi:
> #include "UserUtils.h"

I would include the OSConfig headers after the SDK headers, not before.

---
Refers to: src/tests/fuzzer/target.cpp:7 in 5a7a973. [](commit_id = 5a7a973105fca6ad20c692ccfd62214681bd3a51, deletion_comment = False)

src/common/logging/Logging.h
robertwoj-microsoft: OK, as you wish, I'll make this change in a separate PR.

MariusNi (dismissed):
:shipit:

robertwoj-microsoft:
> I would include the OSConfig headers after the SDK headers, not before.

This will be in a separate PR as it's out of scope here. Please let me know what you consider OSConfig headers and what are SDK headers and I'll move.

> Why is this included here?

Probably leftover, will drop in separate PR as well.

robertwoj-microsoft:
Rebased with dev.

src/modules/compliance/src/lib/Engine.h
```diff
@@ -0,0 +1,92 @@
...
+
+namespace compliance
+{
+    class Engine
+    {
+    public:
+        struct AuditResult
+        {
+            bool result = false;
+            char* payload = nullptr;
```
AhmedBM, line 30: unique_ptr instead so we can avoid having to free?

src/modules/compliance/tests/EngineTest.cpp
```diff
@@ -0,0 +1,337 @@
...
+{
+    return true;
+}
+
+class ComplianceEngineTest : public ::testing::Test
+{
+protected:
+    Engine mEngine;
+    std::map<std::string, std::pair<action_func_t, action_func_t>> mProcedureMap;
+    // std::map<std::string, std::string> mParameters;
```
AhmedBM, line 31: cleanup comment?

src/modules/compliance/tests/EngineTest.cpp
```diff
@@ -0,0 +1,337 @@
...
+
+    void SetUp() override
+    {
+        mProcedureMap = {
+            {"success", {auditSuccess, auditSuccess}},
+            {"failure", {auditFailure, auditFailure}},
+            {"NoAudit", {nullptr, auditSuccess}},
+        };
+        // mEngine.setProcedureMap(mProcedureMap);
```
AhmedBM, line 40: cleanup comment?

AhmedBM (approved):
Looks great, just some minor comments

src/modules/compliance/src/lib/Engine.h
robertwoj-microsoft: Planned to be simplified in next PRs by just using std::string, the value is copied at the end of call chain anyway.

src/modules/compliance/tests/EngineTest.cpp
robertwoj-microsoft: Fixed in a different PR, will be merged later.

src/modules/compliance/tests/EngineTest.cpp
robertwoj-microsoft: Fixed in a different PR, will be merged later.

src/modules/compliance/src/lib/Engine.cpp
```diff
@@ -0,0 +1,329 @@
...
+
+        auto error = procedure.updateUserParameters(std::string(payload, payloadSizeBytes));
+        if (error)
+        {
+            return error.value();
+        }
+
+        Evaluator evaluator(procedure.remediation(), procedure.parameters(), log());
+        auto result = evaluator.ExecuteRemediation();
+        if (!result.has_value())
```
kkanas, line 273: Can't we just `return result` ?

src/modules/compliance/src/lib/Engine.cpp
```diff
@@ -0,0 +1,329 @@
...
+        {
+            auto error = setProcedure(ruleName.substr(strlen(procedurePrefix)), payload, payloadSizeBytes);
+            if (error)
+            {
+                return error.value();
+            }
+
+            return true;
+        }
+        else if (ruleName.find(initPrefix) == 0)
```
kkanas, line 310: Do we need to have else if here ? seems like the branch above always returns

src/modules/compliance/src/lib/Engine.cpp
```diff
@@ -0,0 +1,329 @@
...
+            auto error = initAudit(ruleName.substr(strlen(initPrefix)), payload, payloadSizeBytes);
+            if(error)
+            {
+                OsConfigLogInfo(log(), "Failed to init audit: %s", error->message.c_str());
+                return error.value();
+            }
+
+            return true;
+        }
+        else if (ruleName.find(remediatePrefix) == 0)
```
kkanas, line 321: ditto ?

src/modules/compliance/src/lib/Engine.h
```diff
@@ -0,0 +1,92 @@
...
+// Licensed under the MIT License.
+
+#ifndef COMPLIANCE_ENGINE_H
+#define COMPLIANCE_ENGINE_H
+
+#include <Mmi.h>
+#include <cstdio>
+#include "Logging.h"
+
+#include "Result.h"
```
kkanas, line 11: Do we want to sort headers ? and have convetion - azure-osconfig ours and system ones ?

src/modules/compliance/src/lib/Procedure.cpp
```diff
@@ -0,0 +1,110 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include "Procedure.h"
```
kkanas, line 4: Sort  includes ?

src/modules/compliance/src/lib/Procedure.h
```diff
@@ -0,0 +1,45 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#ifndef PROCEDURE_HPP
+#define PROCEDURE_HPP
+
+#include "Result.h"
```
kkanas, line 7: ditto ?

src/modules/compliance/tests/CMakeLists.txt
```diff
@@ -11,6 +11,8 @@ add_executable(compliancetests
     OptionalTest.cpp
     ResultTest.cpp
     EvaluatorTest.cpp
+    EngineTest.cpp
```
kkanas, line 14: sort here as well ?

src/modules/compliance/src/lib/Procedure.cpp
```diff
@@ -0,0 +1,110 @@
...
+            size_t pos = token.find('=');
+            if (pos == std::string::npos)
+            {
+                continue;
+            }
+
+            std::string key = token.substr(0, pos);
+            std::string value = token.substr(pos + 1);
+
+            if (!value.empty() && value[0] == '"')
```
kkanas, line 39: I am guessing intention is to remove `'` quote from string, can we move it to utility function - I think i seen this patter somewhere in the code. If this is true then consider it, if not then drop it.

src/modules/compliance/src/lib/Procedure.cpp
robertwoj-microsoft: Will be sorted as part of PR #873

src/modules/compliance/src/lib/Engine.h
robertwoj-microsoft: Will be part of PR #873

src/modules/compliance/src/lib/Procedure.cpp
robertwoj-microsoft: I think it's a single place with this functionality. It is similar to https://github.com/Azure/azure-osconfig/blob/compliance/engine/src/modules/compliance/src/lib/Procedure.cpp#L25-L28 and https://github.com/Azure/azure-osconfig/blob/compliance/engine/src/modules/compliance/src/lib/Engine.cpp#L117-L120, but these are probably places that should be adjusted to use json_parse_string which will be done in a separate PR if applicable.

MariusNi (approved):
LGTM

PR #865, opened 2025-02-12
Link Baseline.c with the compliance module

robertwoj-microsoft:
## Description

Add calls from NRP adapter to the compliance engine.

## Checklist

### DIFF
--- a/src/adapters/mc/compliance/Baseline.c
+++ b/src/adapters/mc/compliance/Baseline.c
@@ -26,7 +26,7 @@ void BaselineInitialize(void* log)
 void BaselineShutdown(void* log)
 {
     UNUSED(log);
-    if(NULL == gCompliance)
+    if (NULL == gCompliance)
     {
         return;
     }
@@ -38,30 +38,31 @@ void BaselineShutdown(void* log)
 
 int BaselineMmiGet(const char* componentName, const char* objectName, char** payload, int* payloadSizeBytes, unsigned int maxPayloadSizeBytes, void* log)
 {
-    UNUSED(log);
-    UNUSED(maxPayloadSizeBytes);
-
-    if (0 != strcmp(componentName, gComponentName))
+    if ((NULL == componentName) || (NULL == objectName))
     {
+        OsConfigLogError(log, "BaselineMmiGet called with invalid arguments");
         return EINVAL;
     }
 
-    if (NULL == gCompliance)
+    int result = ComplianceMmiGet(gCompliance, componentName, objectName, payload, payloadSizeBytes);
+    if (MMI_OK != result)
     {
-        return EINVAL;
+        OsConfigLogError(log, "BaselineMmiGet(%s, %s) failed: %d", componentName, objectName, result);
+        return result;
     }
 
-    return ComplianceMmiGet(gCompliance, componentName, objectName, payload, payloadSizeBytes);
+    if ((NULL != *payload) && (*payloadSizeBytes > 0) & (maxPayloadSizeBytes > 0) && ((unsigned)*payloadSizeBytes > maxPayloadSizeBytes))
+    {
+        OsConfigLogInfo(log, "BaselineMmiGet(%s, %s) payload truncated from %d to %u bytes", componentName, objectName, *payloadSizeBytes, maxPayloadSizeBytes);
+        *payloadSizeBytes = (int)maxPayloadSizeBytes;
+        *payload[*payloadSizeBytes] = '\0';
+    }
+
+    return MMI_OK;
 }
 
 int BaselineMmiSet(const char* componentName, const char* objectName, const char* payload, const int payloadSizeBytes, void* log)
 {
     UNUSED(log);
-
-    if (0 != strcmp(componentName, gComponentName))
-    {
-        return EINVAL;
-    }
-
     return ComplianceMmiSet(gCompliance, componentName, objectName, payload, payloadSizeBytes);
 }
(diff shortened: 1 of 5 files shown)


MariusNi (approved):
:shipit:

MariusNi:
> #include "../Common.h"

This Baseline.c file does not need to be in C. Can be written in C++. The baseline implementation is declared extern C so it works with both C and C++ implementations.

---
Refers to: src/adapters/mc/compliance/Baseline.c:4 in 80fa904. [](commit_id = 80fa904e6f0c95ee655c10921674b87172b6220c, deletion_comment = False)

src/modules/compliance/src/lib/Engine.cpp
```diff
@@ -55,7 +55,7 @@ namespace compliance
         return mLog;
     }
 
-    const char* Engine::getMoguleInfo() const noexcept
+    const char* Engine::getMoguleInfo() noexcept
```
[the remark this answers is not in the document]
MariusNi: Shouldn't 'getMoguleInfo()' be 'getModuleInfo()'?

src/adapters/mc/compliance/Baseline.c
```diff
@@ -38,30 +38,25 @@ void BaselineShutdown(void* log)
...
-    UNUSED(log);
-    UNUSED(maxPayloadSizeBytes);
-
-    if (0 != strcmp(componentName, gComponentName))
+    int result = ComplianceMmiGet(gCompliance, componentName, objectName, payload, payloadSizeBytes);
+    if (MMI_OK != result)
     {
-        return EINVAL;
+        OsConfigLogError(log, "ComplianceMmiGet(%s, %s) failed: %d", componentName, objectName, result);
+        return result;
```
MariusNi, line 45: >return result; [](http://example.com/codeflow?start=8&length=14)

Make sure there is no leaked memory. Call FREE_MEMORY on payload, or similar.

src/modules/compliance/src/lib/ComplianceInterface.cpp
```diff
@@ -15,45 +21,123 @@ void ComplianceShutdown(void)
...
-    UNUSED(payload);
-    UNUSED(payloadSizeBytes);
-    return -1;
+    if ((nullptr == payload) || (nullptr == payloadSizeBytes))
+    {
+        OsConfigLogError(g_log, "ComplianceMmiGetInfo(%s, %p, %p) called with invalid arguments", clientName, payload, payloadSizeBytes);
+        return EINVAL;
+    }
+
+    *payload = (char*)strdup(Engine::getModuleInfo());
```
MariusNi, line 42: >*payload = (char*)strdup(Engine::getModuleInfo()); [](http://example.com/codeflow?start=4&length=50)

In general, MMI payloads are not NULL terminated (exclude the NULL terminator).

MariusNi:
>     Result<JsonWrapper> Engine::decodeB64JSON(const char* input) const

Sorry for not spotting this earlier. Shouldn't name be: 'DecodeBase64Json'?

Let's remember that:

- Method names should be in PascalCasing (not camelCasing).
- Be composed of full words (not abbreviations).

Please consider correcting if appliable all other similar instances. It is perfectly fine if not done in this PR but in a subsequent one.

---
Refers to: src/modules/compliance/src/lib/Engine.cpp:110 in 33ede1a. [](commit_id = 33ede1aacc289a0a9a2b7ac348f0cefc161a04ec, deletion_comment = False)

robertwoj-microsoft:
> Shouldn't name be: 'DecodeBase64Json'?

This will be fixed as part of PR #878

src/modules/compliance/src/lib/ComplianceInterface.cpp
wupeka, line 42: Why cast here?

wupeka (approved):

src/adapters/mc/compliance/Baseline.c
kkanas, line 45: If the payload is not allocated on error condition then this function when called from 
 BaselineMmiGet  from src/adapters/mc/compliance/Baseline.c:44  by which in turn  
src/adapters/mc/OsConfigResource.c:412
LogInfo(context, GetLog(), "[%s] BaselineMmiGet(%s, %s): '%s' (%d)", who, componentName, g_reportedObjectName, objectValue, objectValueLength);
and as far as I know printing string that points to null is undefined behaviour

src/modules/compliance/src/lib/ComplianceInterface.cpp
```diff
@@ -15,45 +21,123 @@ void ComplianceShutdown(void)
...
+        }
+
+        *payload = strdup(result.value().payload);
+        *payloadSizeBytes = result.value().payloadSize;
+        OsConfigLogInfo(engine.log(), "MmiGet(%p, %s, %s, %.*s)", clientSession, componentName, objectName, *payloadSizeBytes, *payload);
+        return MMI_OK;
+    }
+    catch (const std::exception& e)
+    {
+        OsConfigLogError(engine.log(), "ComplianceMmiGet failed: %s", e.what());
```
kkanas, line 93: Should we here also set 
   *payload = NULL;
    *payloadSizeBytes = 0;
as they can be changed by MmiGet and be in inconsistent state if exception is thrown ?

robertwoj-microsoft: engine.mmiGet() doesn't modify the payload variable.

kkanas (approved):

PR #869, opened 2025-02-12
Correcting the ASB remediations that include removing user accounts

MariusNi:
## Description

The work in this PR includes the following:

- Several automatic remediation to remove duplicate UIDs, GIDs, usernames, groups, is replaced by audits reporting that automatic remediation is not possible, and that the user must manually remove the duplicate. This is because when duplicates exist, we do not know which is the original, which is the duplicate, which is the correct one to remain, etc. Following audits: **auditEnsureNoDuplicateUidsExist**, **auditEnsureNoDuplicateGidsExist**, **auditEnsureNoDuplicateUserNamesExist**, **auditEnsureNoDuplicateGroupsExist**.
- Following remediations become no-op:  **remediateEnsureNoDuplicateUidsExist**, **remediateEnsureNoDuplicateGidsExist**, **remediateEnsureNoDuplicateUserNamesExist**, **remediateEnsureNoDuplicateGroupsExist**. They only log.
- **remediateEnsureSystemAccountsAreNonLogin** is updated, before removing the respective accounts, to try to make then first non-login. When and if removing the accounts, to do it without removing the home directories to prevent data loss.
- **remediateEnsureUnnecessaryAccountsAreRemoved** also no longer removes home dirs. 
- in UserUtils.c|h, the following functions are removed: SetNoDuplicateUids, SetNoDuplicateGids, SetNoDuplicateUserNames, SetNoDuplicateGroupNames.
- in UserUtils.c, SetRootIsOnlyUidZeroAccount is changed to remove such accounts without deleting the home dirs. This affects: **remediateEnsureNonRootAccountsHaveUniqueUidsGreaterThanZero** and **remediateEnsureRootIsOnlyUidZeroAccount**.

## Checklist


AhmedBM (approved):
👌

robertschaedler3 (approved):

AhmedBM (approved):

src/common/commonutils/UserUtils.c
```diff
@@ -2804,10 +2668,10 @@ int RemoveSystemAccountsThatCanLogin(void* log)
         {
             if ((userList[i].isLocked || userList[i].noLogin || userList[i].cannotLogin) && userList[i].hasPassword && userList[i].userId)
             {
-                OsConfigLogError(log, "RemoveSystemAccountsThatCanLogin: user '%s' (%u, %u, '%s', '%s') is either locked, no-login, or cannot-login, "
+                OsConfigLogError(log, "LockSystemAccountsThatCanLogin: user '%s' (%u, %u, '%s', '%s') is either locked, no-login, or cannot-login, "
                     "but can login with password",  userList[i].username, userList[i].userId, userList[i].groupId, userList[i].home, userList[i].shell);
 
-                if ((0 != (_status = RemoveUser(&(userList[i]), log))) && (0 == status))
+                if ((0 != (_status = LockUser(&(userList[i]), log))) && (0 == status))
```
kkanas, line 2674: Is `0 == status` necessary ? IIUC the status is set `(status = EnumerateUsers(&userList, &userListSize, NULL, log)))` on line 265 and not changed later on.

robertwoj-microsoft: status if overwritten in the if body, 2 lines below. Which means we won't stop the loop, but won't overwrite status with 0 if it's non-zero already.

robertwoj-microsoft (changes requested):
I think we're still removing home directories.
Could you also explain why we sometimes pass true and sometimes false as the force flag? I don't understand why. Maybe you could include a comment when calling RemoveUser and other functions that pass around the force flag.

src/common/commonutils/UserUtils.c
kkanas, line 2674: What is rationale behind locking users ? This would also not lock root user with userId 0 IIUC (It would be the first duplicate userid I would create if I could).

src/common/commonutils/UserUtils.c
```diff
@@ -1601,7 +1465,7 @@ int SetRootIsOnlyUidZeroAccount(void* log)
                 OsConfigLogError(log, "SetRootIsOnlyUidZeroAccount: user '%s' (%u, %u) is not root but has uid 0",
                     userList[i].username, userList[i].userId, userList[i].groupId);
 
-                if ((0 != (_status = RemoveUser(&(userList[i]), log))) && (0 == status))
+                if ((0 != (_status = RemoveUser(&(userList[i]), false, log))) && (0 == status))
```
kkanas, line 1468: Why here we are not adding --force flag to remove user but in  RemediateEnsureUnnecessaryAccountsAreRemoved we add --force flag ?

src/common/commonutils/UserUtils.c
MariusNi: Root user is prevented from being locked, see line 855. Overall lthe rationale behind locking a user account is to block that user account from logging in. As seen as a less radical approach of removing (deleting that user account). Is your opinion that we should keep the user removal here?

src/common/commonutils/UserUtils.c
MariusNi: We do this in order to not cause data loss, as the feedback from Azure Core Linux team was.  Is your opinion to delete the home dir?

src/common/commonutils/UserUtils.c
```diff
@@ -810,7 +810,7 @@ static int RemoveUser(SIMPLIFIED_USER* user, void* log)
         return EPERM;
     }
 
-    if (NULL != (command = FormatAllocateString(commandTemplate, user->username)))
+    if (NULL != (command = FormatAllocateString(commandTemplate, force ? "-f -r" : "-r", user->username)))
```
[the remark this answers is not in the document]
MariusNi: Good catch! Will fix.

src/common/commonutils/UserUtils.c
kkanas: I understand that logic now - but we won't be able to trace from logs which  users  were locked successfully and which were not.  I would at least add log statement when LockUser failed

src/common/commonutils/UserUtils.c
kkanas: No I would not delete the user, not lock the user, fail the policy scan until some administrator or someone will fix it manually. 
If we think the purpose of remediate action is to 'fix' system so the audit check will pass this implementation will break this contract.

src/common/commonutils/UserUtils.c
kkanas: Ok, for root user - for the rest - I the best I can think how to handle situation is https://github.com/Azure/azure-osconfig/pull/869#discussion_r1954990554

MariusNi:
> I think we're still removing home directories. Could you also explain why we sometimes pass true and sometimes false as the force flag? I don't understand why. Maybe you could include a comment when calling RemoveUser and other functions that pass around the force flag.

Replaced 'force' with 'removeHomeDir/removeHomeDirs' to make it clearer what it does. The intent is to avoid deleting home directories unless absolutely necessary.

src/common/commonutils/UserUtils.c
MariusNi: Where we cannot remediate with certitude, we do not. In this case, yes, we can remediate. We can detect for example with an user that is not named root uses uid 0 and remove such account. The intent for remediation is to be as automatic as possible.

src/common/asb/Asb.c
```diff
@@ -3210,14 +3248,14 @@ static int RemediateEnsurePasswordExpirationWarning(char* value, void* log)
 static int RemediateEnsureSystemAccountsAreNonLogin(char* value, void* log)
 {
     UNUSED(value);
-    return RemoveSystemAccountsThatCanLogin(log);
+    return LockSystemAccountsThatCanLogin(log);
```
kkanas, line 3251: IIUC the RemediateEnsureSystemAccountsAreNonLogin is ran for 
```
  ResourceID = "Ensure all system accounts are non-login (CIS: L1 - Server - 5.4.2)"; 
  ...
   DesiredObjectName = "remediateEnsureSystemAccountsAreNonLogin";
```
The publicly available description of that rule I found shows 
https://secscan.acron.pl/centos7/5/4/2
which is quite different approach from your implementation. Is this intentional ?  (the linked Remediation can pose problems as well).  Should we document it somehow ?

robertwoj-microsoft (commented):
Do you plan to add any tests? This seems like quite a critical change.
Also, we switched from removing users to locking users and there is no test update here. Have we ever had test coverage for this functionality? If we had and we made no changes to test suite, I'd expect test failures.

src/common/asb/Asb.c
MariusNi: Very good feedback, thanks! The remediation needs to match the audit, besides matching the rule definition, and work for all target 21+ distros.  What I did here is: removed the newly added lock user, kept remove user (w/o deleting the home dirs), and added a new local function to set an user account non-login in a similar way to that CentOS page, but to work with all possible non-login shells.

MariusNi:
> Do you plan to add any tests? This seems like quite a critical change. Also, we switched from removing users to locking users and there is no test update here. Have we ever had test coverage for this functionality? If we had and we made no changes to test suite, I'd expect test failures.

You are absolutely right; we need these to remain anchored to what was validated. There are no UTs for these for obvious reasons. The functional tests exercise these functions. Removed the newly added lock user. Kept removing users and added a way to make system users non-login (see the other comment).

src/common/commonutils/UserUtils.c
```diff
@@ -929,7 +923,7 @@ int CheckNoDuplicateGidsExist(char** reason, void* log)
     return status;
 }
 
-static int RemoveGroup(SIMPLIFIED_GROUP* group, void* log)
+int RemoveGroup(SIMPLIFIED_GROUP* group, bool force, void* log)
```
kkanas: Is RemoveGroup called somewhere in the code? If not then what is rationale for leaving it here.

robertwoj-microsoft:
> You are absolutely right; we need these to remain anchored to what was validated. There are no UTs for these for obvious reasons. The functional tests exercise these functions.

Can you point me to the functional tests? I would expect them to fail if the logic changed as we no longer remove users.

src/common/commonutils/UserUtils.c
MariusNi: It's not currently used but it might be in the future.

MariusNi:
> > You are absolutely right; we need these to remain anchored to what was validated. There are no UTs for these for obvious reasons. The functional tests exercise these functions.
> 
> Can you point me to the functional tests? I would expect them to fail if the logic changed as we no longer remove users.

The MC tests and the ModuleTest tests, which are run by CI and also can be manually run.

robertwoj-microsoft (approved):

PR #873, opened 2025-02-14
Compliance/code formatting

robertwoj-microsoft:
## Description

Introduce code formatting. This only applies to src/modules/compliance subdirectory.

## Checklist


.clang-format
```diff
@@ -0,0 +1,87 @@
...
+IndentCaseLabels: true
+IndentExternBlock: NoIndent
+IndentWidth: 4
+IndentPPDirectives: AfterHash
+IndentWrappedFunctionNames: true
+KeepEmptyLinesAtTheStartOfBlocks: true
+MacroBlockBegin: ''
+MacroBlockEnd: ''
+MaxEmptyLinesToKeep: 1
+NamespaceIndentation: All
```
jepio, line 59: How about switching to `NamespaceIndentation: None`?

src/modules/compliance/src/lib/Evaluator.cpp
```diff
@@ -1,259 +1,263 @@
...
-#define lenof(str) (sizeof(str)-1)
+#include <cstring>
+#include <map>
+#include <parson.h>
+#include <sstream>
+#include <string>
+#include <utility>
 
-namespace compliance {
+#define lenof(str) (sizeof(str) - 1)
```
jepio, line 16: In all cases where this works (string literals/static string arrays), any compiler that will use will evaluate `strlen` at compile time and just inject the constant.
So how about we get rid of this and just make it `strlen`?

.clang-format
robertwoj-microsoft: Indenting is my preference, but I'm open for suggestions.

wupeka: I'm with Jeremi on that, since usually the whole file is indented it just adds unnecessary level.

src/modules/compliance/src/lib/Evaluator.cpp
robertwoj-microsoft: Does it make any difference? Any way works for me really.

.clang-format
robertwoj-microsoft: Fair enough, changing now.

src/modules/compliance/src/lib/ComplianceInterface.cpp
```diff
@@ -2,8 +2,16 @@
...
 #include "Engine.h"
+#include "Logging.h"
+#include "Mmi.h"
+
+#include <cerrno>
+#include <cstddef>
+#include <cstring>
+#include <exception>
 
 using compliance::Engine;
```
MariusNi, line 16: Is there a formatting rule that will always put the system/SDK header includes above/before header includes from our project? As the later may require the former being included first.

src/modules/compliance/src/lib/Engine.cpp
```diff
@@ -3,327 +3,336 @@
...
+static constexpr const char* cRolledLogFile = "/var/log/osconfig_compliance.bak";
+static constexpr const char* cModuleInfo = "{\"Name\": \"Compliance\","
+                                           "\"Description\": \"Provides functionality to audit and remediate Security Baseline policies on device\","
+                                           "\"Manufacturer\": \"Microsoft\","
+                                           "\"VersionMajor\": 2,"
+                                           "\"VersionMinor\": 0,"
+                                           "\"VersionInfo\": \"Dilithium\","
+                                           "\"Components\": [\"Compliance\"],"
+                                           "\"Lifetime\": 2,"
+                                           "\"UserAccount\": 0}";
```
MariusNi, line 34: Can we have a rule that aligns such constant lines only 4 spaces from left, rather than doing this sort of vertical alignment?

src/modules/compliance/src/lib/Engine.cpp
```diff
@@ -3,327 +3,336 @@
...
+Engine::Engine(void* log) noexcept
+    : mLog{ log }
+{
+}
 
-    Engine::Engine(void* log) noexcept : mLog{ log }
-    {
-    }
+Engine::Engine() noexcept
+    : mLog{ OpenLog(cLogFile, cRolledLogFile) }
```
MariusNi, line 42: > mLog{ OpenLog(cLogFile, cRolledLogFile) } [](http://example.com/codeflow?start=5&length=42)

Can we have a rule that opens all () pairs on separate lines?

'
{
  ...
}
'
instead of:

'
{...}
'

src/modules/compliance/src/lib/Evaluator.cpp
```diff
@@ -1,18 +1,22 @@
...
-#include <string>
-#include <memory>
-#include <map>
-#include <sstream>
-#include <parson.h>
-#include <Logging.h>
 #include "Evaluator.h"
+
+#include "Logging.h"
 #include "Result.h"
```
MariusNi, line 8: Why we always include our own headers first? We may have things in them that need system/SDK headers included first.

src/modules/compliance/src/lib/Evaluator.cpp
```diff
@@ -1,18 +1,22 @@
...
 
-#define lenof(str) (sizeof(str)-1)
+#include <cstring>
+#include <map>
+#include <parson.h>
+#include <sstream>
+#include <string>
+#include <utility>
+
+#define lenof(str) (sizeof(str) - 1)
```
MariusNi, line 17: "lenof" is not correct camelCasing or PascalCasing. We need full words for those both. 

Also, in C++, we want to have constants and inline functions as opposed to macros (which are C style).

Possible to switch this to a static global method and rename it as: 'LenghtOf'? For example

src/modules/compliance/src/lib/JsonWrapper.cpp
```diff
@@ -7,13 +7,13 @@
...
+void JsonWrapperDeleter::operator()(json_value_t* value) const
+{
+    json_value_free(value);
+}
 
-    JsonWrapper parseJSON(const char* input)
-    {
-        return JsonWrapper(json_parse_string(input), JsonWrapperDeleter());
-    }
+JsonWrapper parseJSON(const char* input)
```
MariusNi, line 15: >parseJSON [](http://example.com/codeflow?start=12&length=9)

Correct PascalCasing is: 'ParseJson'

src/modules/compliance/src/lib/Procedure.cpp
```diff
@@ -2,109 +2,112 @@
...
-        std::string token;
+Optional<Error> Procedure::updateUserParameters(const std::string& input)
+{
+    std::istringstream stream(input);
+    std::string token;
 
-        while (std::getline(stream, token, ' '))
+    while (std::getline(stream, token, ' '))
+    {
+        // In the non-NRP scenario the token is delimited by double quotes
```
MariusNi, line 25: >the token is delimited by double quotes [](http://example.com/codeflow?start=35&length=39)

Why? Is this part of JSON encoding or something else?

src/modules/compliance/src/lib/Procedure.h
```diff
@@ -16,30 +16,33 @@ struct json_object_t;
...
-        Optional<Error> updateUserParameters(const std::string &userParameters);
-        Optional<Error> setAudit(const json_value_t *rule);
-        Optional<Error> setRemediation(const json_value_t* rule);
-    };
+        return mParameters;
+    }
+    const json_object_t* audit() const noexcept;
+    const json_object_t* remediation() const noexcept;
+
+    void setParameter(const std::string& key, std::string value);
```
MariusNi, line 41: >setParameter [](http://example.com/codeflow?start=9&length=12)

Function/method names should be in PascalCasing. Thus, here: 'SetParamater'.

src/modules/compliance/src/lib/Base64.cpp
```diff
@@ -1,96 +1,98 @@
...
+
+#include <string>
+
 namespace compliance
 {
-    static inline char Base64Char(const unsigned char c)
+static inline char Base64Char(const unsigned char c)
+{
+    if (c >= 'A' && c <= 'Z')
     {
```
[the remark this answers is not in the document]
robertwoj-microsoft: Unfortunately not.

src/modules/compliance/src/lib/ComplianceInterface.cpp
robertwoj-microsoft: It's the opposite usually - the most local to the most wide headers.

src/modules/compliance/src/lib/Engine.cpp
robertwoj-microsoft: there should be an option. I'll see

src/modules/compliance/src/lib/Engine.cpp
robertwoj-microsoft: Not for this - it's member initialization here, not a statement block.

src/modules/compliance/src/lib/Evaluator.cpp
robertwoj-microsoft: See my other comments.

src/modules/compliance/src/lib/Evaluator.cpp
MariusNi: I am with Jeremi on this one. We should not duplicate public APIs for no added value. One vote for using strlen()

src/modules/compliance/src/lib/Procedure.cpp
robertwoj-microsoft: Good catch, thank you. I need to investigate it in more detail and it should use json_parse_string here. Work on this will be done in a separate PR.

src/modules/compliance/src/lib/JsonWrapper.cpp
robertwoj-microsoft: Not in clang-format, clang-tidy has checks and fixes for it, but it's out of scope here.

src/modules/compliance/src/lib/Engine.cpp
robertwoj-microsoft: I checked and unfortunately it cannot be controlled in a simple way.

.clang-format
kkanas: Still the indentation is to All - I would go with none, or Inner but I hope we don't need to have nested namespaces.

.clang-format
```diff
@@ -0,0 +1,86 @@
...
+BreakBeforeBraces: Custom
+BreakBeforeTernaryOperators: false
+BreakConstructorInitializers: BeforeComma
+ColumnLimit: 152
+CommentPragmas: '^ IWYU pragma:'
+CompactNamespaces: false
+ConstructorInitializerAllOnOneLineOrOnePerLine: false
+ConstructorInitializerIndentWidth: 4
+ContinuationIndentWidth: 4
+Cpp11BracedListStyle: false
```
kkanas, line 45: Cpp11BracedListStyle: true ? 
```
true:                                  false:
vector<int> x{1, 2, 3, 4};     vs.     vector<int> x{ 1, 2, 3, 4 };
vector<T> x{{}, {}, {}, {}};           vector<T> x{ {}, {}, {}, {} };
f(MyMap[{composite, key}]);            f(MyMap[{ composite, key }]);
new int[3]{1, 2, 3};                   new int[3]{ 1, 2, 3 };
```

.clang-format
```diff
@@ -0,0 +1,86 @@
...
+NamespaceIndentation: All
+PenaltyBreakComment: 300
+PenaltyBreakFirstLessLess: 120
+PenaltyBreakOpenParenthesis: 100
+PenaltyBreakString: 1000
+PenaltyExcessCharacter: 1
+PenaltyReturnTypeOnItsOwnLine: 60
+PointerAlignment: Left
+ReflowComments: true
+SortIncludes: false
```
kkanas: SortIncludes: CaseSensitive

AhmedBM (approved):

jepio (approved):
lgtm, let's get this in before any other PRs for the compliance module.

MariusNi:
> lgtm, let's get this in before any other PRs for the compliance module.

There are merge conflicts in src/modules/compliance/src/lib/Engine.h. Please fix this so we can merge this PR.

MariusNi (approved):
:shipit:

PR #881, opened 2025-02-21
Compliance/procedures pt1

wupeka:
## Description
First set of compliance procedures - testing procedures, ensureFilesystemOptions, ensureFilePermissions. 
FilePermissions procedures are missing tests. 

## Checklist


src/modules/compliance/src/lib/CMakeLists.txt
```diff
@@ -16,6 +18,13 @@ add_library(compliancelib STATIC
 
 set_property(TARGET compliancelib PROPERTY POSITION_INDEPENDENT_CODE ON)
 
+include(FindPerl)
+add_custom_command(
+    OUTPUT ProcedureMap.cpp ProcedureMap.h
```
robertwoj-microsoft, line 23: make clean will drop these files

src/modules/compliance/src/lib/GenProcedureMap.py
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env python
```
robertwoj-microsoft, line 1: If we have the perl version, we should drop this one I guess

src/modules/compliance/src/lib/procedures/ensureFilePermissions.cpp
```diff
@@ -0,0 +1,169 @@
...
+#include <sys/types.h>
+#include <string.h>
+#include <unistd.h>
+#include <grp.h>
+#include <pwd.h>
+
+
+namespace compliance {
+
+AUDIT_FN(ensureFilePermissions)
```
robertwoj-microsoft, line 14: Could we get rid of these macros? it reads as 
```suggestion
Result<bool> Audit_ensureFilePermissions(std::map<std::string, std::string> args, std::ostringstream &logstream)
```

src/modules/compliance/src/lib/procedures/ensureFilePermissions.cpp
```diff
@@ -0,0 +1,169 @@
...
+{
+    struct stat statbuf;
```
robertwoj-microsoft, line 16: I think we can skip the struct keyword here

src/modules/compliance/src/lib/procedures/ensureFilePermissions.cpp
```diff
@@ -0,0 +1,169 @@
...
+    if (args.find("filename") == args.end())
+    {
+        return Error("No filename provided");
+    }
+    logstream << "ensureFilePermissions for " << args["filename"];
+
+    if (stat(args["filename"].c_str(), &statbuf) < 0)
+    {
+        return Error("Stat error");
```
robertwoj-microsoft, line 25: Could we include strerror or errno in the message?

src/modules/compliance/src/lib/CMakeLists.txt
```diff
@@ -16,6 +18,13 @@ add_library(compliancelib STATIC
 
 set_property(TARGET compliancelib PROPERTY POSITION_INDEPENDENT_CODE ON)
 
+include(FindPerl)
```
AhmedBM, line 21: Is Perl a new build requirement?

src/modules/compliance/src/lib/GenProcedureMap.pl
```diff
@@ -0,0 +1,99 @@
+#!/usr/bin/env perl
```
AhmedBM, line 1: Can we not just use bash for this? Bringing in Perl just for this seems like overkill

src/modules/compliance/src/lib/procedures/ensureFilesystemOption.cpp
```diff
@@ -0,0 +1,298 @@
...
+    int lineno;
+};
+
+static Result<std::map<std::string, FstabEntry>> ParseFstab(const std::string& filePath)
+{
+    std::map<std::string, FstabEntry> fstabMap;
+    FILE* file = fopen(filePath.c_str(), "r");
+    if (nullptr == file)
+    {
+            return Error("Failed to open file " + filePath + " with error " + strerror(errno));
```
AhmedBM, line 31: more than 4-spaces, pre-commit should fix this...

src/modules/compliance/src/lib/procedures/ensureFilesystemOption.cpp
```diff
@@ -0,0 +1,298 @@
...
+        std::istringstream optionsStream(mnt->mnt_opts);
+        std::string option;
+        while (std::getline(optionsStream, option, ','))
+        {
+            entry.options.push_back(option);
+        }
+
+        fstabMap[mnt->mnt_dir] = entry;
+    }
+    fclose(file);
```
AhmedBM, line 75: ignoring fclose return? Do we care if we cannot close file?

src/modules/compliance/src/lib/procedures/ensureFilePermissions.cpp
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    if (args.find("user") != args.end())
+    {
+        struct passwd *pwd = getpwuid(statbuf.st_uid);
+        if (nullptr == pwd)
+        {
+            logstream << "No user with uid " << statbuf.st_uid;
```
AhmedBM, line 33: why are we not using the OSCONFIG_LOG_INFO/OSCONFIG_LOG_ERROR macros here?

src/modules/compliance/tests/CMakeLists.txt
```diff
@@ -5,6 +5,7 @@ project(compliancetests)
 
 include(CTest)
 find_package(GTest REQUIRED)
+FILE(GLOB PROCEDURES procedures/*.cpp)
```
AhmedBM, line 8: Im generally against CMake file GLOBS because CMake cannot track changes and re-run if there are new/removed files but i believe these procedures are generated dynamically? If so, i understand the need to

AhmedBM (commented):
If Perl is being added as a build requirement, we would need to update every build container to include it. These are located in [devops/docker](https://github.com/Azure/azure-osconfig/tree/dev/devops/docker)

Makes me wonder though, can we not just do this in bash?

src/modules/compliance/src/lib/GenProcedureMap.pl
wupeka: 1. Perl is currently in all of our build environments. (unfortunately, Python is not).
2. We haven't decided finally yet, but we'll probably pre-generate the ProcedureMap and put it in the repo, so this script won't be launched at build time - only manually when someone adds/removes a procedure. For now, with the amount of churn, it's easier to just have it automated.

src/modules/compliance/src/lib/procedures/ensureFilePermissions.cpp
wupeka: The logstream is used to create a string with the result of the audit for the UI. It is then also logged to OSCONFIG_LOG.

src/modules/compliance/tests/CMakeLists.txt
wupeka: Procedures are not generated dynamically, but at this stage we expect to have a lot of churn both here and in src/modules/compliance/src/lib/procedures, and we want to parallelilze the work on different procedures in different files. Having a glob means we don't need to change CMakeLists.txt with every new file, so we won't have conflicts. In the end we'll just put a static list there. (The same rationale applies to generating ProcedureMap at build  time)

src/modules/compliance/src/lib/GenProcedureMap.py
wupeka: With the plan of having the regeneration done manually after the "big" work is done, and that the perl version is only here because some of our build environments lack Python, I'd keep both, and when we move to manual  regeneration drop the perl version, keep the python. But I won't die on this hill.

src/modules/compliance/src/lib/procedures/ensureFilesystemOption.cpp
wupeka: It's open for read only, so we don't care (and it should never fail in those circumstances).

src/modules/compliance/src/lib/GenProcedureMap.py
wupeka: Moved to shell version, dropped everything else.

src/modules/compliance/src/lib/procedures/testingProcedures.cpp
```diff
@@ -0,0 +1,56 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+#include <Evaluator.h>
+#include <Result.h>
+#include <CommonUtils.h>
+
+namespace compliance {
+REMEDIATE_FN(remediationFailure)
```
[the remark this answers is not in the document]
wupeka: We will be using those procedures for end-to-end tests etc, the costs of having them compiled in is neglible, so I'd keep them.

src/modules/compliance/tests/CMakeLists.txt
AhmedBM: i see, can you add a comment above to state that the GLOB is temporary until we stabilize?

src/modules/compliance/src/lib/procedures/ensureFilePermissions.cpp
AhmedBM: When you say the UI, do you mean this is whats displayed in the non-compliance reason in the portal? In the end however, wouldn't this string be a very long concatenated string, im not seeing any newlines?

src/modules/compliance/src/lib/GenProcedureMap.pl
AhmedBM: Ohh nice, i didn't realize the environments already contained Perl, i just validated that ubun14.04/debian-10 have it 👌

src/modules/compliance/src/lib/procedures/ensureFilePermissions.cpp
wupeka: Unfortunately - yes, that's the only thing that we can (at least for now) present to the user.
It usually won't be that complex though.

src/modules/compliance/tests/CMakeLists.txt
wupeka: Done

MariusNi (dismissed):
:shipit:

src/modules/compliance/src/lib/procedures/ensureFilePermissions.cpp
wupeka: Easier to generate the map + guarantees the consistency

MariusNi (approved):
:shipit:

robertwoj-microsoft (approved):

robertwoj-microsoft (approved):

PR #879, opened 2025-02-18
First phase of improvements for 'OSConfig for MC' logging

MariusNi:
## Description

Here is a first phase of improvements applied to the 'OSConfig for MC' logging, including:
- The informational level marker made visible ('[INFO]'). 
- Moved both '[ERROR]' and '[INFO]' to beginning of the trace lines, after date and time and before source file and line number, for better alignment of the log contents and easier parsing.
- Throughout the 'OSConfig for MC' source code, changed the error traces to information traces for all cases that are not either: source code bugs (such as: function called with invalid arguments), fatal errors (such as: out of memory), or input errors (such as, invalid rule parameter values in MOF).
- A 3rd trace label "DEBUG" to be used when Command Logging is enabled. 

For examples how these tracing changes look, see the log artifacts produced by the CI for the various tests run.

## Checklist

- [x] I have read the [contribution guidelines](https://github.com/Azure/azure-osconfig/blob/main/CONTRIBUTING.md).
- [x] I added unit-tests to validate my changes. All unit tests are passing.
- [x] I have merged the latest `dev` branch prior to this PR submission.
- [x] I ran [pre-commit](https://pre-commit.com/) on my changes prior to this PR submission.
- [x] I submitted this PR against the `dev` branch.

PR #825, opened 2024-12-03
Add missing 'log' argument in BackupSshdConfig

jepio:
## Description

`BackupSshdConfig()` passes `log` to `SavePayloadToFile()`, but it doesn't take a `log` parameter. This is picking up the function pointer to the log(arithm) function instead. If there is any error in `SavePayloadToFile`, logging segfaults. Add the missing parameter.

To ensure this doesn't regress, added unit tests for internals of `SshUtils`. These will be extended with new test cases.

## Checklist

- [x] I have read the [contribution guidelines](https://github.com/Azure/azure-osconfig/blob/main/CONTRIBUTING.md).
- [x] I added unit-tests to validate my changes. All unit tests are passing.
- [x] I have merged the latest `main` branch prior to this PR submission.
- [x] I submitted this PR against the `main` branch.

PR #883, opened 2025-02-24
Type safety for OsConfig logger

wupeka:
## Description
Make OsConfig logger type safe (and fix one potential crash issue that was not detected because of lack of thereof - src/common/commonutils/MountUtils.c:186, similiar to PR#825).

## Checklist


MariusNi:
The changes in this PR collide with the work in PR in PR# 879. Please close this PR and instead please discuss with me the idea to change the internal type for the OSCONFIG_LOG_HANDLE type to make it more secure. I will include this change as part of the logging work started with PR #879. Thanks

robertschaedler3 (commented):
Can you provide more details in the PR description about the potential crash issue being fixed? Thanks!

src/common/logging/Logging.h
```diff
@@ -16,59 +17,24 @@
 #define ARRAY_SIZE(a) (sizeof(a) / sizeof(a[0]))
 #endif
 
-typedef void* OSCONFIG_LOG_HANDLE;
+typedef struct OsConfigLog OsConfigLog;
```
MariusNi, line 20: Please do not rename this structure. As commented separately, we do not want callers to get the structure, we want the callers to get an opaque handle. 

Let's discuss separately the pros and cons of changing from opaque handle to well defined type. I want to incorporate this as part of my work for the logging improvements that is in progress right now.

docs/modules.md
```diff
@@ -879,11 +879,11 @@ To enable and use logging via this library, modules must use the following calls
 
 Call | Arguments | Returns | Description
 -----|-----|-----|-----
-OpenLog | The name of the log and of the rollover backup log. These names must be "/var/log/osconfig_*modulename*.log" and "/var/log/osconfig_*modulename*.bak" respectively | A OSCONFIG_LOG_HANDLE handle | Called when the module starts to open the log
-CloseLog | A OSCONFIG_LOG_HANDLE handle | None | Called when the module terminates to close the log
+OpenLog | The name of the log and of the rollover backup log. These names must be "/var/log/osconfig_*modulename*.log" and "/var/log/osconfig_*modulename*.bak" respectively | An OsConfigLog pointer | Called when the module starts to open the log
+CloseLog | An OsConfigLog pointer | None | Called when the module terminates to close the log
```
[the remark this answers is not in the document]
wupeka: Pointer to struct is opaque enough (the caller doesn't know what's inside, just that it's "a structure"), but guarantees that the caller will actually pass the pointer to the proper structure - using void* is error prone as seen in #825 , or even in this PR in src/common/commonutils/MountUtils.c : SetFileSystemMountingOption()

src/adapters/pnp/PnpAgent.c
```diff
@@ -105,7 +105,7 @@ static const char g_productInfoTemplate[] = "Azure OSConfig %d;%s "
     "\"product_vendor\"=\"%s\"&\"product_name\"=\"%s\")";
 static char g_productInfo[DEVICE_PRODUCT_INFO_SIZE] = {0};
 
-OSCONFIG_LOG_HANDLE GetLog()
+OsConfigLog* GetLog()
```
MariusNi, line 108: With the proposed way to keep existing names, files like this won't be changed and minimize the need for wider validation. For example, there is no CI for the PnP Agent where this code is. We had CI, we had an E2E test, that was then cut/removed. So making changes in anything that gets consumed here comes with a risk and  with need of extensive manual validation.

src/common/commonutils/CommonUtils.h
```diff
@@ -4,6 +4,7 @@
 #ifndef COMMONUTILS_H
 #define COMMONUTILS_H
 
+#include <Logging.h>
```
MariusNi, line 7: Aha! Now I remember. 

One reason we had the opaque handle as a void* was this: we do not want to link commonutils to the logging library. By including Logging.h we now do this lin. The idea is that commonutils can be linked only by itself, for example as it is for the UTs, and all those optional void* arguments are passed as null.

src/common/logging/Logging.h
```diff
@@ -16,7 +17,8 @@
 #define ARRAY_SIZE(a) (sizeof(a) / sizeof(a[0]))
 #endif
 
-typedef void* OSCONFIG_LOG_HANDLE;
+typedef struct OSCONFIG_LOG OSCONFIG_LOG;
+typedef OSCONFIG_LOG* OSCONFIG_LOG_HANDLE;
```
MariusNi, line 21: This is an elegant solution here, I like it. 

As you know, each OSConfig component (the PnP Agent, the Platform, each MM, etc) has its own log. In addition to these component specific logs, there are also other logs where we directly or indirectly contribute, such as: the MC agent logs, and syslog (for full Azure OSConfig and our two daemons there, their console traces are automatically routed to syslog). 

One single concern remaining:

How was this change tested to ensure to regression in relation to all of these logs? We do not have CI that validates the logs themselves, we only have some CI that uses these logs as output artifacts. How can be confirm that each component log, and each additional log, still functions like before, no regression?

I can only think of a manual pass with each of our management channels and module, with evidence collected in form logs. Are there other ways?

wupeka: I compiled the code without  (but including the fix for MountUtils) and with those changes, and the generated assembly does not differ except for C++ function signatures and the functions in Logging.c that were changed slightly - the rest of the output is instruction-for-instruction the same, that's good enough for me.

MariusNi: If you compared at compiled object level, that's good for me as well. Let's keep an eye on this though. Just merged 879, please rebase and then let's merge this one in as well so we can continue with logging improvements work.

MariusNi (approved):
:shipit:

AhmedBM (approved):
👌

robertschaedler3 (approved):

PR #884, opened 2025-02-25
Initall Log level proposal

kkanas:
## Description
Initial proposition to unite log levels so that they can be controlled using only one variable
Old log levels in various modules will be migrated to new ones and backward compatible API will be available

## Checklist


src/common/logging/Logging.h
```diff
@@ -43,6 +43,16 @@ bool IsDaemon(void);
     fprintf(GetLogFile(log), "[%s] [%s:%d]%s" format "\n", GetFormattedTime(), __SHORT_FILE__, __LINE__, loglevel, ## __VA_ARGS__);\
 }\
 
+enum OsConfigLogLevel {
+    LOG_LVL_SENSITIVE_DATA = 0,
+    LOG_LVL_DEBUG,
+    LOG_LVL_INFO,
+    LOG_LVL_ERROR,
+};
+
```
MariusNi, line 52: Can you show how do you set each of these combinations (possible today) via this proposed OsConfigLogLevel enum:

1. Full Logging enabled but no Command Logging.
2. Command Logging enabled but no Full Logging.

These two can be used independently of each other with full Azure OSConfig.

kkanas: Do we have to support this case? I omitted it for simplicity to discuss. 
The draft is still missing if (GetLogLevel() <= \#LOG_LEVEL_FROM_LOG_LINE) {LOG_STMT}
By that I mean that if the loglevel is lower or equal than log then configured the log level run log statement, otherwise don't log. 
I would like to get rid of IsFullLoggingEnabled() and IsCommandLoggingEnabled() and move the log statements to correct log level

kkanas: I added that support in https://github.com/Azure/azure-osconfig/pull/884/commits/887a7f1eae87511cada9619e32201d30e47d56f5  and how to use it https://github.com/Azure/azure-osconfig/pull/884/commits/0e3509bd91d7cc8f33357c6a82551ce55596b2e9 for src/common/commonutils/ConfigUtils.c 
Simillar transformation should be possible 
```
        if (IsFullLoggingEnabled())
        {
            OsConfigLogInfo(GetLog(), "Connection string: %s", connectionString);
        }
```
to
```
  OsConfigLogInfo(GetLog(), "Connection string: %s", connectionString);
```

Keep in mind this is just draft how I would do it, it's not tested etc

MariusNi:
Thanks for this proposal! This helped the discussion. The PR is now duplicated by merged PRs 883 and 885 and is being closed now.

PR #900, opened 2025-03-06
Rework ExecuteCommand - use single fork, read stdout/stderr via a pipe

wupeka:
## Description
Rework ExecuteCommand:
 - use only single fork() instead of three.
 - read stdout/stderr via a pipe and not via a temporary file, reducing IO overhead
 - reduce number of possible failure points
 - simplify the function flow and increase readability

## Checklist

### DIFF
--- a/src/common/commonutils/CommandUtils.c
+++ b/src/common/commonutils/CommandUtils.c
@@ -3,296 +3,245 @@
 
 #include "Internal.h"
 
-static void KillProcess(pid_t processId)
+#include <sys/select.h>
+
+static long MonotonicTime()
 {
-    fflush(NULL);
-    if (processId > 0)
+    struct timespec ts;
+    if (clock_gettime(CLOCK_MONOTONIC, &ts) == 0)
     {
-        kill(processId, SIGKILL);
+        return (long)(ts.tv_sec);
     }
+    return -1;
 }
 
-static int NormalizeStatus(int status)
+#define BUFFER_SIZE 1024
+
+int ExecuteCommand(void* context, const char* command, bool replaceEol, bool forJson, unsigned int maxTextResultBytes, unsigned int timeoutSeconds,
+    char** textResult, CommandCallback callback, OsConfigLogHandle log)
 {
-    int newStatus = status;
-    if ((ETIME != newStatus) && (ECANCELED != newStatus))
+    int workerPid = -1;
+    int pipefd[2] = {0};
+    long startTime = 0;
+
+    if (NULL == command)
     {
-        if (WIFEXITED(newStatus))
-        {
-            newStatus = WEXITSTATUS(newStatus);
-        }
-        else
-        {
-            newStatus = errno ? errno : -1;
-        }
-    }
-    return newStatus;
-}
-
-static int SystemCommand(void* context, const char* command, int timeoutSeconds, CommandCallback callback, OsConfigLogHandle log)
-{
-    const int callbackIntervalSeconds = 5; //seconds
-    const int defaultCommandTimeout = 60; //seconds
-
-    pid_t intermediateProcess = -1;
-    pid_t workerProcess = -1;
-    pid_t timerProcess = -1;
-    pid_t childProcess = -1;
-    int status = -1;
-    int intermediateStatus = -1;
-    int totalWaitSeconds = 0;
-    int timeout = (timeoutSeconds > 0) ? timeoutSeconds : defaultCommandTimeout;
-
-    bool mainProcessThread = (bool)(getpid() == gettid());
-
-    fflush(NULL);
-
-    if ((timeoutSeconds > 0) || (NULL != callback))
-    {
-        OsConfigLogDebug(log, "SystemCommand: executing command '%s' with timeout of %d seconds and%scancelation on %s thread",
-            command, timeout, (NULL == callback) ? " no " : " ", mainProcessThread ? "main process" : "worker");
-
-        // Fork an intermediate process to act as the parent for two more forked processes:
-        // one to actually execute the system command and the other to sleep and act as a timer.
-        // Whichever of these two children processes finishes first, that causes the other to be killed.
-        // The intermediate process exists as a parent for these (instead of the current process being
-        // the parent) to avoid collision with any other children processes.
-
-        if (0 == (intermediateProcess = fork()))
-        {
-            // Intermediate process
-
-            if (0 == (workerProcess = fork()))
-            {
-                // Worker process
-                status = execl("/bin/sh", "sh", "-c", command, (char*)NULL);
-                _exit(status);
-            }
-            else if (workerProcess < 0)
-            {
-                // Intermediate process
-                OsConfigLogError(log, "Failed forking process to execute command (errno: %d, '%s')", errno, strerror(errno));
-                status = -1;
-
-                // Kill the timer process if created and wait on it before exiting otherwise timer becomes a zombie process
-                if (timerProcess > 0)
-                {
-                    KillProcess(timerProcess);
-                    waitpid(timerProcess, &intermediateStatus, 0);
-                }
-                _exit(status);
-            }
-
-            if (0 == (timerProcess = fork()))
-            {
-                // Timer process
-                status = ETIME;
-                if (NULL == callback)
-                {
-                    sleep(timeout);
-                }
-                else
-                {
-                    while (totalWaitSeconds < timeout)
-                    {
-                        // If the callback returns non zero, cancel the command
-                        if (0 != callback(context))
-                        {
-                            status = ECANCELED;
-                            break;
-                        }
-                        sleep(callbackIntervalSeconds);
-                        totalWaitSeconds += callbackIntervalSeconds;
-                    }
-                }
-                _exit(status);
-            }
-            else if (timerProcess < 0)
-            {
-                // Intermediate process
-                OsConfigLogError(log, "Failed forking timer process (errno: %d, '%s')", errno, strerror(errno));
-                status = -1;
-
-                // Kill the worker process if created and wait on it before exiting otherwise worker becomes a zombie process
-                if (workerProcess > 0)
-                {
-                    KillProcess(workerProcess);
-                    waitpid(workerProcess, &intermediateStatus, 0);
-                }
-                _exit(status);
-            }
-
-            // Wait on the child process (worker or timer) that finishes first
-            childProcess = waitpid(0, &status, 0);
-            status = NormalizeStatus(status);
-            if (childProcess == workerProcess)
-            {
-                OsConfigLogDebug(log, "Command execution returning %d", status);
-                KillProcess(timerProcess);
-            }
-            else
-            {
-                // Timer process is done, kill the timed out worker process
-                OsConfigLogDebug(log, "Command timed out or it was canceled, command process killed (%d)", status);
-                KillProcess(workerProcess);
-            }
-
-            // Wait on the remaining child (either worker or timer) and exit (the intermediate process)
-            waitpid(0, &intermediateStatus, 0);
-            _exit(status);
-        }
-        else if (intermediateProcess > 0)
-        {
-            // Wait on the intermediate process to finish
-            waitpid(intermediateProcess, &status, 0);
-        }
-        else
-        {
-            status = -1;
-            OsConfigLogError(log, "Failed forking intermediate process (errno: %d, '%s')", errno, strerror(errno));
-        }
-    }
-    else
-    {
-        OsConfigLogDebug(log, "SystemCommand: executing command '%s' without timeout or cancelation on %s thread", command, mainProcessThread ? "main process" : "worker");
-        if (0 == (workerProcess = fork()))
-        {
-            // Worker process
-            status = execl("/bin/sh", "sh", "-c", command, (char*)NULL);
-            _exit(status);
-        }
-        else if (workerProcess > 0)
-        {
-            // Wait on the worker process to terminate
-            waitpid(workerProcess, &status, 0);
-        }
-        else
-        {
-            // If our fork fails, try system(), if that also fails then the call fails
-            OsConfigLogError(log, "Failed forking process to execute command (errno: %d, '%s'), attempting system", errno, strerror(errno));
-            status = system(command);
-        }
-    }
-
-    status = NormalizeStatus(status);
-    OsConfigLogDebug(log, "SystemCommand: command '%s' returning %d", command, status);
-
-    return status;
-}
-
-#define MAX_COMMAND_RESULT_FILE_NAME 100
-
-int ExecuteCommand(void* context, const char* command, bool replaceEol, bool forJson, unsigned int maxTextResultBytes, unsigned int timeoutSeconds, char** textResult, CommandCallback callback, OsConfigLogHandle log)
-{
-    const char commandTextResultFileTemplate[] = "/tmp/~OSConfig.TextResult%u";
-    const char commandSeparator[] = " > ";
-    const char commandTerminator[] = " 2>&1";
-
-    int status = -1;
-    FILE* resultsFile = NULL;
-    int fileSize = 0;
-    int next = 0;
-    int i = 0;
-    char* commandLine = NULL;
-    size_t commandLineLength = 0;
-    size_t maximumCommandLine = 0;
-    char commandTextResultFile[MAX_COMMAND_RESULT_FILE_NAME] = {0};
-    bool wrappedCommand = false;
-
-    if ((NULL == command) || (0 == system(NULL)))
-    {
-        OsConfigLogDebug(log, "Cannot run command '%s'", command);
+        OsConfigLogDebug(log, "Command cannot be NULL");
         return -1;
     }
-
-    commandLineLength = strlen(command);
-    wrappedCommand = (('(' == command[0]) || (')' == command[commandLineLength]));
-
-    // Append a random number to the results file to prevent parallel commands overwriting each other results
-    snprintf(commandTextResultFile, sizeof(commandTextResultFile), commandTextResultFileTemplate, rand());
-
-    commandLineLength += strlen(commandSeparator) + strlen(commandTextResultFile) + strlen(commandTerminator) + (wrappedCommand ? 0 : 2) + 1;
-
-    maximumCommandLine = (size_t)sysconf(_SC_ARG_MAX);
-    if (commandLineLength > maximumCommandLine)
+    if (strlen(command) > (size_t)sysconf(_SC_ARG_MAX))
     {
-        OsConfigLogDebug(log, "Cannot run command '%s', command too long (%u), ARG_MAX: %u", command, (unsigned)commandLineLength, (unsigned)maximumCommandLine);
+        OsConfigLogError(log, "Command too long: %s", command);
         return E2BIG;
     }
 
-    commandLine = (char*)malloc(commandLineLength);
-    if (NULL == commandLine)
+    // Create a pipe, then fork. Forked process duplicates the write pipe end to stdout and stderr,
+    // then execs the shell with the given command. The main process uses select() with a timeout.
+    // to read from the pipe, and keep track of both command  timeout and callbacks. The read loop
+    // ends when the read() returns EOF or when the command times out or the callback returns a non-zero value.
+    // The read loop also replaces the EOL characters with spaces if requested, and replaces all special
+    // characters with spaces if requested. The output is returned in the textResult buffer, which is
+    // allocated by this function. The caller is responsible for freeing the buffer when done.
+
+    startTime = MonotonicTime();
+    if (startTime < 0)
     {
-        OsConfigLogError(log, "Cannot run command '%s', cannot allocate %u bytes for command, out of memory", command, (unsigned)commandLineLength);
-        return ENOMEM;
+        OsConfigLogError(log, "Cannot get time for command '%s', clock_gettime() failed with %d (%s)", command, errno, strerror(errno));
+        return errno;
     }
 
-    snprintf(commandLine, commandLineLength, wrappedCommand ? "%s%s%s%s" : "(%s)%s%s%s", command, commandSeparator, commandTextResultFile, commandTerminator);
-
-    // Execute the command with the requested timeout: error ETIME (62) means the command timed out
-    status = SystemCommand(context, commandLine, timeoutSeconds, callback, log);
-
-    free(commandLine);
-
-    // Read the text result from the output of the command, if any, whether command succeeded or failed
-    if (NULL != textResult)
+    if (0 != pipe(pipefd))
     {
-        *textResult = NULL;
-        resultsFile = fopen(commandTextResultFile, "r");
-        if (resultsFile)
+        OsConfigLogError(log, "Cannot create pipe for command '%s', pipe() failed with %d (%s)", command, errno, strerror(errno));
+        return errno;
+    }
+
+    workerPid = fork();
+    if (workerPid < 0)
+    {
+        OsConfigLogError(log, "Cannot fork for command '%s', fork() failed with %d (%s)", command, errno, strerror(errno));
+        close(pipefd[0]);
+        close(pipefd[1]);
+        return errno;
+    }
+
+    if (0 == workerPid)
+    {
+        // Child process
+        close(pipefd[0]);
+        if (STDOUT_FILENO != dup2(pipefd[1], STDOUT_FILENO))
         {
-            fseek(resultsFile, 0, SEEK_END);
-            fileSize = ftell(resultsFile);
-            fseek(resultsFile, 0, SEEK_SET);
+            exit(errno);
+        }
+        if (STDERR_FILENO != dup2(pipefd[1], STDERR_FILENO))
+        {
+            exit(errno);
+        }
+        execl("/bin/sh", "sh", "-c", command, (char*)NULL);
+        // If execl() fails, exit with the error code
+        exit(errno);
+    }
+    else
+    {
+        // Main process
+        const int defaultCommandTimeout = 60;  // seconds
+        const int callbackIntervalSeconds = 5; // seconds
+        long lastCallbackTime = 0;
+        int status = -1;
+        int childStatus = 0;
+        unsigned int outputBufferPos = 0;
+        unsigned int outputBufferSize = 0;
 
-            if (fileSize > 0)
+        close(pipefd[1]);
+
+        if ((NULL != callback) && (timeoutSeconds == 0))
+        {
+            timeoutSeconds = defaultCommandTimeout;
+        }
+
+        for (;;)
+        {
+            const struct timeval selectInterval = {0, 100 * 1000}; // 100 ms, accuracy of timeouts.
+            struct timeval tv;
+            int bytesRead = 0;
+            int inputBufferPos = 0;
+            long currentTime = 0;
+            char buffer[BUFFER_SIZE] = {0};
+            fd_set fdset;
+            int ret = 0;
+            char* tmp = NULL;
+
+            FD_ZERO(&fdset);
+            FD_SET(pipefd[0], &fdset);
+
+            tv = selectInterval;
+            ret = select(pipefd[0] + 1, &fdset, NULL, NULL, &tv);
+            if (ret < 0)
             {
-                // Truncate to desired maximum, if any
-                if ((maxTextResultBytes > 0) && (((size_t)fileSize + 1) > maxTextResultBytes))
+                if (EINTR == errno)
                 {
-                    fileSize = (maxTextResultBytes > 1) ? (maxTextResultBytes - 1) : 0;
+                    continue;
                 }
-
-                *textResult = (char*)malloc(fileSize + 1);
-                if (NULL != *textResult)
-                {
-                    memset(*textResult, 0, fileSize + 1);
-                    for (i = 0; i < fileSize; i++)
-                    {
-                        next = fgetc(resultsFile);
-
-                        if (EOF == next)
-                        {
-                            break;
-                        }
-
-                        // Copy the data. Following characters are replaced with spaces:
-                        // all special characters from 0x00 to 0x1F except 0x0A (LF) when replaceEol is false
-                        // plus 0x22 (") and 0x5C (\) characters that break the JSON envelope when forJson is true
-                        if ((replaceEol && (EOL == next)) || ((next < 0x20) && (EOL != next)) || (0x7F == next) || (forJson && (('"' == next) || ('\\' == next))))
-                        {
-                            (*textResult)[i] = ' ';
-                        }
-                        else
-                        {
-                            (*textResult)[i] = (char)next;
-                        }
-                    }
-                }
+                OsConfigLogError(log, "Error doing select for command '%s', select() failed with %d (%s)", command, errno, strerror(errno));
+                status = errno;
+                break;
             }
 
-            fclose(resultsFile);
+            currentTime = MonotonicTime();
+            if (currentTime < 0)
+            {
+                OsConfigLogError(log, "Error getting time for command '%s', clock_gettime() failed with %d (%s)", command, errno, strerror(errno));
+                status = errno;
+                break;
+            }
+            if ((timeoutSeconds > 0) && (currentTime - startTime >= timeoutSeconds))
+            {
+                OsConfigLogError(log, "Timeout reading from pipe for command '%s', %d seconds", command, (int)(currentTime - startTime));
+                status = ETIME;
+                break;
+            }
+            if ((NULL != callback) && (currentTime - lastCallbackTime >= callbackIntervalSeconds))
+            {
+                if (0 != callback(context))
+                {
+                    OsConfigLogError(log, "Canceled reading from pipe for command '%s'", command);
+                    status = ECANCELED;
+                    break;
+                }
+                lastCallbackTime = currentTime;
+            }
+
+            if (!FD_ISSET(pipefd[0], &fdset))
+            {
+                // It was a timeout, nothing to read, loop.
+                continue;
+            }
+
+            bytesRead = read(pipefd[0], buffer, BUFFER_SIZE);
+            if (bytesRead == 0)
+            {
+                // Child closed the pipe, we are done.
+                status = 0;
+                break;
+            }
+            if (bytesRead < 0)
+            {
+                if (EINTR == errno)
+                {
+                    continue;
+                }
+                OsConfigLogError(log, "Error reading from pipe for command '%s', read() failed with %d (%s)", command, errno, strerror(errno));
+                status = errno;
+                break;
+            }
+
+            if (((maxTextResultBytes > 0) && (outputBufferPos == maxTextResultBytes)) || textResult == NULL)
+            {
+                // We don't want any more data, loop to read the rest of the output.
+                continue;
+            }
+
+            outputBufferSize = bytesRead + outputBufferPos;
+            if ((maxTextResultBytes > 0) && (outputBufferSize > maxTextResultBytes - 1))
+            {
+                outputBufferSize = maxTextResultBytes - 1;
+            }
+
+            tmp = realloc(*textResult, outputBufferSize + 1);
+            if (NULL == tmp)
+            {
+                OsConfigLogError(log, "Cannot allocate buffer for command '%s'", command);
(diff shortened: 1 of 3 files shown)


src/common/commonutils/CommonUtils.h
```diff
@@ -54,7 +54,6 @@ int RenameFile(const char* original, const char* target, OsConfigLogHandle log);
 
 typedef int(*CommandCallback)(void* context);
 
-// If called from the main process thread the timeoutSeconds and callback arguments are ignored
```
MariusNi, line 57: If I remember correctly, we had this for IoT Hub, we had there a quite nasty bug that led to this.

src/common/commonutils/CommandUtils.c
```diff
@@ -29,270 +31,194 @@ static int NormalizeStatus(int status)
     return newStatus;
 }
 
-static int SystemCommand(void* context, const char* command, int timeoutSeconds, CommandCallback callback, OsConfigLogHandle log)
```
MariusNi, line 32: >SystemCommand [](http://example.com/codeflow?start=11&length=13)

We could keep this function and make it hold the 'main process' code

src/common/commonutils/CommandUtils.c
```diff
@@ -29,270 +31,194 @@ static int NormalizeStatus(int status)
...
+        close(pipefd[0]);
+        close(pipefd[1]);
+        return -1;
+    }
 
-        // Fork an intermediate process to act as the parent for two more forked processes:
-        // one to actually execute the system command and the other to sleep and act as a timer.
-        // Whichever of these two children processes finishes first, that causes the other to be killed.
-        // The intermediate process exists as a parent for these (instead of the current process being
-        // the parent) to avoid collision with any other children processes.
```
MariusNi, line 59: Can you please add similarly detailed comments that explain how the approach with pipes works and why done this way?

src/common/commonutils/CommandUtils.c
```diff
@@ -29,270 +31,194 @@ static int NormalizeStatus(int status)
...
-    }
-
-    commandLine = (char*)malloc(commandLineLength);
-    if (NULL == commandLine)
-    {
-        OsConfigLogError(log, "Cannot run command '%s', cannot allocate %u bytes for command, out of memory", command, (unsigned)commandLineLength);
-        return ENOMEM;
-    }
-
-    snprintf(commandLine, commandLineLength, wrappedCommand ? "%s%s%s%s" : "(%s)%s%s%s", command, commandSeparator, commandTextResultFile, commandTerminator);
```
MariusNi, line 230: >wrappedCommand [](http://example.com/codeflow?start=45&length=14)

Is this refactor (rework) keeping support for wrapped commands? Commands which by their syntax could interfere with our command execution command?

src/common/commonutils/CommandUtils.c
```diff
@@ -29,270 +31,194 @@ static int NormalizeStatus(int status)
...
-
-    commandLineLength = strlen(command);
-    wrappedCommand = (('(' == command[0]) || (')' == command[commandLineLength]));
-
-    // Append a random number to the results file to prevent parallel commands overwriting each other results
-    snprintf(commandTextResultFile, sizeof(commandTextResultFile), commandTextResultFileTemplate, rand());
-
-    commandLineLength += strlen(commandSeparator) + strlen(commandTextResultFile) + strlen(commandTerminator) + (wrappedCommand ? 0 : 2) + 1;
-
-    maximumCommandLine = (size_t)sysconf(_SC_ARG_MAX);
```
MariusNi, line 216: >_SC_ARG_MAX [](http://example.com/codeflow?start=41&length=11)

Why aren't we any longer needing to watch for __SC_ARG_MAX?

[https://www.man7.org/linux/man-pages/man3/sysconf.3.html](https://www.man7.org/linux/man-pages/man3/sysconf.3.html)

'ARG_MAX - _SC_ARG_MAX
              The maximum length of the arguments to the exec(3) family
              of functions.  Must not be less than _POSIX_ARG_MAX (4096).'

src/common/commonutils/CommandUtils.c
```diff
@@ -29,270 +31,194 @@ static int NormalizeStatus(int status)
...
-    return status;
-}
-
-#define MAX_COMMAND_RESULT_FILE_NAME 100
-
-int ExecuteCommand(void* context, const char* command, bool replaceEol, bool forJson, unsigned int maxTextResultBytes, unsigned int timeoutSeconds, char** textResult, CommandCallback callback, OsConfigLogHandle log)
-{
-    const char commandTextResultFileTemplate[] = "/tmp/~OSConfig.TextResult%u";
-    const char commandSeparator[] = " > ";
-    const char commandTerminator[] = " 2>&1";
```
MariusNi, line 189: >const char commandTerminator[] = " 2>&1"; [](http://example.com/codeflow?start=4&length=41)

We needed this to capture error output from the executed command and return as text results. How are we achieving the same with the pipes?

src/common/commonutils/CommandUtils.c
wupeka: Done

src/common/commonutils/CommandUtils.c
wupeka: We don't need it as we're capturing both stderr and stdout with the pipe.

src/common/commonutils/CommandUtils.c
wupeka: execl() will fail with E2BIG, so this check is reduntant.

src/common/commonutils/CommandUtils.c
wupeka: Previously the commands were called in subshell (if they weren't wrapped in () by the caller, hence this check) to make sure that stdout/stderr redirection works in shell. That's no longer necessary.

src/common/commonutils/CommonUtils.h
wupeka: This comment was incorrect even with the old code - timeoutSeconds and callback were never ignored, guessing the history from  the code that's why the "intermediate" process was launched.

src/common/commonutils/CommandUtils.c
```diff
@@ -29,270 +31,194 @@ static int NormalizeStatus(int status)
...
+        OsConfigLogError(log, "Cannot create pipe for command '%s' - %s", command, strerror(errno));
+        return -1;
+    }
 
-    if ((timeoutSeconds > 0) || (NULL != callback))
+    workerPid = fork();
+    if (workerPid < 0)
     {
-        OsConfigLogDebug(log, "SystemCommand: executing command '%s' with timeout of %d seconds and%scancelation on %s thread",
-            command, timeout, (NULL == callback) ? " no " : " ", mainProcessThread ? "main process" : "worker");
```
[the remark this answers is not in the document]
wupeka: What is "this" here? The mainProcessThread is actually used only for logging, more explanation in the comment for removal of the comment in CommonUtils.h.

src/common/commonutils/CommandUtils.c
```diff
@@ -29,270 +31,194 @@ static int NormalizeStatus(int status)
...
-        else if (workerProcess > 0)
-        {
-            // Wait on the worker process to terminate
-            waitpid(workerProcess, &status, 0);
-        }
-        else
-        {
-            // If our fork fails, try system(), if that also fails then the call fails
-            OsConfigLogError(log, "Failed forking process to execute command (errno: %d, '%s'), attempting system", errno, strerror(errno));
-            status = system(command);
```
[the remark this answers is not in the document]
MariusNi: If there is a chance, even remote, that any of the piping calls will fail, system might still be offer a worth backup try before bailing out with error. I'll let you decide on this, just curious.

src/common/commonutils/CommandUtils.c
MariusNi: Can we add a test case for this and confirm the E2BIG with it?

src/common/commonutils/CommandUtils.c
MariusNi: Awesome!

src/common/commonutils/CommandUtils.c
```diff
@@ -29,270 +31,200 @@ static int NormalizeStatus(int status)
...
+        {
+            exit(-1);
+        }
+        execl("/bin/sh", "sh", "-c", command, (char*)NULL);
+        // If execl() fails, exit with the error code
+        exit(errno);
+    }
+    else
+    {
+        // Main process
```
MariusNi, line 85: >// Main process [](http://example.com/codeflow?start=8&length=15)

This comment could be misleading. The main process is the one that calls this function. The forked process is the worker process that executes the command. Which one is here? May want to clarify in comment.

src/common/commonutils/CommandUtils.c
```diff
@@ -29,270 +31,200 @@ static int NormalizeStatus(int status)
...
+        int status = -1;
+        int childStatus = 0;
+        int bytesRead = 0;
+        int inputBufferPos = 0;
+        unsigned int outputBufferPos = 0;
+        unsigned int outputBufferSize = 0;
+        time_t startTime = time(NULL);
+        time_t lastCallbackTime = 0;
+        time_t currentTime = 0;
+        char buffer[BUFFER_SIZE];
```
MariusNi, line 97: >char buffer[BUFFER_SIZE]; [](http://example.com/codeflow?start=8&length=25)

nit: initialize with ={0}.

src/common/commonutils/CommandUtils.c
```diff
@@ -29,270 +31,200 @@ static int NormalizeStatus(int status)
...
+
+        close(pipefd[1]);
+
+        if ((NULL != callback) && (timeoutSeconds == 0))
+        {
+            timeoutSeconds = defaultCommandTimeout;
+        }
 
-        if (0 == (intermediateProcess = fork()))
+        for (;;)
```
MariusNi, line 106: >for (;;) [](http://example.com/codeflow?start=8&length=8)

Why not a 'while' or 'do' loop if the 'for' arguments are not used?

src/common/commonutils/CommandUtils.c
```diff
@@ -29,270 +31,200 @@ static int NormalizeStatus(int status)
...
         {
-            // Intermediate process
+            fd_set fdset;
+            int ret = 0;
+            char* tmp = NULL;
```
MariusNi, line 110: Can we please declare all locals at beginning of the function block?

src/common/commonutils/CommandUtils.c
```diff
@@ -29,270 +31,200 @@ static int NormalizeStatus(int status)
...
-                status = -1;
+            FD_ZERO(&fdset);
+            FD_SET(pipefd[0], &fdset);
 
-                // Kill the timer process if created and wait on it before exiting otherwise timer becomes a zombie process
-                if (timerProcess > 0)
-                {
-                    KillProcess(timerProcess);
-                    waitpid(timerProcess, &intermediateStatus, 0);
+            ret = select(pipefd[0] + 1, &fdset, NULL, NULL, &(struct timeval){.tv_sec = 0, .tv_usec = 500 * 1000});
```
MariusNi, line 115: >500 * 1000 [](http://example.com/codeflow?start=102&length=10)

What's this 'magic number' and can we have a constant declared for it?

src/common/commonutils/CommandUtils.c
```diff
@@ -29,270 +31,200 @@ static int NormalizeStatus(int status)
...
+            if (ret < 0)
+            {
+                if (EINTR == errno) {
+                    continue;
                 }
-                _exit(status);
+                OsConfigLogError(log, "Error doing select for command '%s' - %s", command, strerror(errno));
```
MariusNi, line 121: >'%s' - %s", command, strerror(errno)); [](http://example.com/codeflow?start=70&length=38)

nit: also log the errno decimal value and do it in way consistent that rest of the code in project does it.

src/common/commonutils/CommandUtils.c
```diff
@@ -29,270 +31,200 @@ static int NormalizeStatus(int status)
...
+                status = -1;
+                break;
             }
 
-            if (0 == (timerProcess = fork()))
+            currentTime = time(NULL);
+            if ((timeoutSeconds > 0) && (currentTime - startTime >= timeoutSeconds))
             {
-                // Timer process
+                OsConfigLogError(log, "Timeout reading from pipe for command '%s'", command);
```
MariusNi, line 129: >"Timeout reading from pipe for command '%s'", command); [](http://example.com/codeflow?start=38&length=55)

Logging should tell the user when the timeout for the command execution is exceeded, not just reading from pipe. Do we have this?

src/common/commonutils/CommandUtils.c
```diff
@@ -29,270 +31,200 @@ static int NormalizeStatus(int status)
...
-                KillProcess(workerProcess);
+                // Child closed the pipe, we are done.
+                status = 0;
+                break;
+            }
+            if (bytesRead < 0)
+            {
+                if (EINTR == errno) {
+                    continue;
+                }
```
MariusNi, line 161: Here and elsewhere, please use curly braces {} consistent with existing code. 

Also, here local var status is used. Can we please use this everywhere in the whole ExecuteCommand function?

src/common/commonutils/CommandUtils.c
```diff
@@ -29,270 +31,200 @@ static int NormalizeStatus(int status)
...
-    // Execute the command with the requested timeout: error ETIME (62) means the command timed out
-    status = SystemCommand(context, commandLine, timeoutSeconds, callback, log);
+            if (((maxTextResultBytes > 0) && (outputBufferPos == maxTextResultBytes)) || textResult == NULL)
+            {
+                // We don't want any more data, loop to read the rest of the output.
+                continue;
+            }
 
-    free(commandLine);
+            outputBufferSize = bytesRead + outputBufferPos;
```
MariusNi, line 173: >bytesRead + outputBufferPos; [](http://example.com/codeflow?start=31&length=28)

Can this overflow?

src/common/commonutils/CommandUtils.c
```diff
@@ -29,270 +31,200 @@ static int NormalizeStatus(int status)
...
-            fseek(resultsFile, 0, SEEK_END);
-            fileSize = ftell(resultsFile);
-            fseek(resultsFile, 0, SEEK_SET);
+            tmp = realloc(*textResult, outputBufferSize + 1);
+            if (NULL == tmp)
+            {
+                OsConfigLogError(log, "Cannot allocate buffer for command '%s'", command);
+                status = ENOMEM;
+                free(*textResult);
+                *textResult = NULL;
```
MariusNi, line 185: Please use the FREE_MEMORY macro

src/common/commonutils/CommandUtils.c
MariusNi: Sorry, where? I am not seeing them.... Did you forget to push a commit or I hit a glitch at my end?

src/common/commonutils/CommandUtils.c
```diff
@@ -29,270 +31,200 @@ static int NormalizeStatus(int status)
     return newStatus;
 }
 
-static int SystemCommand(void* context, const char* command, int timeoutSeconds, CommandCallback callback, OsConfigLogHandle log)
-{
-    const int callbackIntervalSeconds = 5; //seconds
-    const int defaultCommandTimeout = 60; //seconds
+#define BUFFER_SIZE 1024
```
MariusNi, line 34: >1024 [](http://example.com/codeflow?start=20&length=4)

Is this the most optimal buffer size, or how was the 1KB size chosen?

src/common/commonutils/CommandUtils.c
MariusNi: Saw it now, lines 47-50. Thanks

src/common/commonutils/CommandUtils.c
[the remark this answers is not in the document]
MariusNi: The intermediate process part. Got from your other response that this intermediate process is no longer necessary. Correct?

MariusNi (approved):
Thanks for this rework/refactor which will increase the efficiency of running commands! 

Sorry for the many comments, they are all mostly nit cosmetic and style ones. My intent was, if possible, to have the new version fully polished as part of this PR. However, if you do not have time or want to deal with all these non-important details here, we can follow-up subsequent PRs to further clean and polish this, the main change coming in with this PR. 

LGTM overall and signed off. 

(and added a few more nits after, sorry about that, signoff still remains however)

src/common/tests/CommonUtilsUT.cpp
```diff
@@ -642,35 +642,6 @@ TEST_F(CommonUtilsTest, ExecuteLongCommand)
...
-        command[commandLength - 1] = 0;
-
-        EXPECT_EQ(E2BIG, ExecuteCommand(nullptr, command, true, true, 0, 0, &textResult, nullptr, nullptr));
-        EXPECT_EQ(nullptr, textResult);
-
-        free(command);
-    }
-
-    FREE_MEMORY(textResult);
-}
```
MariusNi, line 672: Let's keep this unit-test. We know to expect E2BIG when the command is too long. We should keep UT-ing this.

Why:

E2BIG is not a normal command failure. It's our failure because we passed in a too long command. It's something that we need to watch for and if this ever happens at run-time, we need to treat this case as a true error case for us and a code bug, and not just a normal command result.

Personally I would also put back the validation against _SC_ARG_MAX early in the ExecuteCommand call, validate, log error and fail early, before forking and rest.

src/common/commonutils/CommonUtils.h
MariusNi: You cannot see the full history here. Recorded 'history' for this project is only from the time when we went public. 

There were far more versions of ExecuteCommand before that. That was the time when our primary management channel was IoT Hub and ExecuteCommand was primarily used by CommandRunner for the Digital Twins work. During that, we pass in the request to execute command on one thread, and we report the command results async and on a separate thread. The code is as-is because we had to shield in depth for all the corner cases in that scenario, where we run as an extension of the Azure PnP C SDK.

src/common/commonutils/CommandUtils.c
```diff
@@ -29,270 +31,200 @@ static int NormalizeStatus(int status)
     return newStatus;
 }
 
-static int SystemCommand(void* context, const char* command, int timeoutSeconds, CommandCallback callback, OsConfigLogHandle log)
-{
-    const int callbackIntervalSeconds = 5; //seconds
-    const int defaultCommandTimeout = 60; //seconds
```
MariusNi, line 35: What's the callback frequency with this new version of the command? Do we keep the previous 5 seconds?

MariusNi, line 35: >defaultCommandTimeout [](http://example.com/codeflow?start=14&length=21)

Do we still have the default timeout of 60 seconds with the new version?

src/common/commonutils/CommandUtils.c
```diff
@@ -29,270 +31,200 @@ static int NormalizeStatus(int status)
...
+    int pipefd[2] = {0};
 
-    bool mainProcessThread = (bool)(getpid() == gettid());
+    if (NULL == command)
+    {
+        OsConfigLogDebug(log, "Command cannot be NULL");
+        return -1;
+    }
+    // Create a pipe, then fork. Forked process dup2()s the write pipe end to stdout and stderr,
+    // then execl()s the shell with the given command. The main process uses timeoutable select()
```
MariusNi, line 48: >timeoutable [](http://example.com/codeflow?start=77&length=11)

Typo? 'timeoutable' (what's that?)

src/common/commonutils/CommandUtils.c
```diff
@@ -29,270 +31,200 @@ static int NormalizeStatus(int status)
...
+    int workerPid = -1;
+    int pipefd[2] = {0};
 
-    bool mainProcessThread = (bool)(getpid() == gettid());
+    if (NULL == command)
+    {
+        OsConfigLogDebug(log, "Command cannot be NULL");
+        return -1;
+    }
+    // Create a pipe, then fork. Forked process dup2()s the write pipe end to stdout and stderr,
```
MariusNi, line 47: >dup2()s [](http://example.com/codeflow?start=48&length=7)

What exactly means 'dup2()s'?

You mean, 'duplicates the write pipe by calling dup2()'?

Can you please make this overall comment easier to be understood by all readers, including the ones who are not familiar with pipes, forks, etc.?

src/common/commonutils/CommandUtils.c
```diff
@@ -29,270 +31,200 @@ static int NormalizeStatus(int status)
...
+    // then execl()s the shell with the given command. The main process uses timeoutable select()
+    // to read from the pipe, and keep track of both command  timeout and callbacks. The read loop
+    // ends when the read() returns EOF.
 
-    fflush(NULL);
+    if (0 != pipe(pipefd))
+    {
+        OsConfigLogError(log, "Cannot create pipe for command '%s' - %s", command, strerror(errno));
```
MariusNi, line 54: >command '%s' - %s" [](http://example.com/codeflow?start=54&length=18)

nit This ( '-')is a log pattern that is not used in the existing code of this project. 

Instead of this:

```
OsConfigLogError(log, "Cannot create pipe for command '%s' - %s", command, strerror(errno));
````

Can this new version keep the old version's trace style, for example:

```
OsConfigLogError(log, "Cannot create pipe for command '%s', pipe() failed with %d ('%s')", command, errno, strerror(errno));
````
Here and elsewhere.

File: src/common/commonutils/CommandUtils.c [hunk 25]
```diff
@@ -29,270 +31,200 @@ static int NormalizeStatus(int status)
...
+    {
+        // Child process
+        close(pipefd[0]);
+        if (STDOUT_FILENO != dup2(pipefd[1], STDOUT_FILENO))
+        {
+            exit(-1);
+        }
+        if (STDERR_FILENO != dup2(pipefd[1], STDERR_FILENO))
+        {
+            exit(-1);
```
MariusNi, line 77: >-1 [](http://example.com/codeflow?start=17&length=2)

Instead of failing with -1 (in several places), let's use  defined error codes such as EPERM, EACCESS, EPIPE, etc? In each case the one that's most appropriate for that error.

src/common/commonutils/CommandUtils.c
```diff
@@ -29,270 +31,200 @@ static int NormalizeStatus(int status)
...
+        }
+        execl("/bin/sh", "sh", "-c", command, (char*)NULL);
+        // If execl() fails, exit with the error code
+        exit(errno);
```
MariusNi, line 81: >errno [](http://example.com/codeflow?start=13&length=5)

Are we sure errno is set here? What if it's 0?

Here and elsewhere, let's use errno if non-zero and if zero, use a predefined error code

src/common/commonutils/CommandUtils.c
```diff
@@ -29,270 +31,200 @@ static int NormalizeStatus(int status)
...
-        // Whichever of these two children processes finishes first, that causes the other to be killed.
-        // The intermediate process exists as a parent for these (instead of the current process being
-        // the parent) to avoid collision with any other children processes.
+    if (0 == workerPid)
+    {
+        // Child process
+        close(pipefd[0]);
+        if (STDOUT_FILENO != dup2(pipefd[1], STDOUT_FILENO))
+        {
+            exit(-1);
```
MariusNi, line 73: >exit(-1); [](http://example.com/codeflow?start=12&length=9)

We need to error trace all these failures, here and elsewhere

src/common/commonutils/CommandUtils.c
```diff
@@ -29,270 +31,200 @@ static int NormalizeStatus(int status)
...
+        int bytesRead = 0;
+        int inputBufferPos = 0;
+        unsigned int outputBufferPos = 0;
+        unsigned int outputBufferSize = 0;
+        time_t startTime = time(NULL);
+        time_t lastCallbackTime = 0;
+        time_t currentTime = 0;
+        char buffer[BUFFER_SIZE];
+
+        close(pipefd[1]);
```
MariusNi, line 99: >pipefd[1] [](http://example.com/codeflow?start=14&length=9)

Can pipefd[1] be undefined here?

src/common/commonutils/CommandUtils.c
MariusNi, line 115: >{.tv_sec = 0, .tv_usec = 500 * 1000}); [](http://example.com/codeflow?start=77&length=38)

Please add a comment that explains how this works.

src/common/commonutils/CommandUtils.c
```diff
@@ -29,270 +31,200 @@ static int NormalizeStatus(int status)
...
                 }
-                _exit(status);
+                OsConfigLogError(log, "Error doing select for command '%s' - %s", command, strerror(errno));
+                status = -1;
+                break;
             }
 
-            if (0 == (timerProcess = fork()))
+            currentTime = time(NULL);
+            if ((timeoutSeconds > 0) && (currentTime - startTime >= timeoutSeconds))
```
MariusNi, line 127: >currentTime - startTime  [](http://example.com/codeflow?start=41&length=24)

Can currentTime be 'older' (greater) than startTime?

src/common/commonutils/CommandUtils.c
```diff
@@ -29,270 +31,200 @@ static int NormalizeStatus(int status)
...
+            for (inputBufferPos = 0; (inputBufferPos < bytesRead) && (outputBufferPos < outputBufferSize); inputBufferPos++, outputBufferPos++)
             {
-                // Truncate to desired maximum, if any
-                if ((maxTextResultBytes > 0) && (((size_t)fileSize + 1) > maxTextResultBytes))
+                // Copy the data. Following characters are replaced with spaces:
+                // all special characters from 0x00 to 0x1F except 0x0A (LF) when replaceEol is false
+                // plus 0x22 (") and 0x5C (\) characters that break the JSON envelope when forJson is true
+                const char c = buffer[inputBufferPos];
+                if ((replaceEol && (EOL == c)) || ((c < 0x20) && (EOL != c)) || (0x7F == c) ||
+                    (forJson && (('"' == c) || ('\\' == c))))
```
MariusNi, line 197: nit: this fits on one single line, why break in two?

src/common/commonutils/CommandUtils.c
```diff
@@ -29,270 +31,200 @@ static int NormalizeStatus(int status)
...
 
-            if (fileSize > 0)
+            for (inputBufferPos = 0; (inputBufferPos < bytesRead) && (outputBufferPos < outputBufferSize); inputBufferPos++, outputBufferPos++)
             {
-                // Truncate to desired maximum, if any
-                if ((maxTextResultBytes > 0) && (((size_t)fileSize + 1) > maxTextResultBytes))
+                // Copy the data. Following characters are replaced with spaces:
+                // all special characters from 0x00 to 0x1F except 0x0A (LF) when replaceEol is false
+                // plus 0x22 (") and 0x5C (\) characters that break the JSON envelope when forJson is true
+                const char c = buffer[inputBufferPos];
```
MariusNi, line 195: >const char c = buffer[inputBufferPos]; [](http://example.com/codeflow?start=16&length=38)

Move local variable definition at beginning of function, C-style. Please correct this everywhere.

In general, can the new variant please more closely follow the original style for names, comments, traces, etc?

src/common/commonutils/CommandUtils.c
wupeka: "The exec() functions return only if an error has occurred.  The return value is -1, and errno is set to indicate the error."

src/common/commonutils/CommandUtils.c
wupeka: As you stated - this is the main process, the one that calls this function.

src/common/commonutils/CommandUtils.c
wupeka: Done

src/common/commonutils/CommandUtils.c
```diff
@@ -29,270 +31,200 @@ static int NormalizeStatus(int status)
...
-    if ((timeoutSeconds > 0) || (NULL != callback))
+    workerPid = fork();
+    if (workerPid < 0)
     {
-        OsConfigLogDebug(log, "SystemCommand: executing command '%s' with timeout of %d seconds and%scancelation on %s thread",
-            command, timeout, (NULL == callback) ? " no " : " ", mainProcessThread ? "main process" : "worker");
+        OsConfigLogError(log, "Cannot fork for command '%s' - %s", command, strerror(errno));
+        close(pipefd[0]);
+        close(pipefd[1]);
+        return -1;
```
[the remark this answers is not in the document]
wupeka: That'd add another level of nesting which would contain virtually the whole function, for me this way is way clearer.

src/common/commonutils/CommandUtils.c
wupeka: Done.

src/common/commonutils/CommandUtils.c
wupeka: Done.

src/common/commonutils/CommandUtils.c
wupeka: There's no optimal buffer size here, it was chosen as a random power of 2 that seemed reasonable.

src/common/commonutils/CommandUtils.c
wupeka: It cannot.

src/common/commonutils/CommandUtils.c
wupeka: Why not? That's a standard construct for C infinite loop.

src/common/commonutils/CommandUtils.c
wupeka: Fixed.

src/common/commonutils/CommandUtils.c
wupeka: Done.

src/common/commonutils/CommandUtils.c
wupeka: Done.

src/common/commonutils/CommandUtils.c
wupeka: Done.

src/common/commonutils/CommandUtils.c
wupeka: Done.

src/common/tests/CommonUtilsUT.cpp
wupeka: Realistically speaking - what are the chances of getting ExecuteCommand with a 2 megabyte long command?

MariusNi: The function cand also execute whole scripts, not just single commands. There could also be a bug with a non-terminated command string.

src/common/commonutils/CommandUtils.c
```diff
@@ -3,296 +3,235 @@
...
     }
-    return newStatus;
-}
-
-static int SystemCommand(void* context, const char* command, int timeoutSeconds, CommandCallback callback, OsConfigLogHandle log)
-{
-    const int callbackIntervalSeconds = 5; //seconds
-    const int defaultCommandTimeout = 60; //seconds
+    // Create a pipe, then fork. Forked process duplicates the write pipe end to stdout and stderr,
+    // then execs the shell with the given command. The main process uses select() with a timeout.
```
MariusNi, line 50: >with a timeout [](http://example.com/codeflow?start=83&length=14)

Speaking of 'a timeout', this is not any timeout, but the one specified by the function 'timeoutSeconds' argument, correct?

src/common/commonutils/CommandUtils.c
```diff
@@ -3,296 +3,235 @@
...
+    // to read from the pipe, and keep track of both command  timeout and callbacks. The read loop
+    // ends when the read() returns EOF.
```
MariusNi, line 52: ... how does the cancellation via the callback work here? In that case the loop won't end at EOF. Also, similar question for when timeout is reached.

Can the comment describe these alternate endings as well, and how do they fit with the rest of the approach?

src/common/commonutils/CommandUtils.c
MariusNi: Thanks (and also I do agree with your points stated here).

MariusNi:
Thanks for responding to comments and for updating the PR! Changes look better and better. Just few questions remaining.  My signoff remains (LGTM).

src/common/commonutils/CommandUtils.c
wupeka: In this case that's the selectInterval.

src/common/commonutils/CommandUtils.c
wupeka: Improved the comment.

src/common/commonutils/CommandUtils.c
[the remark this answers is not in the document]
wupeka: In this case that's about the exec* call family.

src/common/tests/CommonUtilsUT.cpp
wupeka: Done.

src/common/commonutils/CommandUtils.c
wupeka: If you agree (and a quick survey on my team shows that they agree too) then why can't we do it this way? Although It's not a "brace location" style change that is automatable, we can do it gradually in the new and/or modified code, improving the quality and readability. I do understand worries that we won't have consistency in this area, but I think it's still worth it.

MariusNi: It appears that declaring local variables at beginning of function is actually C89/ANSI C standard. C99 does not have this requirement. C11 is newer than C99 and also does not have it either. I do agree and my only ask is to do this change gradually, in a way that does not look out of the place within same source code file, component or even per project.

jepio (approved):

MariusNi:
Hitting consistent UT failures now on local Ubuntu 22.04 machine with the reworked ExecuteCommand:

```
[2025-03-19 12:15:25][ERROR][CommandUtils.c:147] Canceled reading from pipe for command 'sleep 20'
[       OK ] CommonUtilsTest.CancelCommandOnWorkerThread (10017 ms)
[ RUN      ] CommonUtilsTest.CancelCommand
TestCommandCallback: context (nil)
TestCommandCallback: context (nil)
TestCommandCallback: context (nil)
TestCommandCallback: context (nil)
TestCommandCallback: context (nil)
/home/mn/azure-osconfig/src/common/tests/CommonUtilsUT.cpp:550: Failure
Expected equality of these values:
  125
  ExecuteCommand(nullptr, "sleep 20", false, true, 0, 120, &textResult, &(CallbackContext::TestCommandCallback), nullptr)
    Which is: 0
[  FAILED  ] CommonUtilsTest.CancelCommand (20002 ms)
```
And some horrendous extra logging of 0 and 1s for ExecuteTooLongCommand test in same, without having debug logging enabled: 

```
(...)
0101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101
[       OK ] CommonUtilsTest.ExecuteTooLongCommand (399 ms)
```

Sorry but I am reverting this PR to be fixed and re-submitted. Cannot anymore locally validate changes on my local Ubuntu box because of these regressions.

PR #888, opened 2025-02-27
Using the 'LoggingLevelCritical' logging level to collect information for telemetry events

MariusNi:
## Description

This PR adds support for the first version of telemetry for this project as part of the regular trace logs, via lines logged at the LoggingLevelCritical (2) logging level (the trace lines marked with '[CRITICAL]' will be picked up by the future telemetry client)

## Checklist

- [x] I have read the [contribution guidelines](https://github.com/Azure/azure-osconfig/blob/main/CONTRIBUTING.md).
- [x] All unit tests are passing.
- [x] I have merged the latest `dev` branch prior to this PR submission.
- [x] I ran [pre-commit](https://pre-commit.com/) on my changes prior to this PR submission.
- [x] I submitted this PR against the `dev` branch.

PR #902, opened 2025-03-06
Adding the global LoggingLevel to OSConfig configuration and logging support for all 7 levels matching RFC 5424 severities

MariusNi:
## Description

Continuation on improvements for logging for Azure OSConfig, including 'OSConfig for MC', here including the following:
- New global (for all OSConfig components) configuration for the maximum log size and the newly added multiplier for debug logs: 'MaxLogSize' (default of 1 MB), and 'MaxLogSizeDebugMultiplier' (default 5).
- Previous 'DebugLogging' configuration option replaced by a new 'LoggingLevel'.
- New global logging macro 'OsConfigLog' that takes as an argument a LoggingLevel value.
- Legacy 'OsConfigLogInfo' and 'OsConfigLogError' refactored to wrap the new global macro.
- Rest of 5 'shorcut' macros added ('OsConfigLogWarning, OsConfigLogDebug, etc) to cover all 7 levels.
- The 'Informational (6)' level remains as the default for OSConfig, users can enable the 'Debug (7)' logging level.
- The OSConfig Agent, the Platform and the ASB library all use the new 'OsConfigLogWarning' to warn when debug logging is enabled.

## Checklist


robertschaedler3 (approved):

robertschaedler3 (approved):

robertschaedler3 (approved):

.github/workflows/universalnrp-test.yml
```diff
@@ -187,7 +187,7 @@ jobs:
         uses: ./.github/actions/check-size
         with:
           package: '${{ steps.normalize.outputs.path }}'
-          limit: 290
+          limit: 300
```
robertwoj-microsoft, line 190: Could you clarify why is this changed here?

src/adapters/pnp/PnpAgent.c
```diff
@@ -538,7 +540,7 @@ int main(int argc, char *argv[])
 
     if (IsDebugLoggingEnabled())
     {
-        OsConfigLogInfo(GetLog(), "WARNING: debug logging is enabled. To disable debug logging, edit '%s' and restart OSConfig", CONFIG_FILE);
+        OsConfigLogWarning(GetLog(), "Debug logging is enabled. To disable debug logging, set 'LoggingLevel' to 6 in '%s' and restart OSConfig", CONFIG_FILE);
```
robertwoj-microsoft, line 543: nit: s/To disable debug logging/To disable it/

src/common/asb/Asb.c
```diff
@@ -640,8 +642,8 @@ static PerfClock g_perfClock = {{0, 0}, {0, 0}};
 // Expected time limits under ideal conditions
 // Maximum per-rule audit time: 5 seconds
 static const long g_maxAuditTime = 5000000;
-// Maximum ASB rule remediation time: 55 seconds
-static const long g_maxRemediateTime = 55000000;
+// Maximum ASB rule remediation time: 99 seconds
+static const long g_maxRemediateTime = 99000000;
```
robertwoj-microsoft, line 646: See same comment under #888

src/common/asb/Asb.c
```diff
@@ -881,9 +884,27 @@ void AsbInitialize(OsConfigLogHandle log)
 
     StartPerfClock(&g_perfClock, GetPerfLog());
 
+    if (FileExists(g_configurationFile))
```
robertwoj-microsoft, line 887: Should we log an information if the configuration file does not exist? I guess this may happen in the MC scenario, but would be good to have information that we're not parsing anything here, because the file is missing.

src/common/asb/Asb.c
```diff
@@ -881,9 +884,27 @@ void AsbInitialize(OsConfigLogHandle log)
...
+        if (NULL != (jsonConfiguration = LoadStringFromFile(g_configurationFile, false, log)))
+        {
+            SetLoggingLevel(GetLoggingLevelFromJsonConfig(jsonConfiguration, log));
+            SetMaxLogSize(GetMaxLogSizeFromJsonConfig(jsonConfiguration, log));
+            SetMaxLogSizeDebugMultiplier(GetMaxLogSizeDebugMultiplierFromJsonConfig(jsonConfiguration, log));
+            FREE_MEMORY(jsonConfiguration);
+        }
+    }
+
+    RestrictFileAccessToCurrentAccountOnly(g_configurationFile);
```
robertwoj-microsoft, line 898: This should be inside the above if statement I think

src/common/asb/Asb.c
```diff
@@ -881,9 +884,27 @@ void AsbInitialize(OsConfigLogHandle log)
...
+
     OsConfigLogInfo(log, "AsbInitialize: %s", g_asbName);
 
-    if (NULL != (cpuModel = GetCpuModel(GetPerfLog())))
+    if (IsDebugLoggingEnabled())
+    {
+        OsConfigLogWarning(log, "Debug logging is enabled. To disable debug logging, set 'LoggingLevel' to 6 in '%s'", g_configurationFile);
```
robertwoj-microsoft, line 904: nit: s/To disable debug logging/To disable it/

src/common/logging/Logging.c
```diff
@@ -104,27 +185,29 @@ char* GetFormattedTime(void)
     return g_logTime;
 }
 
-// Checks and rolls the log over if larger than MAX_LOG_SIZE
+// Checks and rolls the log over if larger than maximum size
 void TrimLog(OsConfigLogHandle log)
 {
-    int fileSize = 0;
+    unsigned int maxLogSize = IsDebugLoggingEnabled() ? ((g_maxLogSize < (INT_MAX / 5)) ? (g_maxLogSize * 5) : INT_MAX) : g_maxLogSize;
```
robertwoj-microsoft, line 191: UINT_MAX?

src/common/logging/Logging.c
```diff
@@ -104,27 +185,29 @@ char* GetFormattedTime(void)
...
+    log->trimLogCount = (log->trimLogCount < maxLogTrim) ? (log->trimLogCount + 1) : 1;
 
     // Check every 10 calls:
     if (0 == (log->trimLogCount % 10))
     {
         // In append mode the file pointer will always be at end of file:
         fileSize = ftell(log->log);
 
-        if ((fileSize >= MAX_LOG_SIZE) || (-1 == fileSize))
+        if ((((unsigned int)fileSize) >= maxLogSize) || (-1 == fileSize))
```
robertwoj-microsoft, line 210: nit: long vs int here, are we sure this will never wrap?

src/common/asb/Asb.c
MariusNi: Presence of the OSConfig configuration file is completely optional.  The file may exist and do not contain any of these values  - the logging settings themselves are all optional.  Better logging would be if the logging options were overwritten. The configuration from the policy is not yet in, when that will be added, we'll want to log about the local override. This is pending.

src/adapters/pnp/PnpAgent.c
MariusNi: Intent is to be fully explicit. Keeping this as-is.

src/common/logging/Logging.c
MariusNi: Fixing

src/common/logging/Logging.c
MariusNi: The only reason long is there is because ftell returns a long and the only case when ftell returns a negative value is in case of error when it returns -1. Also, on 64 bit, long and int are same size. Could move the cast to line 208 but won't change anything.

src/common/asb/Asb.c
MariusNi: After further consideration, decided that to log that the file does not exist may alarm users. The users do not need to know of this file unless they want to enable debug logging or manually increase the log size, following our or CSE (support engineering) instructions.

src/common/logging/Logging.c
jepio: > Also, on 64 bit, long and int are same size. Could move the cast to line 208 but won't change anything.

That's not the case on Linux. On Linux int is always 32-bit and long has the same size as pointers (64-bit on 64-bit):
```bash
$ cat test.c
#include <stdio.h>
int main()
{
        long a;
        int b;
        printf("sizeof long=%zu\nsizeof int=%zu\n", sizeof(a), sizeof(b));
}
$ gcc test.c && ./a.out
sizeof long=8
sizeof int=4
```

MariusNi: Fixed by moving the cast to right, from unsigned int to long. If this is not enough, we could also check that the long is smaller than UINT_MAX. Please take a look at latest and let me know what you think.

robertwoj-microsoft (approved):

src/common/asb/Asb.c
```diff
@@ -893,18 +893,18 @@ void AsbInitialize(OsConfigLogHandle log)
...
+        RestrictFileAccessToCurrentAccountOnly(g_configurationFile);
+    }
 
     OsConfigLogInfo(log, "AsbInitialize: %s", g_asbName);
 
-    if (IsDebugLoggingEnabled())
+    if (IsConsoleLoggingEnabled())
     {
-        OsConfigLogWarning(log, "Debug logging is enabled. To disable debug logging, set 'LoggingLevel' to 6 in '%s'", g_configurationFile);
+        OsConfigLogWarning(log, "AsbInitialize: console logging is enabled. If the log rotator is not enabled this may result in a fill-up of the local storage space");
```
robertschaedler3, line 904: nit: Maybe useful to change around the wording here?
```suggestion
        OsConfigLogWarning(log, "AsbInitialize: console logging is enabled. If log rotation is not enabled, this may result in excessive storage use.");
```

robertschaedler3 (approved):

src/common/asb/Asb.c
MariusNi: The actual name is 'rotater' so yes, we may do better. Changing to simple 'rotation'

AhmedBM (approved):

robertschaedler3 (approved):

robertschaedler3 (approved):

PR #911, opened 2025-03-17
Simplify RemoveUserAccounts

robertwoj-microsoft:
## Description

Simplify RemoveUserAccounts:

* Avoid 2-stage iteration over /etc/passwd entries
* Add module test to make sure we no longer remove users whose names end with one of supplied names to be removed
* Add a temporary function to remove users based on username instead of SimplifiedUser structure
* Simplify iteration over /etc/passwd entries
* Drop home directory removal functionality

## Checklist

### DIFF
--- a/src/common/commonutils/UserUtils.c
+++ b/src/common/commonutils/UserUtils.c
@@ -3004,13 +3004,12 @@ int CheckUserAccountsNotFound(const char* names, char** reason, OsConfigLogHandl
             break;
         }
 
-        OsConfigLogInfo(log, "CheckUserAccountsNotFound: checking user '%s' (%u, %u) against '%s'", pw->pw_name, pw->pw_uid, pw->pw_gid, usernames);
         for (token = strtok(usernames, ","); token != NULL; token = strtok(NULL, ","))
         {
-            OsConfigLogInfo(log, "CheckUserAccountsNotFound: checking user '%s' (%u, %u) against '%s'", pw->pw_name, pw->pw_uid, pw->pw_gid, token);
-
             if (0 != strcmp(pw->pw_name, token))
+            {
                 continue;
+            }
             OsConfigLogInfo(log, "CheckUserAccountsNotFound: user '%s' found with id %u, gid %u, home '%s'",
                 pw->pw_name, pw->pw_uid, pw->pw_gid, pw->pw_dir);
 
@@ -3025,7 +3024,7 @@ int CheckUserAccountsNotFound(const char* names, char** reason, OsConfigLogHandl
 
         FREE_MEMORY(usernames);
     }
-    if (status == 0 && 0 != errno)
+    if ((0 == status) && (0 != errno))
     {
         status = errno;
         OsConfigLogError(log, "CheckUserAccountsNotFound: getpwent() failed with error %d", status);
@@ -3074,10 +3073,8 @@ int RemoveUserAccounts(const char* names, OsConfigLogHandle log)
             break;
         }
 
-        OsConfigLogInfo(log, "RemoveUserAccounts: checking user '%s' (%u, %u) against '%s'", pw->pw_name, pw->pw_uid, pw->pw_gid, usernames);
         for (token = strtok(usernames, ","); token != NULL; token = strtok(NULL, ","))
         {
-            OsConfigLogInfo(log, "RemoveUserAccounts: checking user '%s' (%u, %u) against '%s'", pw->pw_name, pw->pw_uid, pw->pw_gid, token);
             // Skip root user
             if (pw->pw_uid == 0)
             {
--- a/src/common/commonutils/UserUtils.c
+++ b/src/common/commonutils/UserUtils.c
@@ -11,7 +11,6 @@
 
 static const char* g_root = "root";
 static const char* g_shadow = "shadow";
-static const char* g_etcShadow = "/etc/shadow";
 static const char* g_etcPasswd = "/etc/passwd";
 
 static const char* g_noLoginShell[] = { "/usr/sbin/nologin", "/sbin/nologin", "/bin/false", "/bin/true", "/usr/bin/true", "/usr/bin/false", "/dev/null", "" };
@@ -2981,15 +2980,10 @@ int SetUsersRestrictedDotFiles(unsigned int* modes, unsigned int numberOfModes, 
 
 int CheckUserAccountsNotFound(const char* names, char** reason, OsConfigLogHandle log)
 {
-    const char* userTemplate = "%s:";
-    size_t namesLength = 0;
     char* name = NULL;
-    char* decoratedName = NULL;
-    SimplifiedUser* userList = NULL;
-    SimplifiedGroup* groupList = NULL;
-    unsigned int userListSize = 0, groupListSize = 0, i = 0, j = 0;
     bool found = false;
     int status = 0;
+    struct passwd* pw;
 
     if (NULL == names)
     {
@@ -2997,98 +2991,48 @@ int CheckUserAccountsNotFound(const char* names, char** reason, OsConfigLogHandl
         return EINVAL;
     }
 
-    namesLength = strlen(names);
-
-    if (0 == (status = EnumerateUsers(&userList, &userListSize, reason, log)))
+    setpwent();
+    for (errno = 0, pw = getpwent(); pw != NULL; errno = 0, pw = getpwent())
     {
-        for (i = 0; i < userListSize; i++)
+        char* usernames = NULL;
+        char* token = NULL;
+
+        usernames = strdup(names);
+        if (NULL == usernames)
         {
-            for (j = 0; j < namesLength; j++)
+            OsConfigLogError(log, "CheckUserAccountsNotFound: failed to duplicate string");
+            status = ENOMEM;
+            break;
+        }
+
+        OsConfigLogInfo(log, "CheckUserAccountsNotFound: checking user '%s' (%u, %u) against '%s'", pw->pw_name, pw->pw_uid, pw->pw_gid, usernames);
+        for (token = strtok(usernames, ","); token != NULL; token = strtok(NULL, ","))
+        {
+            OsConfigLogInfo(log, "CheckUserAccountsNotFound: checking user '%s' (%u, %u) against '%s'", pw->pw_name, pw->pw_uid, pw->pw_gid, token);
+
+            if (0 == strcmp(pw->pw_name, token))
             {
-                if (NULL == (name = DuplicateString(&(names[j]))))
+                OsConfigLogInfo(log, "CheckUserAccountsNotFound: user '%s' found with id %u, gid %u, home '%s'",
+                    pw->pw_name, pw->pw_uid, pw->pw_gid, pw->pw_dir);
+
+                if (DirectoryExists(pw->pw_dir))
                 {
-                    OsConfigLogError(log, "CheckUserAccountsNotFound: failed to duplicate string");
-                    status = ENOMEM;
-                    break;
-                }
-                else
-                {
-                    TruncateAtFirst(name, ',');
-
-                    if (0 == strcmp(userList[i].username, name))
-                    {
-                        EnumerateUserGroups(&userList[i], &groupList, &groupListSize, reason, log);
-                        FreeGroupList(&groupList, groupListSize);
-
-                        OsConfigLogInfo(log, "CheckUserAccountsNotFound: user '%s' found with id %u, gid %u, home '%s' and present in %u group(s)",
-                            userList[i].username, userList[i].userId, userList[i].groupId, userList[i].home, groupListSize);
-
-                        if (DirectoryExists(userList[i].home))
-                        {
-                            OsConfigLogInfo(log, "CheckUserAccountsNotFound: home directory of user '%s' exists ('%s')", name, userList[i].home);
-                        }
-
-                        OsConfigCaptureReason(reason, "User '%s' found with id %u, gid %u, home '%s' and present in %u group(s)",
-                            userList[i].username, userList[i].userId, userList[i].groupId, userList[i].home, groupListSize);
-
-                        found = true;
-                    }
+                    OsConfigLogInfo(log, "CheckUserAccountsNotFound: home directory of user '%s' exists ('%s')", name, pw->pw_dir);
                 }
 
-                j += strlen(name);
-                FREE_MEMORY(name);
+                OsConfigCaptureReason(reason, "User '%s' found with id %u, gid %u, home '%s'", pw->pw_name, pw->pw_uid, pw->pw_gid, pw->pw_dir);
+                found = true;
             }
         }
+
+        FREE_MEMORY(usernames);
     }
-
-    FreeUsersList(&userList, userListSize);
-
-    if ((false == found) && (0 == status))
+    if (status == 0 && 0 != errno)
     {
-        OsConfigLogInfo(log, "CheckUserAccountsNotFound: none of the requested user accounts ('%s') were found in the users database", names);
+        status = errno;
+        OsConfigLogError(log, "CheckUserAccountsNotFound: getpwent() failed with error %d", status);
     }
-
-    if (0 == status)
-    {
-        for (j = 0; j < namesLength; j++)
-        {
-            if (NULL == (name = DuplicateString(&(names[j]))))
-            {
-                OsConfigLogError(log, "CheckUserAccountsNotFound: failed to duplicate string");
-                status = ENOMEM;
-                break;
-            }
-            else
-            {
-                TruncateAtFirst(name, ',');
-
-                if (NULL == (decoratedName = FormatAllocateString(userTemplate, name)))
-                {
-                    OsConfigLogError(log, "CheckUserAccountsNotFound: out of memory, unable to check for user '%s' presence in '%s' andr '%s'",
-                        name, g_etcPasswd, g_etcShadow);
-                }
-                else
-                {
-                    if (0 == FindTextInFile(g_etcPasswd, decoratedName, log))
-                    {
-                        OsConfigCaptureReason(reason, "Account '%s' found mentioned in '%s'", name, g_etcPasswd);
-                        found = true;
-                    }
-
-                    if (0 == FindTextInFile(g_etcShadow, decoratedName, log))
-                    {
-                        OsConfigCaptureReason(reason, "Account '%s' found mentioned in '%s'", name, g_etcShadow);
-                        found = true;
-                    }
-
-                    FREE_MEMORY(decoratedName);
-                }
-
-                j += strlen(name);
-                FREE_MEMORY(name);
-            }
-        }
-    }
+    endpwent();
 
     if (found)
     {
--- a/src/common/commonutils/UserUtils.c
+++ b/src/common/commonutils/UserUtils.c
@@ -940,69 +940,6 @@ int CheckNoDuplicateGidsExist(char** reason, OsConfigLogHandle log)
     return status;
 }
 
-int RemoveGroup(SimplifiedGroup* group, bool removeHomeDirs, OsConfigLogHandle log)
-{
-    const char* commandTemplate = "groupdel -f %s";
-    char* command = NULL;
-    SimplifiedUser* userList = NULL;
-    unsigned int userListSize = 0;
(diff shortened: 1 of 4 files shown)


src/common/commonutils/UserUtils.c
```diff
@@ -867,6 +864,38 @@ int RemoveUser(SimplifiedUser* user, bool removeHome, OsConfigLogHandle log)
     return status;
 }
 
+int RemoveUser2(const char* username, OsConfigLogHandle log)
```
robertwoj-microsoft, line 867: This is temporary and intended to be unified with RemoveUser.

src/common/commonutils/UserUtils.c
```diff
@@ -867,6 +864,38 @@ int RemoveUser(SimplifiedUser* user, bool removeHome, OsConfigLogHandle log)
...
+{
+    const char* commandTemplate = "userdel -f %s";
+    char command[64] = { 0 };
```
robertwoj-microsoft, line 870: manpage for useradd says username must be 32 characters long. We check the result of snprintf later, so in case this limit causes failures, we'll know that from the logs and can extend if there's a need.

src/common/commonutils/UserUtils.c
```diff
@@ -910,69 +939,6 @@ int CheckNoDuplicateGidsExist(char** reason, OsConfigLogHandle log)
     return status;
 }
 
-int RemoveGroup(SimplifiedGroup* group, bool removeHomeDirs, OsConfigLogHandle log)
```
robertwoj-microsoft, line 913: Not used.

src/common/commonutils/UserUtils.c
```diff
@@ -3014,114 +2980,57 @@ int SetUsersRestrictedDotFiles(unsigned int* modes, unsigned int numberOfModes,
...
-                            userList[i].username, userList[i].userId, userList[i].groupId, userList[i].home, groupListSize);
-
-                        found = true;
-                    }
-                }
-
-                j += strlen(name);
-                FREE_MEMORY(name);
-            }
+        usernames = strdup(names);
```
robertwoj-microsoft, line 2999: strtok requires modifiable memory.

src/common/commonutils/UserUtils.c
```diff
@@ -3014,114 +2980,57 @@ int SetUsersRestrictedDotFiles(unsigned int* modes, unsigned int numberOfModes,
...
+        OsConfigLogInfo(log, "CheckUserAccountsNotFound: checking user '%s' (%u, %u) against '%s'", pw->pw_name, pw->pw_uid, pw->pw_gid, usernames);
+        for (token = strtok(usernames, ","); token != NULL; token = strtok(NULL, ","))
+        {
+            OsConfigLogInfo(log, "CheckUserAccountsNotFound: checking user '%s' (%u, %u) against '%s'", pw->pw_name, pw->pw_uid, pw->pw_gid, token);
 
-    if ((false == found) && (0 == status))
-    {
-        OsConfigLogInfo(log, "CheckUserAccountsNotFound: none of the requested user accounts ('%s') were found in the users database", names);
-    }
+            if (0 != strcmp(pw->pw_name, token))
```
robertwoj-microsoft, line 3012: Need to add braces.

src/modules/test/recipes/SecurityBaselineTests.json
```diff
@@ -2403,6 +2403,33 @@
...
+  {
+    "ObjectType": "Desired",
+    "ComponentName": "SecurityBaseline",
+    "ObjectName": "remediateEnsureUnnecessaryAccountsAreRemoved",
+    "Payload": "securitybaselinetest"
+  },
+  {
+    "RunCommand": "! id securitybaselinetest"
+  },
+  {
```
robertwoj-microsoft, line 2427: Cleanup

src/common/commonutils/UserUtils.c
```diff
@@ -3014,114 +2980,56 @@ int SetUsersRestrictedDotFiles(unsigned int* modes, unsigned int numberOfModes,
...
-
-                        OsConfigLogInfo(log, "CheckUserAccountsNotFound: user '%s' found with id %u, gid %u, home '%s' and present in %u group(s)",
-                            userList[i].username, userList[i].userId, userList[i].groupId, userList[i].home, groupListSize);
-
-                        if (DirectoryExists(userList[i].home))
-                        {
-                            OsConfigLogInfo(log, "CheckUserAccountsNotFound: home directory of user '%s' exists ('%s')", name, userList[i].home);
-                        }
+        char* usernames = NULL;
+        char* token = NULL;
```
MariusNi, line 2997: Why remove EnumerateUsers to duplicate functionality calling getpwent, etc.?

Overall, can you explain what's the intent here and what 'simplify' means? Are there bugs with current implementation?

robertwoj-microsoft: > Why remove EnumerateUsers to duplicate functionality calling getpwent, etc.? Please consider keeping the common pattern used by several functions here.

EnumerateUsers has some drawbacks:
- it reads the whole /etc/passwd file to get the number of lines
- it allocates full list of users 
- it reads /etc/passwd again to obtain the needed information
- it performs additional work for password loading which is unnecessary most of the time
- all the needed information is already in struct passwd

All this additional complexity is not worth pursuing the common API in my opinion.

robertwoj-microsoft: > Overall, can you explain what's the intent here and what 'simplify' means?

It means we reduce overhead mentioned in the previous comment, AND we no longer make 2 separate passes over /etc/passwd reading, the other pass uses a wrong tool for the purpose.

robertwoj-microsoft: > Are there bugs with current implementation?

Yes. As discussed on Friday, as in the PR description and in the ADO ticket.
So once again:
There is a severe issue that we remove more users than expected. If we pass 'a' as the parameter for RemediateEnsureUnnecessaryAccountsAreRemoved, we will remove all users who's names end with 'a' letter from the system, along with their home directories.
Another thing is we make our module tests useless by removing the 'cloudtest' user which is supposed to run module tests other than ASB.

MariusNi: Understood, thanks for explaining. Please go ahead and finalize this change and when ready for final code review, please ping me. Thanks!

src/common/commonutils/UserUtils.c
MariusNi: Please go ahead and unify it. Unless you want to have this PR serve only as a model/guidance, close it and have us open a new PR with the fix for the ADO ticket following this guidance. Whichever way feels better to you. Thanks

src/common/commonutils/UserUtils.c
```diff
@@ -819,49 +816,33 @@ int CheckNoDuplicateUidsExist(char** reason, OsConfigLogHandle log)
...
 }
 
-int RemoveUser(SimplifiedUser* user, bool removeHome, OsConfigLogHandle log)
+static int RemoveUser(const char* username, OsConfigLogHandle log)
 {
-    const char* commandTemplate = "userdel %s %s";
-    char* command = NULL;
-    int status = 0, _status = 0;
+    const char* commandTemplate = "userdel -f %s";
+    char command[64] = { 0 };
```
MariusNi, line 822: >{ 0 } [](http://example.com/codeflow?start=23&length=5)

nit: '''{0}'''

src/common/commonutils/UserUtils.c
```diff
@@ -819,49 +816,33 @@ int CheckNoDuplicateUidsExist(char** reason, OsConfigLogHandle log)
...
-    if (NULL == user)
+    if (NULL == username)
     {
         OsConfigLogError(log, "RemoveUser: invalid argument");
         return EINVAL;
     }
-    else if (0 == user->userId)
+
+    commandLen = snprintf(command, sizeof(command), commandTemplate, username);
+    if((int)sizeof(command) <= commandLen || 0 > commandLen)
```
MariusNi, line 833: Do instead a FormatAllocateString, and check the length of the now dynamically allocated string?

src/common/commonutils/UserUtils.c
```diff
@@ -819,49 +816,33 @@ int CheckNoDuplicateUidsExist(char** reason, OsConfigLogHandle log)
...
-                OsConfigLogInfo(log, "RemoveUser: home directory of user '%s' successfully removed ('%s')", user->username, user->home);
-            }
-        }
-        else
-        {
-            OsConfigLogInfo(log, "RemoveUser: cannot remove user '%s' (%u, %u) (%d)", user->username, user->userId, user->groupId, _status);
-        }
-
-        FREE_MEMORY(command);
+        OsConfigLogInfo(log, "RemoveUser: removed user '%s'", username);
```
MariusNi, line 841: >OsConfigLogInfo(log, "RemoveUser: removed user '%s'", username); [](http://example.com/codeflow?start=8&length=64)

Keep existing trace with full info?

```
OsConfigLogInfo(log, "RemoveUser: removed user '%s' (%u, %u, '%s')", user->username, user->userId, user->groupId, user->home);
```

src/common/commonutils/UserUtils.c
```diff
@@ -819,49 +816,33 @@ int CheckNoDuplicateUidsExist(char** reason, OsConfigLogHandle log)
...
+    if (0 == (status = ExecuteCommand(NULL, command, false, false, 0, 0, NULL, NULL, log)))
     {
-        if (0 == (status = ExecuteCommand(NULL, command, false, false, 0, 0, NULL, NULL, log)))
-        {
-            OsConfigLogInfo(log, "RemoveUser: removed user '%s' (%u, %u, '%s')", user->username, user->userId, user->groupId, user->home);
-
-            if (DirectoryExists(user->home))
-            {
-                OsConfigLogInfo(log, "RemoveUser: home directory of user '%s' remains ('%s') and needs to be manually deleted", user->username, user->home);
-            }
```
MariusNi, line 848: We need this to remain and perhaps the Info changed to a Warning (OsConfigLogWarning)

src/common/commonutils/UserUtils.c
```diff
@@ -3014,114 +2932,56 @@ int SetUsersRestrictedDotFiles(unsigned int* modes, unsigned int numberOfModes,
...
-    const char* userTemplate = "%s:";
-    size_t namesLength = 0;
-    char* name = NULL;
-    char* decoratedName = NULL;
-    SimplifiedUser* userList = NULL;
-    SimplifiedGroup* groupList = NULL;
-    unsigned int userListSize = 0, groupListSize = 0, i = 0, j = 0;
     bool found = false;
     int status = 0;
+    struct passwd* pw;
```
MariusNi, line 2937: >struct passwd* pw; [](http://example.com/codeflow?start=4&length=18)

(nit) Add explicit initialization

src/common/commonutils/UserUtils.c
```diff
@@ -3014,114 +2932,56 @@ int SetUsersRestrictedDotFiles(unsigned int* modes, unsigned int numberOfModes,
...
-                        }
-
-                        OsConfigCaptureReason(reason, "User '%s' found with id %u, gid %u, home '%s' and present in %u group(s)",
-                            userList[i].username, userList[i].userId, userList[i].groupId, userList[i].home, groupListSize);
-
-                        found = true;
-                    }
-                }
+        char* usernames = NULL;
+        char* token = NULL;
```
MariusNi, line 2949: We discussed to do this change longver term, but here let's  keep consistence with the rest of the file and pls move these up

src/common/commonutils/UserUtils.c
```diff
@@ -3014,114 +2932,56 @@ int SetUsersRestrictedDotFiles(unsigned int* modes, unsigned int numberOfModes,
...
-                    }
-
-                    FREE_MEMORY(decoratedName);
-                }
-
-                j += strlen(name);
-                FREE_MEMORY(name);
+            if (DirectoryExists(pw->pw_dir))
+            {
+                OsConfigLogInfo(log, "CheckUserAccountsNotFound: home directory of user '%s' exists ('%s')", pw->pw_name, pw->pw_dir);
```
MariusNi, line 2970: >OsConfigLogInfo [](http://example.com/codeflow?start=16&length=15)

Make this a OSConfigLogWarning instead?

src/common/commonutils/UserUtils.c
```diff
@@ -3140,134 +3000,64 @@ int CheckUserAccountsNotFound(const char* names, char** reason, OsConfigLogHandl
...
+    setpwent();
+    for (errno = 0, pw = getpwent(); pw != NULL; errno = 0, pw = getpwent())
     {
-        OsConfigLogInfo(log, "RemoveUserAccounts: CheckUserAccountsNotFound('%s') returned %d", names, status);
-        return status;
-    }
-
-    namesLength = strlen(names);
+        char* usernames = NULL;
+        char* token = NULL;
```
MariusNi, line 3018: Move up for consistence with rest of the code in this file (we discussed to do this but let's also keep consistence)

src/common/commonutils/UserUtils.c
```diff
@@ -3140,134 +3000,64 @@ int CheckUserAccountsNotFound(const char* names, char** reason, OsConfigLogHandl
...
-    }
-
-    if (0 == status)
-    {
-        for (j = 0; j < namesLength; j++)
+        for (token = strtok(usernames, ","); token != NULL; token = strtok(NULL, ","))
         {
-            if (NULL == (name = DuplicateString(&(names[j]))))
+            // Skip root user
+            if (pw->pw_uid == 0)
```
MariusNi, line 3031: We need to log an error here saying that we cannot remove root user. Why error: this may come from the policy parameter in the future, submitted by user or from MOF.

src/modules/test/recipes/SecurityBaselineTests.json
```diff
@@ -2403,6 +2403,33 @@
...
+    "RunCommand": "userdel securitybaselinetest || true"
+  },
+  {
+    "RunCommand": "groupdel securitybaselinetest || true"
+  },
   {
```
MariusNi, line 2433: We need the test to be updated for the Init and Get (audit) phase as well. Also, we need to avoid inserting 'RunCommand" in here, they may corrupt the overall test (we need to try to do as much as possible and ideally everything via our own remediation)

src/modules/test/recipes/SecurityBaselineTests.json
```diff
@@ -2403,6 +2403,33 @@
...
     "Action": "UnloadModule"
   }
```
MariusNi, line 2435: Don't mix desired with reported. The desired sequence for remediation is higher in the file. Please add desired, init and reported test cases or update the existing to correct the test behavior

AhmedBM (approved):

src/common/commonutils/UserUtils.c
robertwoj-microsoft: Why? This is simpler and much faster (stack vs heap). Leaving as is.

src/common/commonutils/UserUtils.c
robertwoj-microsoft: Why do we need this to remain?

src/modules/test/recipes/SecurityBaselineTests.json
robertwoj-microsoft: > they may corrupt the overall test

Can you tell how? It's a user created specifically for the test and I clean it up later.

robertwoj-microsoft: > we need to try to do as much as possible and ideally everything via our own remediation

No. This is a specific case that is tested here. We previously dropped the 'securitybaselinetest' user by passing the 'games,test' argument and this sequence verifies this particular case.

src/modules/test/recipes/SecurityBaselineTests.json
robertwoj-microsoft: I don't really understand, why add reported test here? Audit was flawed the same as remediation and that is one of the reasons you should not only check remediations with audits. That's why I check remediation with the id command in line 2425.

src/common/commonutils/UserUtils.c
robertwoj-microsoft: > We need to log an error

Previous equivalent:
```
OsConfigLogInfo(log, "RemoveUser: cannot remove user with uid 0 ('%s' %u, %u)", user->username, user->userId, user->groupId);
```

Making it as error as you request.

src/common/commonutils/UserUtils.c
robertwoj-microsoft: Done

src/common/commonutils/UserUtils.c
robertwoj-microsoft: Done as requested, but I'm curious about your stance here, so not resolving this thread just yet.

src/common/commonutils/UserUtils.c
```diff
@@ -1387,7 +1309,7 @@ int RemoveUsersWithoutPasswords(OsConfigLogHandle log)
                     OsConfigLogInfo(log, "RemoveUsersWithoutPasswords: the root account's password must be manually fixed");
                     status = EPERM;
                 }
-                else if ((0 != (_status = RemoveUser(&(userList[i]), false, log))) && (0 == status))
+                else if ((0 != (_status = RemoveUser(userList[i].username, userList[i].userId, userList[i].groupId, userList[i].home, log))) && (0 == status))
```
MariusNi, line 1312: >userList[i].username, userList[i].userId, userList[i].groupId, userList[i].home [](http://example.com/codeflow?start=53&length=79)

Passing in as argument userList[i] was simpler (1 arg istead of 2) and consistent with the other functions in this file. Any particular reason why instead of the 'simplified user' structure we unpacked and use name and id as separate arguments?

src/common/commonutils/UserUtils.c
```diff
@@ -2661,7 +2583,7 @@ int SetSystemAccountsNonLogin(OsConfigLogHandle log)
                 // If the account is not already true non-login, try to make it non-login and if that does not work, remove the account
                 if (0 != (_status = SetUserNonLogin(&(userList[i]), log)))
                 {
-                    _status = RemoveUser(&(userList[i]), false, log);
+                    _status = RemoveUser(userList[i].username, userList[i].userId, userList[i].groupId, userList[i].home, log);
```
MariusNi, line 2586: >_status = RemoveUser(userList[i].username, userList[i].userId, userList[i].groupId, userList[i].home, log); [](http://example.com/codeflow?start=20&length=107)

If we can rework this fix to not change lines like this, it would simplify the PR and also be consistent with existing code.

src/modules/test/recipes/SecurityBaselineTests.json
MariusNi: For every rule, we have a remediation test, and an audit test which also needs the initialization route. Please consider balancing and adding the check to all cases, in a fashion alike to existing tests.

src/common/commonutils/UserUtils.c
robertwoj-microsoft: Yes, it was simpler to call, but it required filling out the full structure along with hashed user passwords and what not. Filling this whole information just to log an entry is much more of an overkill to me.

src/modules/test/recipes/SecurityBaselineTests.json
robertwoj-microsoft: > we need to avoid inserting 'RunCommand" in here

> in a fashion alike to existing tests

I have left previous tests untouched, but it's been proven the functionality is not tested properly. I don't intend to reduce tests just for the file to look better.

robertwoj-microsoft:
This PR was meant to be a proposition to solve an issue by reducing complexity and unnecessary work. The comments above suggest the code to go back to old, existing implementation utilizing EnumerateUsers which in my opinion is unnecessary. I have added tests which I feel are not welcome as they don't follow existing style. Considering that, I'm closing this PR as it doesn't fit the vision and let you propose better solutions. Feel free to start from scratch or commit to this branch and do it properly.

src/common/commonutils/UserUtils.c
robertwoj-microsoft: I am leaving this as is.

robertwoj-microsoft:
Rebased with dev, moved the newly added tests to a separate file.

PR #920, opened 2025-03-19
ASB: change NOZEROCONF handling on RHEL like distros

jepio:
## Description

A user running NRP in Audit mode on RHEL 9.5 reported an issue with our AuditEnsureZeroconfNetworkingIsDisabled rule. The rule indicated non-compliance and the reported remediation was for the user to add `NOZEROCONF=yes` to `/etc/sysconfig/network`. That file is managed by cloud-init and the user's changes were being discarded on every boot.

Investigation found that:
1. NOZEROCONF setting is not relevant on a majority of RHEL systems which use network-manger, only those relying on legacy networking initscripts are affected by this setting. (see: https://access.redhat.com/solutions/5692821). This can be detected by checking for the `network.service` systemd unit, which is provided by the `network-scripts` package. Add this check to the Audit and Remediate paths.
2. User modifications to `/etc/sysconfig/network` are only preserved if they preceed the cloud-init header. To simplify remediation, always prepend the `NOZEROCONF=yes` line to the top of the file (if we need to remediate at all). This is covered by new tests, that show some possible contents of the file.

## Checklist


src/common/tests/CommonUtilsUT.cpp
```diff
@@ -2316,6 +2316,153 @@ TEST_F(CommonUtilsTest, ReplaceMarkedLinesInFile)
...
+    // Expected output when prepend=true and NOZEROCONF is commented out
+    // The commented line should be removed and new one prepended
+    const char* outFilePrependWithCommentedMatch =
+        "NOZEROCONF=yes\n"
+        "# Created by cloud-init automatically, do not edit.\n"
+        "#\n"
+        "NETWORKING=yes\n"
+        "NETWORKING_IPV6=yes\n"
+        "IPV6_AUTOCONF=no\n"
+        "# NOZEROCONF=yes\n";
```
robertwoj-microsoft, line 2378: The comment says this line should be removed

robertwoj-microsoft (approved):

MariusNi (approved):
This is a hot fix, correct? If so, we need to cherry pick this to main. Can you please do this so we will capture both both fixes (this and the one for tndf cache) in same MC package update?

AhmedBM (approved):

jepio:
> This is a hot fix, correct? If so, we need to cherry pick this to main. Can you please do this so we will capture both fixes (this and the one for tndf cache) in same MC package update?

Not a hotfix, the issue this addresses has been mitigated, so please give this a thorough review.

PR #936, opened 2025-03-25
Add the ability to mock ExecuteCommand calls

wupeka:
## Description
Add functionality that allows mocking ExecuteCommand calls, to be used in unit tests.

## Checklist

### DIFF
--- a/src/common/commonutils/CommandUtils.c
+++ b/src/common/commonutils/CommandUtils.c
@@ -15,6 +15,45 @@ static long MonotonicTime()
     return -1;
 }
 
+#ifdef TEST_CODE
+struct MockCommand
+{
+    const char* expectedCommand;
+    bool matchPrefix;
+    const char* output;
+    int returnCode;
+    struct MockCommand* next;
+};
+
+static struct MockCommand* g_mockCommand;
+
+void AddMockCommand(const char* expectedCommand, bool matchPrefix, const char* output, int returnCode)
+{
+    struct MockCommand* mock = malloc(sizeof(struct MockCommand));
+    // Handling memory error in a brutal way as it's only for tests, not for production usage.
+    if (NULL == mock)
+    {
+        abort();
+    }
+    mock->expectedCommand = expectedCommand;
+    mock->matchPrefix = matchPrefix;
+    mock->output = output;
+    mock->returnCode = returnCode;
+    mock->next = g_mockCommand;
+    g_mockCommand = mock;
+}
+
+void CleanupMockCommands()
+{
+    while (NULL != g_mockCommand)
+    {
+	struct MockCommand* next = g_mockCommand->next;
+        free(g_mockCommand);
+        g_mockCommand = next;
+    }
+}
+#endif
+
 #define BUFFER_SIZE 1024
 
 int ExecuteCommand(void* context, const char* command, bool replaceEol, bool forJson, unsigned int maxTextResultBytes, unsigned int timeoutSeconds,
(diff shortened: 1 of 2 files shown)


src/common/commonutils/CommandUtils.c
```diff
@@ -15,6 +15,44 @@ static long MonotonicTime()
     return -1;
 }
 
+struct mockCommand {
```
MariusNi, line 18: >{ [](http://example.com/codeflow?start=19&length=1)

nit: move to next line

MariusNi, line 18: >mockCommand [](http://example.com/codeflow?start=7&length=11)

Nit: use PascalCasing for struct names (camelCasing is for members)

src/common/commonutils/CommandUtils.c
```diff
@@ -15,6 +15,44 @@ static long MonotonicTime()
...
+    struct mockCommand* next;
+};
+
+static struct mockCommand* g_mockCommand;
+
+void AddMockCommand(const char* expectedCommand, bool matchPrefix, const char* output, int returnCode)
+{
+    struct mockCommand* mock;
+    mock = malloc(sizeof(struct mockCommand));
+    // Handling memory error in a brutal way as it's only for tests, not for production usage.
```
MariusNi, line 32: > // Handling memory error in a brutal way as it's only for tests, not for production usage. [](http://example.com/codeflow?start=3&length=91)

Doesn't this warrant to have this mocking code as part of the unit-tests instead?

MariusNi: Maybe a good opportunity to add a new common test library, that all tests can link to and use?

wupeka: Unfortunately gtest/gmock can't mock "free" (non-C++-class-member) functions, so this code needs to be close to ExecuteCommand itself. We could move AddMock* to a separate unit, but then we'd have to export g_mockCommand, and the "execution" part would still be left in ExecuteCommand body - so I don't think it's worth it.

src/common/commonutils/CommandUtils.c
wupeka: Done

src/common/commonutils/CommandUtils.c
```diff
@@ -15,6 +15,45 @@ static long MonotonicTime()
...
+    int returnCode;
+    struct MockCommand* next;
+};
+
+static struct MockCommand* g_mockCommand;
+
+void AddMockCommand(const char* expectedCommand, bool matchPrefix, const char* output, int returnCode)
+{
+    struct MockCommand* mock;
+    mock = malloc(sizeof(struct MockCommand));
```
robertwoj-microsoft, line 32: Merge with the line above

src/common/commonutils/CommandUtils.c
```diff
@@ -35,6 +74,22 @@ int ExecuteCommand(void* context, const char* command, bool replaceEol, bool for
         return E2BIG;
     }
 
+    // Allow mocked call for unit testing of things that execute commands.
+    struct MockCommand* mock = g_mockCommand;
+    while (NULL != mock) {
+        size_t stringLen = strlen(mock->expectedCommand);
+        if (!mock->matchPrefix && (strlen(command) > stringLen))
+        {
+            stringLen = strlen(mock->expectedCommand);
```
robertwoj-microsoft, line 83:
```suggestion
            stringLen = strlen(command));
```

robertwoj-microsoft (approved):

MariusNi (commented):
This is great as concept but we also need to set this TEST_CODE when building the unit-tests and only then. 

This should also not collide with the macro that we use to enable building the tests. 

Maybe change from TEST _CODE to MOCK_FOR_TEST or similar (harder to collide with something else)?

PR #938, opened 2025-03-26
Fixes and tests for ensureFilePermissions procedures

wupeka:
## Description
1. Use "owner" and not "user" parameter
2. Change behaviour of mask and permissions - mask has bits that must be cleared, permissions has bits that must be set
3. Full suite of unit tests

## Checklist

### DIFF
--- a/src/modules/compliance/src/lib/procedures/ensureFilePermissions.cpp
+++ b/src/modules/compliance/src/lib/procedures/ensureFilePermissions.cpp
@@ -27,7 +27,7 @@ AUDIT_FN(ensureFilePermissions)
         return false;
     }
 
-    if (args.find("user") != args.end())
+    if (args.find("owner") != args.end())
     {
         struct passwd* pwd = getpwuid(statbuf.st_uid);
         if (nullptr == pwd)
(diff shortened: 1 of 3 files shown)


robertwoj-microsoft (approved):

src/modules/compliance/src/lib/procedures/ensureFilePermissions.cpp
```diff
@@ -27,17 +27,17 @@ AUDIT_FN(ensureFilePermissions)
...
         if (nullptr == pwd)
         {
             logstream << "No user with uid " << statbuf.st_uid;
             return false;
         }
-        if (args["user"] != pwd->pw_name)
+        if (args["owner"] != pwd->pw_name)
         {
-            logstream << "Invalid user - is " << pwd->pw_name << " should be " << args["user"];
+            logstream << "Invalid owner - is " << pwd->pw_name << " should be " << args["owner"];
```
MariusNi, line 40: >logstream << "Invalid owner - is " << pwd->pw_name << " should be " << args["owner"]; [](http://example.com/codeflow?start=12&length=85)

Please use the OSConfig logging macros instead, to capture correct context from the place of failure or trace.

We have now macros for all levels:

```
OsConfigLogEmergency
OsConfigLogAlert
OsConfigLogCritical
OsConfigLogError
OsConfigLogWarning
OsConfigLogNotice
OsConfigLogInfo
OsConfigLogDebug

```

Please do this consistently for all of the Compliance code.

Not necessarily as part of this PR but would be great to clean-up everywhere the logging part.

Storing error and trace messages and passing them from one method to another is not recommended as it loses the context. Please use the logging macros and feel free to use any of the 8 levels that you feel are appropriate.

MariusNi (approved):
:shipit:

src/modules/compliance/src/lib/procedures/ensureFilePermissions.cpp
wupeka: We have discussed this before - this is not logging as in OsConfigLog*, this is the stream of information that will be presented to the user for him in the UI to show what part of audit has failed.

PR #939, opened 2025-03-26
Fix and simplify RemoveUserAccounts

robertwoj-microsoft:
## Description

Reopening #911 as it was rebased and won't let me reopen that one.

Note: Compared to #911, I have rebased with dev and moved the tests I have previously added to a separate file.

## Checklist

### DIFF
--- a/src/common/commonutils/UserUtils.c
+++ b/src/common/commonutils/UserUtils.c
@@ -3004,13 +3004,12 @@ int CheckUserAccountsNotFound(const char* names, char** reason, OsConfigLogHandl
             break;
         }
 
-        OsConfigLogInfo(log, "CheckUserAccountsNotFound: checking user '%s' (%u, %u) against '%s'", pw->pw_name, pw->pw_uid, pw->pw_gid, usernames);
         for (token = strtok(usernames, ","); token != NULL; token = strtok(NULL, ","))
         {
-            OsConfigLogInfo(log, "CheckUserAccountsNotFound: checking user '%s' (%u, %u) against '%s'", pw->pw_name, pw->pw_uid, pw->pw_gid, token);
-
             if (0 != strcmp(pw->pw_name, token))
+            {
                 continue;
+            }
             OsConfigLogInfo(log, "CheckUserAccountsNotFound: user '%s' found with id %u, gid %u, home '%s'",
                 pw->pw_name, pw->pw_uid, pw->pw_gid, pw->pw_dir);
 
@@ -3025,7 +3024,7 @@ int CheckUserAccountsNotFound(const char* names, char** reason, OsConfigLogHandl
 
         FREE_MEMORY(usernames);
     }
-    if (status == 0 && 0 != errno)
+    if ((0 == status) && (0 != errno))
     {
         status = errno;
         OsConfigLogError(log, "CheckUserAccountsNotFound: getpwent() failed with error %d", status);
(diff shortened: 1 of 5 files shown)


src/common/commonutils/UserUtils.c
```diff
@@ -819,49 +816,37 @@ int CheckNoDuplicateUidsExist(char** reason, OsConfigLogHandle log)
     return status;
 }
 
-int RemoveUser(SimplifiedUser* user, bool removeHome, OsConfigLogHandle log)
+static int RemoveUser(const char* username, const uid_t uid, const gid_t gid, const char* home, OsConfigLogHandle log)
```
MariusNi, line 819: >const char* username, const uid_t uid, [](http://example.com/codeflow?start=22&length=38)

Same feedback as before: why not keep the SimpliedUser argument, basically keep the RemoveUser declaration as used before just without the removeHome argument?

This would minimize the changes in this PR even further. Plus, most functions like this use the SimplifiedUser argument instead of separate user name and uid.

src/common/commonutils/UserUtils.c
```diff
@@ -819,49 +816,37 @@ int CheckNoDuplicateUidsExist(char** reason, OsConfigLogHandle log)
...
         return EINVAL;
     }
-    else if (0 == user->userId)
+
+    commandLen = snprintf(command, sizeof(command), commandTemplate, username);
+    if(((int)sizeof(command) <= commandLen) || (0 > commandLen))
     {
-        OsConfigLogInfo(log, "RemoveUser: cannot remove user with uid 0 ('%s' %u, %u)", user->username, user->userId, user->groupId);
-        return EPERM;
+        OsConfigLogError(log, "RemoveUser: command buffer too small");
```
MariusNi, line 835: >command buffer too small"); [](http://example.com/codeflow?start=43&length=27)

Add how many bytes the buffer is, and how many bytes are necessary. So we will understand the bug from the trace. E.g.:

OsConfigLogError(log, "RemoveUser: command buffer length of %lu bytes is under required length of %ly bytes");

src/common/commonutils/UserUtils.c
```diff
@@ -819,49 +816,37 @@ int CheckNoDuplicateUidsExist(char** reason, OsConfigLogHandle log)
...
-    if (NULL == user)
+    if (NULL == username)
     {
         OsConfigLogError(log, "RemoveUser: invalid argument");
         return EINVAL;
     }
-    else if (0 == user->userId)
+
+    commandLen = snprintf(command, sizeof(command), commandTemplate, username);
+    if(((int)sizeof(command) <= commandLen) || (0 > commandLen))
```
MariusNi, line 833: This PR says it simplifies the code. But here actually there is an added complication. Instead of using a single call, FormatAllocateString to allocate and initialize the command string, we have this static buffer, then length calculation, then snprintf call, then the error check etc?

The reason for having this function, FormatAllocateString, is exactly for cases like this. Let's use it more.

src/common/commonutils/UserUtils.c
```diff
@@ -3014,114 +2936,55 @@ int SetUsersRestrictedDotFiles(unsigned int* modes, unsigned int numberOfModes,
 
 int CheckUserAccountsNotFound(const char* names, char** reason, OsConfigLogHandle log)
```
MariusNi, line 2937: >CheckUserAccountsNotFound [](http://example.com/codeflow?start=4&length=25)

This function is indeed simplified, thanks!

src/common/commonutils/UserUtils.c
```diff
@@ -3140,134 +3003,65 @@ int CheckUserAccountsNotFound(const char* names, char** reason, OsConfigLogHandl
     return status;
 }
 
-int RemoveUserAccounts(const char* names, bool removeHomeDirs, OsConfigLogHandle log)
+int RemoveUserAccounts(const char* names, OsConfigLogHandle log)
```
MariusNi, line 3006: >RemoveUserAccounts [](http://example.com/codeflow?start=4&length=18)

We need the warning trace for a home dir remaining after its owner user account is deleted in this function as well (actually, is even more important here than in the Check function). Do you think you could add?

MariusNi (approved):
:shipit:

robertwoj-microsoft:
Closing again as despite you saying you like the PR you still insist on reverting the old functionality which I don't agree with. See PR #940 as follow up.

PR #949, opened 2025-04-02
[Hot Fix] Optimizing auditing presence of packages 

MariusNi:
## Description

We are unintendedly blocking the package managers during ASB audit by invoking the package managers at every rule to check if one of more packages are installed, one by one. We can speed up the overall ASB audit time and also avoid blocking the package managers so much by listing all installed packages once per session and saving to a cache buffer, then when looking to see if a package is installed, to search in that buffer without making any new package manager request. 

We refresh this cache when installing and uninstalling packages during automatic remediation (where we still need to do one package at a time, but nobody complained yet about that, and also remediation does not happen continuously like audit, being executed only once in most cases). In audit-only case however, there shall be only one cache initialization per NRP loaded session.

The Hot Fix in this PR applies to all package managers that we support:
- apt-get
- dpkg
- rpm
- tdnf
- dnf
- yum
- zypper

For building the cache (holding the list of installed packages) we give priority to dpkg and rpm. If these are not present, we try individual managers (tdnf, dnf, yum, zypper) as some distros can have such configurations.

## Checklist

### DIFF
--- a/src/common/commonutils/PackageUtils.c
+++ b/src/common/commonutils/PackageUtils.c
@@ -211,7 +211,7 @@ int IsPackageInstalled(const char* packageName, OsConfigLogHandle log)
         }
         else if (NULL == strstr(g_installedPackages, searchTarget))
         {
-            OsConfigLogInfo(log, "IsPackageInstalled: '%s' not found in the list of installed packages", packageName);
+            OsConfigLogInfo(log, "IsPackageInstalled: '%s' is not installed", packageName);
             status = ENOENT;
         }
 
@@ -221,10 +221,6 @@ int IsPackageInstalled(const char* packageName, OsConfigLogHandle log)
     if (0 == status)
     {
         OsConfigLogInfo(log, "IsPackageInstalled: package '%s' is installed", packageName);
-    }
-    else
-    {
-        OsConfigLogInfo(log, "IsPackageInstalled: package '%s' is not installed", packageName);
     }
 
     return status;
--- a/src/common/commonutils/PackageUtils.c
+++ b/src/common/commonutils/PackageUtils.c
@@ -110,6 +110,9 @@ static int CheckOrInstallPackage(const char* commandTemplate, const char* packag
 
     FREE_MEMORY(command);
 
+    // Refresh the cache holding list of installed packages
+    g_updateInstalledPackagesCache = 1;
+
     return status;
 }
 
@@ -140,7 +143,7 @@ static int CheckAllPackages(const char* commandTemplate, const char* packageMana
     return status;
 }
 
-static int ListAllInstalledPackages(OsConfigLogHandle log)
+static int UpdateInstalledPackagesCache(OsConfigLogHandle log)
 {
     const char* commandTemplateDpkg = "%s-query -W -f='${binary:Package}\n'";
     const char* commandTemplateYumDnf = "%s list installed  --cacheonly | awk '{print $1}'";
@@ -177,7 +180,7 @@ static int ListAllInstalledPackages(OsConfigLogHandle log)
         FREE_MEMORY(g_installedPackagesCache);
         if (NULL == (g_installedPackagesCache = DuplicateString(results)))
         {
-            OsConfigLogError(log, "ListAllInstalledPackages: out of memory");
+            OsConfigLogError(log, "UpdateInstalledPackagesCache: out of memory");
             status = ENOENT;
         }
     }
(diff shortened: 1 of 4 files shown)


src/common/commonutils/PackageUtils.c
```diff
@@ -94,47 +110,152 @@ static int CheckOrInstallPackage(const char* commandTemplate, const char* packag
 
     FREE_MEMORY(command);
 
+    // Refresh the cache holding list of installed packages
+    g_updateInstalledPackagesCache = 1;
```
MariusNi, line 114: > g_updateInstalledPackagesCache = 1; [](http://example.com/codeflow?start=3&length=36)

Here is a test of a comment

src/common/commonutils/PackageUtils.c
```diff
@@ -90,51 +105,168 @@ static int CheckOrInstallPackage(const char* commandTemplate, const char* packag
...
+    FREE_MEMORY(command);
+
+    return status;
+}
+
+static int UpdateInstalledPackagesCache(OsConfigLogHandle log)
+{
+    const char* commandTemplateDpkg = "%s-query -W -f='${binary:Package}\n'";
+    const char* commandTemplateYumDnf = "%s list installed  --cacheonly | awk '{print $1}'";
+    const char* commandTmeplateZypper = "%s search -i";
```
jepio, line 149: This might be better kept for a follow-up but someone suggested using `rpm -qa --queryformat "%{NAME}\n"` for all rpm distros, which produces the same output and skips locking/refresh.

MariusNi: Awesome! I will try this.

MariusNi: Rpm query works great and is more efficient than going the tdnf/yum/zypper route. We keep the later to try for those special cases when neither dpkg or rpm are present, but another package manager may be present. 

It's also better for the Hot Fix in particular, as with rpm we will put even less strain on zypper (what the customer complained about).

robertschaedler3 (dismissed):

src/common/commonutils/PackageUtils.c
```diff
@@ -3,12 +3,17 @@
 
 #include "Internal.h"
 
+#if ((defined(__GNUC__) && ((__GNUC__ > 4) || ((__GNUC__ == 4) && (__GNUC_MINOR__ >= 9)))) || defined(__clang__))
```
AhmedBM, line 6: For readability can we use named constants instead eg.

```c
#define USE_STD_ATOMIC ((defined(__GNUC__) && ((__GNUC__ > 4) || ((__GNUC__ == 4) && (__GNUC_MINOR__ >= 9)))) || defined(__clang__))

#if USE_STD_ATOMIC
#include <stdatomic.h>
#endif
```

src/common/commonutils/PackageUtils.c
```diff
@@ -90,51 +108,173 @@ static int CheckOrInstallPackage(const char* commandTemplate, const char* packag
...
+        if (NULL != (buffer = DuplicateString(results)))
+        {
+            FREE_MEMORY(g_installedPackagesCache);
+            g_installedPackagesCache = buffer;
+        }
+        else
+        {
+            // Leave the cache as-is, just log the error
+            OsConfigLogError(log, "UpdateInstalledPackagesCache: out of memory");
+            status = ENOENT;
```
AhmedBM, line 197: Should this be `ENOMEM` out of memory?

src/common/commonutils/PackageUtils.c
MariusNi: I like the proposal and we could move that USE_STD_ATOMIC macro in commonutils.h. This same pattern is in use in several places, is not new in here, and also is throw away code when we will go to gcc 10, and also other places may be deleted when we will cut deprecated parts. So will keep this feedback for applying it in a future PR. Thanks!

AhmedBM (dismissed):

robertschaedler3 (approved):

src/common/commonutils/PackageUtils.c
```diff
@@ -90,51 +108,173 @@ static int CheckOrInstallPackage(const char* commandTemplate, const char* packag
...
+        }
+
+        if (NULL == searchTarget)
+        {
+            OsConfigLogError(log, "IsPackageInstalled: out of memory");
+            status = ENOMEM;
+        }
+        else
+        {
+            if (NULL != strstr(g_installedPackagesCache, searchTarget))
```
AhmedBM, line 266: We should make sure the `searchTarget` is the whole word search as you may be picking a partial name of a package? eg.
 * `python`, `python3`
 * `ninja`, `ninja-build`

MariusNi: This is taken care of the "\n...\n" for the dpkg and rpm and other ways for the other managers. It's tested by our CI tests actually. We only get exact name packages. If we look for "bluetooth" it will be just that, no "bluetooth-nx" etc.

AhmedBM (approved):

PR #950, opened 2025-04-02
Coding style definition, take 1

wupeka:
## Description

## Checklist


CONTRIBUTING.md
```diff
@@ -10,10 +10,54 @@ This project has adopted the [Microsoft Open Source Code of Conduct](https://ope
...
+- Always chose readability over compactness
+- Avoid side-effects in if statements
+- Avoid ternary operators with complex statements
+## Naming Conventions
+- PascalCase for functions, types, and structs
+- camelCase for parameters and variables
+- g_ prefix for global variables
+- UPPER_CASE for macros
+## Variables
+- Variables must always be properly initialized
```
robertwoj-microsoft, line 34: Define 'properly':
int i;
...
i = param;
(no read, so technically that's correct)
Is it proper? OR it should be:
int i = 0;
...
i = param;

robertwoj-microsoft:
```suggestion
- Variables must always be initialized
```

CONTRIBUTING.md
```diff
@@ -10,10 +10,54 @@ This project has adopted the [Microsoft Open Source Code of Conduct](https://ope
...
+- Variables should be defined at the beginning of the code block they're used
+- Variables must not be reused for different purposes, except for simple index variables in for loops
```
robertwoj-microsoft, line 36: in other words no shadowing?

CONTRIBUTING.md
```diff
@@ -10,10 +10,54 @@ This project has adopted the [Microsoft Open Source Code of Conduct](https://ope
...
+- Variable names must be verbose enough to explain their purpose if not obvious
+- Use boolean true and false rather than 1 and 0 for boolean operations
+## Error Handling
+- Return integer status codes for functions that can fail, typically using standard errno values, and 0 for success
```
robertwoj-microsoft, line 40: In C, in compliance module we return Error() which is not int

CONTRIBUTING.md
wupeka: More like:
{
  int foo = 0;
  foo = resultOfSomeOperation();
  foo += 17;
  if (2 == foo % 4) { 
    return ERROR_INVALID_SOMETHING;
  }
  (...)
  foo = someVeryDifferentOperation();
  (...)
 }

CONTRIBUTING.md
```diff
@@ -10,10 +10,54 @@ This project has adopted the [Microsoft Open Source Code of Conduct](https://ope
...
+- Always validate function inputs
+- Use early returns for validation and error conditions
+- Log errors with descriptive messages using OsConfigLogError or OsConfigLogInfo
+## Memory Management
+- Always check for memory allocation failures, fail gracefully if possible
+- Avoid dangling invalid pointers - if it's freed it should be NULL, use FREE_MEMORY(ptr) macro to safely free memory.
+- Free all allocated memory before function returns, including on error paths
+- Functions allocating memory should document ownership transfer to caller
```
robertwoj-microsoft, line 48: Add an entry for C++ memory management:
- use smart pointers as much as possible - avoid explicit new and delete calls.

CONTRIBUTING.md
```diff
@@ -10,10 +10,54 @@ This project has adopted the [Microsoft Open Source Code of Conduct](https://ope
...
+## Logging
+- Include an OsConfigLogHandle log parameter in functions that need to log
+- Use appropriate log levels:
+  - OsConfigLogDebug: Detailed debugging information
+  - OsConfigLogInfo: General operational information
+  - OsConfigLogError: Function errors and failures
+- Log return values and status codes for debugging
```
robertwoj-microsoft, line 55: To consider: Add an entry to avoid logging sensitive information if possible (e.g. user-provided data, values that we find on customer machine may disclose information)

CONTRIBUTING.md
```diff
@@ -10,10 +10,54 @@ This project has adopted the [Microsoft Open Source Code of Conduct](https://ope
...
+- Use boolean true and false rather than 1 and 0 for boolean operations
+## Error Handling
+- Return integer status codes for functions that can fail, typically using standard errno values, and 0 for success
+- Always validate function inputs
+- Use early returns for validation and error conditions
+- Log errors with descriptive messages using OsConfigLogError or OsConfigLogInfo
+## Memory Management
+- Always check for memory allocation failures, fail gracefully if possible
+- Avoid dangling invalid pointers - if it's freed it should be NULL, use FREE_MEMORY(ptr) macro to safely free memory.
+- Free all allocated memory before function returns, including on error paths
```
robertwoj-microsoft, line 47:
```suggestion
- Free all allocated memory before function returns, including error paths
```

CONTRIBUTING.md
```diff
@@ -10,10 +15,59 @@ This project has adopted the [Microsoft Open Source Code of Conduct](https://ope
...
 Most contributions require you to agree to a Contributor License Agreement (CLA) declaring that you have the right to, and actually do, grant us the rights to use your contribution. For details, visit https://cla.microsoft.com. When you submit a pull request, a CLA-bot will automatically determine whether you need to provide a CLA and decorate the PR appropriately (e.g., label, comment). Simply follow the instructions provided by the bot. You will only need to do this once across all repositories using our CLA.
 
+# Coding style
+All code needs to be formatted according to .pre-commit-config.yaml. Each pull request is checked by [Formatting Tests](https://github.com/Azure/azure-osconfig/blob/dev/.github/workflows/formatting.yml).
+Apart from that there are style rules that can't be expressed in clang-format but must be followed nevertheless:
+
+## Formatting quirks
+- Boolean statements (e.g. in if) should explicitly use parenthesis and never rely on operators order:
+  - ${\color{red}No: `(a != b && c != d)`}$
+  - Yes: `((a != b) && (c !=d))`
```
robertwoj-microsoft, line 29:
```suggestion
  - Yes: `((a != b) && (c != d))`
```

CONTRIBUTING.md
robertwoj-microsoft: How about the generic status variable which is usually returned at the end?
```
int status = EINVAL;

status = foo();
if (status)
{
  // log
  return status;
}

status = bar();
if (status)
{
  // log
  return status;
}

return status;
```

CONTRIBUTING.md
```diff
@@ -10,10 +15,59 @@ This project has adopted the [Microsoft Open Source Code of Conduct](https://ope
...
+- camelCase for parameters and variables
+- g_ prefix for global variables
+- UPPER_CASE for macros
+## Variables
+- Variables must always be initialized
+- Variables should be defined at the beginning of the code block they're used
+- Variables must not be reused for different purposes, except for simple index variables in for loops
+- Variable names must be verbose enough to explain their purpose if not obvious
+- Use boolean true and false rather than 1 and 0 for boolean operations
+## Error Handling
```
robertwoj-microsoft, line 48: C++: avoid throwing exceptions, but handle when necessary? (e.g. stoi throws, that should be handled)

CONTRIBUTING.md
```diff
@@ -10,10 +10,54 @@ This project has adopted the [Microsoft Open Source Code of Conduct](https://ope
...
+- Avoid side-effects in if statements
+- Avoid ternary operators with complex statements
+## Naming Conventions
+- PascalCase for functions, types, and structs
+- camelCase for parameters and variables
+- g_ prefix for global variables
+- UPPER_CASE for macros
+## Variables
+- Variables must always be properly initialized
+- Variables should be defined at the beginning of the code block they're used
```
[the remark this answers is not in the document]
robertwoj-microsoft: Also, for C only, preferred in C++

CONTRIBUTING.md
```diff
@@ -10,10 +10,226 @@ This project has adopted the [Microsoft Open Source Code of Conduct](https://ope
...
+              return y;
+            }
+
+    - Good:
+
+            int foo(int bar)
+            {
+              int y = 5;
+              if (1 == bar)
+                          {
```
robertwoj-microsoft, line 196: Fix indentation

CONTRIBUTING.md
```diff
@@ -10,10 +10,226 @@ This project has adopted the [Microsoft Open Source Code of Conduct](https://ope
...
+- For C, return integer status codes for functions that can fail, typically using standard errno values, and 0 for success
+- For C++ using complex/optional error types is allowed
+- Always validate function inputs
+- Use early returns for validation and error conditions
+- Log errors with descriptive messages using OsConfigLogError or OsConfigLogInfo
+- For C++ avoid throwing exceptions, but handle them properly when thrown by external code
+## Memory Management
+- Always check for memory allocation failures, fail gracefully if possible
+- Avoid dangling invalid pointers - if it's freed it should be NULL, use FREE_MEMORY(ptr) macro to safely free memory.
+- Free all allocated memory before function returns, including error paths
```
robertwoj-microsoft, line 217: Except output parameters

AhmedBM (approved):

CONTRIBUTING.md
```diff
@@ -10,10 +10,13 @@ This project has adopted the [Microsoft Open Source Code of Conduct](https://ope
 
 We welcome contributions to OSConfig. The main way of contributing to and extending OSConfig is via developing new **Management Modules**. For more information see [OSConfig Management Modules](docs/modules.md).
 Pull requests with few exceptions must contain appropriate unit-tests. We cannot allow test coverage to go down. Pull requests containing code changes without accompanying unit tests may be rejected.
-Pull requests needs to be formatted according to .pre-commit-config.yaml. Each pull request is checked by [Forrmating Tests](https://github.com/Azure/azure-osconfig/blob/dev/.github/workflows/formatting.yml).
```
kkanas, line 13: Should we leave this here - or maybe add `All Pull requests need to be formatted according to [Stye document)(Style.md)`

STYLE.md
```diff
@@ -0,0 +1,228 @@
...
+- For C++ use smart pointers wherever possible, avoid new and delete
+
+# Logging
+- Include an OsConfigLogHandle log parameter in functions that need to log
+- Use appropriate log levels:
+  - OsConfigLogDebug: Detailed debugging information
+  - OsConfigLogInfo: General operational information
+  - OsConfigLogError: Function errors and failures
+- Log return values and status codes for debugging
+- Avoid logging sensitive data, eg. information read from the user machine
```
kkanas, line 215: Should we `Avoid logging sensitive data, eg. information read from the user machine` for any log level or Debug is ok ?

kkanas (approved):

CONTRIBUTING.md
```diff
@@ -10,10 +10,13 @@ This project has adopted the [Microsoft Open Source Code of Conduct](https://ope
...
+
 
 Most contributions require you to agree to a Contributor License Agreement (CLA) declaring that you have the right to, and actually do, grant us the rights to use your contribution. For details, visit https://cla.microsoft.com. When you submit a pull request, a CLA-bot will automatically determine whether you need to provide a CLA and decorate the PR appropriately (e.g., label, comment). Simply follow the instructions provided by the bot. You will only need to do this once across all repositories using our CLA.
 
+# Coding style
+The coding style is described in a separate [style document](STYLE.md).
```
MariusNi, line 18: >[style document](STYLE.md). [](http://example.com/codeflow?start=44&length=27)

Can you please move this under docs rather than keep it in root of the project (also renamed in lowercase as "codingstyle.md)? Thanks

STYLE.md
```diff
@@ -0,0 +1,228 @@
...
+
+            if (NULL == (x = malloc(10))) // Avoid
+            {
+            }
+
+            if ((7 == a) && (NULL == (y = malloc(10)))) // Must not use
+            {
+            }
+
+
```
MariusNi, line 25: Disagree with this

STYLE.md
```diff
@@ -0,0 +1,228 @@
...
+            if (7 == a)
+            {
+              y = malloc(10);
+              if (NULL == y)
+              {
+              }
+            }
+
+- Avoid ternary operators with complex statements
+- Avoid deeply nested if statements, use helper functions, early returns or goto to cleanup code:
```
MariusNi, line 42: >Avoid deeply nested if statements, use helper functions, early returns or goto to cleanup code: [](http://example.com/codeflow?start=2&length=95)

Thanks for putting the time to write this.

I do not agree with making this a blank statement for everything. 

Also, this "STYLE" document goes beyond coding style, practically making some really strict requirements that contradict existing code style for the project (this is my concern).

Existing code style takes precedence over a later added coding style document. Such coding style doc should be written to anchor the existing code style, not to go against it. 

This applies for more strict statements line these, I will not comment on each.

If you want to agree to disagree, please consider moving this STYLE.md to under Compliance and have it enforced there. If keeping for the entire project, sorry but we'll have to agree to disagree (I do not agree with this approach, sorry).

MariusNi (changes requested):
:clock1:

CONTRIBUTING.md
MariusNi, line 18: >separate [style document](STYLE.md). [](http://example.com/codeflow?start=35&length=36)

Can we rename STYLE.md to style.md and move it to src/docs rather than having it here in rot of the project? There is where we have guidance for developers how to write modules (modules.md), etc.

STYLE.md
MariusNi: Avoiding all machine and user data would cripple the logs and would be realistically very difficult in case of logging the raw responses from system commands. We have the debug logging level for this purpose, and we warn the user at beginning of the log when they run with debug level enabled.

CONTRIBUTING.md
MariusNi: I agree, we need CONTRIBUTING.md to keep the ask for pre-commit. I use the command from there myself all the time. Please do not remove,

STYLE.md
```diff
@@ -0,0 +1,228 @@
+# Introduction
+All code needs to be formatted according to .pre-commit-config.yaml. Each pull request is checked by [Formatting Tests](https://github.com/Azure/azure-osconfig/blob/dev/.github/workflows/formatting.yml).
```
MariusNi, line 2: >All code needs to be formatted according to .pre-commit-config.yaml [](http://example.com/codeflow?start=0&length=67)

pre-commit does not format for the coding style. Only does additional and separate clean-up. Having it here sounds misleading as it makes the reader think that by running it this coding style will be automatically applied. Recommend to move it from STYLE back to CONTRIBUTING where it was.

STYLE.md
MariusNi, line 42: >Avoid deeply nested if statements, use helper functions, early returns or goto to cleanup code: [](http://example.com/codeflow?start=2&length=95)

This is not inline with existing code, with 2 exceptions:

1. The NRP code which follows a slightly different coding style due to how it was generated and modeled the first time. Also because of the code there the pattern to 'if (falure) then goto cleanup' works best.

2. The existing code does early returns for input parameter validation. We need to anchor this one in the STYLE doc.

STYLE.md
```diff
@@ -0,0 +1,228 @@
...
+              }
+              return y;
+            }
+
+- Variables must not be reused for different purposes, except for simple index variables in for loops and a function-wide result variable
+- Variable names must be verbose enough to explain their purpose if not obvious
+- Use boolean true and false rather than 1 and 0 for boolean operations
+
+# Error Handling
+- For C, return integer status codes for functions that can fail, typically using standard errno values, and 0 for success
```
MariusNi, line 194: >For C, return integer status codes for functions that can fail, typically using standard errno values, and 0 for success [](http://example.com/codeflow?start=2&length=120)

... or boolean, for functions that make sense (fictive example: bool IsSomethingEnabled)

STYLE.md
```diff
@@ -0,0 +1,228 @@
...
+- For C++ using complex/optional error types is allowed
+- Always validate function inputs
+- Use early returns for validation and error conditions
+- Log errors with descriptive messages using OsConfigLogError or OsConfigLogInfo
```
MariusNi, line 198: >Log errors with descriptive messages using OsConfigLogError or OsConfigLogInfo [](http://example.com/codeflow?start=2&length=78)

... or OsConfigLogWarning etc

STYLE.md
```diff
@@ -0,0 +1,228 @@
...
+- Free all allocated memory (except output values) before function returns, including error paths
+- Functions allocating memory should document ownership transfer to caller
+- For C++ use smart pointers wherever possible, avoid new and delete
+
+# Logging
+- Include an OsConfigLogHandle log parameter in functions that need to log
+- Use appropriate log levels:
+  - OsConfigLogDebug: Detailed debugging information
+  - OsConfigLogInfo: General operational information
+  - OsConfigLogError: Function errors and failures
```
MariusNi, line 213: > OsConfigLogError: Function errors and failures [](http://example.com/codeflow?start=3&length=47)

Also Warning.

STYLE.md
MariusNi, line 215: >Avoid logging sensitive data, eg. information read from the user machine [](http://example.com/codeflow?start=2&length=72)

Unless logging level is debug. At debug level we log everything that is necessary for debugging. Such as, the raw text from system commands (which itself can contain machine or user data)

STYLE.md
```diff
@@ -0,0 +1,228 @@
...
+- Log return values and status codes for debugging
+- Avoid logging sensitive data, eg. information read from the user machine
+
+# Comments and Documentation
+- Each source code file starts with a copyright notice
+- Document non-obvious behavior or complex algorithms
+- ...but don't be too verbose (if it can be easily found in man page it doesn't have to be documented)
+
+# Testing
+- All the new code must have reasonable unit test coverage
```
MariusNi, line 223: >reasonable unit test coverage [](http://example.com/codeflow?start=29&length=29)

test coverage. Which can be either unit-test or functional test (via module test recipes). For example the functions that deal with user accounts, those cannot be fully tested in UTs, but they can be tested in test recipes.

STYLE.md
```diff
@@ -0,0 +1,228 @@
...
+- Unit tests must:
+  - not depend on each other nor interfere with each other (hint: use --gtest_shuffle)
+  - be confined and not interfere with the system
+  - not require special privileges if not needed
+    - if privileges are required (e.g. testing changing file access) the test must detect that it's not being run privileged and be skipped
```
MariusNi, line 228: >the test must detect that it's not being run privileged and be skipped [](http://example.com/codeflow?start=69&length=70)

This would hide problems.

STYLE.md
```diff
@@ -0,0 +1,228 @@
...
+  - OsConfigLogError: Function errors and failures
+- Log return values and status codes for debugging
+- Avoid logging sensitive data, eg. information read from the user machine
+
+# Comments and Documentation
+- Each source code file starts with a copyright notice
+- Document non-obvious behavior or complex algorithms
+- ...but don't be too verbose (if it can be easily found in man page it doesn't have to be documented)
+
+# Testing
```
MariusNi, line 222: >Testing [](http://example.com/codeflow?start=2&length=7)

This section is beyond coding style.

STYLE.md
MariusNi, line 228: >if [](http://example.com/codeflow?start=6&length=2)

We need a few more STYLE details, such as:

1. How to close {}. 

```
if (foo)
{
    ...
}
```
instead of:

```
if (foo) {
    ...
}
```

2. Add a space before and after each operator.

Good:

```
if (a == b)
```

Bad:

```
if (a==b)
```

3. Use inner parenthesis, never rely on just operator precedence.

Good:

```
if ((a < b) && (c != d))
```

Bad:

```
if (a < b && c != d)
```

4. Log all errors.

5. Informationally log as much as necessary in order to understand the whole code execution from a log. 

6. Debug log everything that can help for debugging a problem encountered at customers, in testing, etc.

7. Overall, logging should be no less, no more, than what's necessary to investigate any problem without having to debug.

.... there may be more that don't come to mind just now.

Are you looking at this STYLE as something that we can continue to grow and improve?

MariusNi (changes requested):
:clock1:

MariusNi:
Thanks for adding this STYLE doc! I find it overall very well written, clean and concise. 

Apologies for setting 'request changes'. I rarely use this but here it is important that we reach consensus before this PR is merged (and the PR already has 2 signoffs that would allow it to be merged). I will remove this status once we agree.

When adding a coding style document for a project that was in development for 5+ years (as this one is), we want with that document to anchor the style of the existing code, not go against it.

This STYLE document anchors some of the existing code style, but not all, in some cases describing as bad style from existing code. Let's discuss more about these parts to ensure we have a STYLE document that works well for both existing and future code.

One (another) general observation. The document for each case starts with the Bad, and then Good example. In general we want to avoid starting something with bad/what does not work. Also, let's not use the word 'bad' (no code that's correctly written is bad). How about we use labels such as:

'Do this:" and: 'Don't do this:'
Or:  'This is recommended:' and  'This is not recommended'

STYLE.md
```diff
@@ -0,0 +1,228 @@
+# Introduction
```
simonjaeger, line 1: It seems like there are two parallel topics being discussed here: one concerns the style guide itself, and the other pertains to how we should handle the existing pattern in the codebase.

First of all, thank you for emphasizing readability - I think many of the suggestions are good, and having a document like this will help promote consistency as well.

Regarding the second point: has there been any discussion about what to do with the current pattern? Some of the recommendations here are quite different from the approach used so far (as pointed out in other comments). Are we looking to apply these suggestions only to new code, or is there an intention to refactor existing code as well (which does carry some risk)?

STYLE.md
```diff
@@ -0,0 +1,228 @@
...
+              if (7 == x)
+              {
+                return EINVAL;
+              }
+
+              foo = func();
+              if (NULL == foo)
+              {
+                result = ENOMEM;
+                goto cleanup;
```
simonjaeger, line 116: Is this suggest for C++ as well as C?

STYLE.md
wupeka: Most users are running tests in an unprivileged setup, and it doesn't look good when "by default" there are failing tests.
To avoid hiding problems we can set up  our CI so that skipped tests are considered failures.

docs/style.md
```diff
@@ -0,0 +1,223 @@
+# Introduction
+This coding style applies to all of the new code that's introduced into the codebase.
```
pallakatos, line 2: I know it's relative - but would it be possible to somehow define what is "new" and what is "old" code? :) Whether some form of "from date" or something - also - there could be new code added to the old code part of the project like in case of ASB...just a thought...

docs/style.md
```diff
@@ -0,0 +1,223 @@
...
+When fixing old code that might not adhere to this style guide, use common sense to decide whether this code style guide should be applied,
```
pallakatos, line 3: "And clearly mark or put in a comment that the reason you've decided to implement the new code is xyz - to give enough context for the maintainers when commiting such changes - this is particularly for the "old code fixing" context and common sense

docs/style.md
```diff
@@ -0,0 +1,223 @@
...
+based on how much of the content is actually new, but never mix conventions on a function level.
+
+# General formatting
+All code needs to be formatted according to rules defined in .pre-commit-config.yaml - specifally with regards to the clang-format and clang-tidy rules.
```
pallakatos, line 7: can we insert a link the ".pre-commit-config.yaml" ?

docs/style.md
```diff
@@ -0,0 +1,223 @@
...
+Apart from that there are style rules that can't be expressed in clang-format but must be followed nevertheless:
+
+# Formatting quirks
+## Boolean statements (e.g. in if) should explicitly use parenthesis and never rely on operators order:
+  - Do: `((a != b) && (c != d))`
+  - Don't: `(a != b && c != d)`
+## In comparison statements the constant should be on the left of the comparison operator
+  - Do: `if (0 < variable)`
```
pallakatos, line 15: I guess this is to follow the rest of the code guidance right?

docs/style.md
```diff
@@ -0,0 +1,223 @@
...
+              return y;
+            }
+
+- Variables must not be reused for different purposes, except for simple index variables in for loops and a function-wide result variable
+- Variable names must be verbose enough to explain their purpose if not obvious
+- Use boolean true and false rather than 1 and 0 for boolean operations
+
+# Error Handling
+- For C, return integer status codes for functions that can fail, typically using standard errno values, and 0 for success
+- For C++ using complex/optional error types is allowed
```
kkanas, line 189: is suggested ?  maybe mention RAII ?

docs/style.md
```diff
@@ -0,0 +1,223 @@
...
+                free(baz);
+              }
+              return result;
+            }
+
+
+# Naming Conventions
+- PascalCase for files, functions, types, and structs
+- camelCase for parameters and variables
+- g_ prefix for global variables
```
pallakatos, line 149: global variables to be defined with G_HIGHER_CASE or g_lower_case?

docs/style.md
```diff
@@ -0,0 +1,223 @@
...
+- Variables must not be reused for different purposes, except for simple index variables in for loops and a function-wide result variable
+- Variable names must be verbose enough to explain their purpose if not obvious
+- Use boolean true and false rather than 1 and 0 for boolean operations
+
+# Error Handling
+- For C, return integer status codes for functions that can fail, typically using standard errno values, and 0 for success
+- For C++ using complex/optional error types is allowed
+- Always validate function inputs
+- Use early returns for validation and error conditions
+- Log errors with descriptive messages using OsConfigLogError, OsConfigLogWarning, or OsConfigLogInfo
```
pallakatos, line 192: Is it aligned with the latest logging structure which Marius has implemented recently?

docs/style.md
```diff
@@ -0,0 +1,223 @@
...
+- For C++ avoid throwing exceptions, but handle them properly when thrown by external code
```
pallakatos, line 193: It would be nice to show an example here or point somewhere in the code as "properly" could be interpreted very individually...

kkanas:
@wupeka Could you add section about macros - avoid side effects in macros and in C++ prefer templates to macros 
lasty all  'functiona' macros should be in do{ }while(0) if possible.

@wupeka Maybe small note about Use of namespaces in C++

docs/style.md
```diff
@@ -0,0 +1,223 @@
...
+              if (NULL != baz)
+              {
+                free(baz);
+              }
+              return result;
+            }
+
+
+# Naming Conventions
+- PascalCase for files, functions, types, and structs
```
MariusNi, line 147: >files, functions, types, and structs [](http://example.com/codeflow?start=17&length=36)

Consider changing this line as following, to complete names and list:

```
- PascalCase for file names, function names, types (including class and struct), and namespaces.
```

We need the namespaces added here, also struct is a type and we also need there at least class added.

docs/style.md
```diff
@@ -0,0 +1,223 @@
...
+
+# Logging
+- Include an OsConfigLogHandle log parameter in functions that need to log
+- Use appropriate log levels:
+  - OsConfigLogDebug: Detailed debugging information
+  - OsConfigLogInfo: General operational information
+  - OsConfigLogWarning: Non-critical warnings
+  - OsConfigLogError: Function errors and failures
+- Log return values and status codes for debugging
+- Avoid logging sensitive data, eg. information read from the user machine on log levels lower than Debug
```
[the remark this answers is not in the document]
wupeka: "Forbidden" is a very strong word that I'd like to avoid here, otherwise we'd need to get rid of e.g. logging the matching lines in /etc/sysctl.

docs/style.md
wupeka: Done

docs/style.md
wupeka: added "recommended"

docs/style.md
wupeka: Added a comment

docs/style.md
wupeka: Yes

docs/style.md
```diff
@@ -0,0 +1,223 @@
...
+# Error Handling
+- For C, return integer status codes for functions that can fail, typically using standard errno values, and 0 for success
+- For C++ using complex/optional error types is allowed
+- Always validate function inputs
+- Use early returns for validation and error conditions
+- Log errors with descriptive messages using OsConfigLogError, OsConfigLogWarning, or OsConfigLogInfo
+- For C++ avoid throwing exceptions, but handle them properly when thrown by external code
+
+# Memory Management
+- Always check for memory allocation failures, fail gracefully if possible
```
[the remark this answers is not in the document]
wupeka: Usually - crash.

PR #951, opened 2025-04-03
Introduce user-defined parameters quoting

robertwoj-microsoft:
## Description

For the compliance module we accept a basic syntax for key-value pairs as input for init and remediation functions:
`key1=value1 key2=value2`
Currently we only split string by space which is obviously not enough for cases where the value contains spaces or equality sign.

This PR introduces a possibility to quote values:
`key1="value1" key2='value2'`
Both single quote and double quote are supported.

To handle cases where the quote itself is part of the value, we allow escaping it with a backslash:
`key1="value1\"s" key2="value2"`
To fulfill this, we also support escaping the backslash itself:
`key1="value1\\" key2=value2`
Here the value for KEY1 is `value1\`.

Special thanks to Jeremi for adding fuzzing target for this feature.

## Checklist

### DIFF
--- a/src/modules/compliance/src/lib/Procedure.cpp
+++ b/src/modules/compliance/src/lib/Procedure.cpp
@@ -140,13 +140,25 @@ Optional<Error> Procedure::UpdateUserParameters(const std::string& input)
         {
             const size_t valueStart = pos;
             pos = ParseQuotedValue(input, pos, value);
+
+            // Check if the pos points past a proper quote
             if (input[pos - 1] != input[valueStart])
             {
                 return Error("Invalid key-value pair: missing closing quote or invalid escape sequence");
             }
+
+            // If at the end of the input, we need to check if a closing quote was found, e.g.:
+            // k1=" should fail (ParseQuotedValue will terminate at the opening quote and return valueStart+1)
             if ((pos >= input.size()) && (pos - valueStart == 1))
             {
                 return Error("Invalid key-value pair: missing closing quote at the end of the input");
+            }
+
+            // We want to have at least one space after the value, e.g.:
+            // k1="1"k2="v2" should fail
+            if ((pos < input.size()) && (!isspace(input[pos])))
+            {
+                return Error("Invalid key-value pair: space expected after quoted value");
             }
         }
         else
--- a/src/modules/compliance/src/lib/Procedure.cpp
+++ b/src/modules/compliance/src/lib/Procedure.cpp
@@ -115,6 +115,11 @@ Optional<Error> Procedure::UpdateUserParameters(const std::string& input)
         {
             return result.Error();
         }
+        if (keyStart == pos)
+        {
+            return Error("Invalid key-value pair: empty key");
+        }
+
         pos = result.Value();
         auto key = input.substr(keyStart, pos - keyStart);
         if ((pos >= input.size()) || (input[pos] != '='))
--- a/src/modules/compliance/src/lib/Procedure.cpp
+++ b/src/modules/compliance/src/lib/Procedure.cpp
@@ -47,8 +47,9 @@ std::size_t ParseQuotedValue(const std::string& input, std::size_t pos, std::str
     // Assuming the first character is a quote
     // pos is always > 1 as we are passing value which must follow a valid key, assignment character and a quote
     assert(pos > 1);
-    assert(input[pos] == '"');
+    assert((input[pos] == '"') || (input[pos] == '\''));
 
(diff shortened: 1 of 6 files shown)


src/modules/compliance/src/lib/Procedure.cpp
```diff
@@ -5,52 +5,147 @@
...
 #include <parson.h>
-#include <sstream>
 
 namespace compliance
 {
+namespace
+{
+std::size_t SkipSpaces(const std::string& input, std::size_t pos)
+{
+    while (pos < input.size() && isspace(input[pos]))
```
MariusNi, line 17: >pos < input.size() [](http://example.com/codeflow?start=11&length=18)

A new set of braces may be useful here, what do you think? One rule we try to follow is to never rely on compiler operator precedence.

src/modules/compliance/src/lib/Procedure.cpp
```diff
@@ -5,52 +5,147 @@
...
+    {
+        ++pos;
+    }
+    return pos;
+}
+
+Result<std::size_t> ParseKey(const std::string& input, std::size_t pos)
+{
+    bool first = true;
+    while (pos < input.size() && !isspace(input[pos]) && input[pos] != '=')
```
MariusNi, line 27: >while (pos < input.size() && !isspace(input[pos]) && input[pos] != '=') [](http://example.com/codeflow?start=4&length=71)

Add braces to separate conditions? Like the previous comment, I will stop commenting on this in this PR.

src/modules/compliance/src/lib/Procedure.cpp
```diff
@@ -5,52 +5,147 @@
...
+    {
+        if (islower(input[pos]))
+        {
+            return Error("Invalid key: must be uppercase");
```
MariusNi, line 31: >return Error("Invalid key: must be uppercase"); [](http://example.com/codeflow?start=12&length=47)

Can we fix the case and do not fail, basically become case insensitive? Just a question

MariusNi (dismissed):
:shipit:

src/modules/compliance/src/lib/Procedure.cpp
```diff
@@ -5,52 +5,152 @@
...
 
-        if (!value.empty() && value[0] == '"')
+        // Parse value
+        std::string value;
+        if (input[pos] == '"')
         {
-            value.erase(0, 1);
-            while (!value.empty() && value.back() == '\\')
+            const size_t valueStart = pos;
+            pos = ParseQuotedValue(input, pos, value);
```
jepio, line 135: i wanted to capture that single quotes are currently preserved.
e.g. if the input is `X='1'` then the outcome will be `'1'` and not `1`.

src/modules/compliance/src/lib/Procedure.cpp
```diff
@@ -5,52 +5,152 @@
...
+        if (!result.HasValue())
+        {
+            return result.Error();
+        }
+        pos = result.Value();
+        auto key = input.substr(keyStart, pos - keyStart);
+        if ((pos >= input.size()) || (input[pos] != '='))
+        {
+            return Error("Invalid key-value pair: '=' expected");
+        }
```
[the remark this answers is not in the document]
robertwoj-microsoft: fixed: added check for an empty key

src/modules/compliance/src/lib/Procedure.cpp
```diff
@@ -5,52 +5,157 @@
...
+    }
+
+    return pos;
+}
+
+std::size_t ParseQuotedValue(const std::string& input, std::size_t pos, std::string& value)
+{
+    // Assuming the first character is a quote
+    // pos is always > 1 as we are passing value which must follow a valid key, assignment character and a quote
+    assert(pos > 1);
```
kkanas, line 49: throw ValueError instead of assert as it's not checked in runtime

robertwoj-microsoft: This is a contract here. If this was false it'd mean there is a bug in the calling code (which has been already found by the fuzzer thanks to the assertion). Leaving this as is.

src/modules/compliance/src/lib/Procedure.cpp
robertwoj-microsoft: Added support for both types of quotes as per discussion.

robertwoj-microsoft:
Rebased with dev.

jepio (approved):

PR #955, opened 2025-04-07
EnsureSysctl: Initial implementation

kkanas:
## Description

EnsureSysctl validates the runtime sysctl value by reading /proc/sys/sysctl and if that is compliant validates the configuration values to make sure system stays compliant upon next reboot.

The configuration validation is done using
/lib/systemd/systemd-sysctl --cat-config

## Checklist

### DIFF
--- a/src/modules/compliance/src/lib/procedures/ensureSysctl.cpp
+++ b/src/modules/compliance/src/lib/procedures/ensureSysctl.cpp
@@ -1,0 +1,203 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include <CommonUtils.h>
+#include <Evaluator.h>
+#include <Regex.h>
+#include <algorithm>
+#include <iostream>
+#include <string>
+#include <sys/stat.h>
+#include <sys/types.h>
+
+namespace compliance
+{
+
+std::string TrimWhitesSpace(const std::string& str);
+
+AUDIT_FN(EnsureSysctl, "sysctlName:Name of the sysctl:M:^([a-zA-Z0-9_]+[\\.a-zA-Z0-9_-]+)$", "value:Regex that the value of sysctl has to match:M",
+    "test_procfs:Prefix for the /proc/sys from where sysctl will be read")
+{
+    auto log = context.GetLogHandle();
+    char* output = NULL;
+    std::string procfsLocation = "/proc/sys";
+    std::string systemdSysctl = "/lib/systemd/systemd-sysctl";
+
+    auto it = args.find("sysctlName");
+    if (it == args.end())
+    {
+        return Error("Missing 'sysctlName' parameter", EINVAL);
+    }
+    auto sysctlName = std::move(it->second);
+
+    it = args.find("value");
+    if (it == args.end())
+    {
+        return Error("Missing 'value' parameter", EINVAL);
+    }
+    auto sysctlValue = std::move(it->second);
+
+    it = args.find("test_procfs");
+    if (it != args.end())
+    {
+        procfsLocation = std::move(it->second);
+    }
+
+    auto sysctlPath = sysctlName;
+    std::replace(sysctlPath.begin(), sysctlPath.end(), '.', '/');
+    std::string procSysPath = procfsLocation + std::string("/") + sysctlPath;
+
+    output = LoadStringFromFile(procSysPath.c_str(), false, log);
+    if (output == NULL)
+    {
+        context.GetLogstream() << "Fail sysctl: '" << sysctlName << "' NOT compliant. Expected value: '" << sysctlValue << "' got NULL";
+        return false;
+    }
+    std::string sysctlOutput(output);
+    FREE_MEMORY(output);
+
+    // remove newline caracter
+    if (*sysctlOutput.rbegin() == '\n')
+    {
+        sysctlOutput.erase(sysctlOutput.size() - 1);
+    }
+
+    regex valueRegex;
+    try
+    {
+        valueRegex = regex(sysctlValue);
+    }
+    catch (const std::exception& e)
+    {
+        OsConfigLogError(log, "Regex error: %s", e.what());
+        return Error("Failed to compile regex '" + sysctlValue + "' error: " + e.what());
+    }
+
+    if (regex_search(sysctlOutput, valueRegex) == false)
+    {
+        context.GetLogstream() << "Fail sysctl: '" << sysctlName << "' NOT compliant. Expected value: '" << sysctlValue << "' got '" << sysctlOutput << "'";
+        return false;
+    }
+    if (args.find("test_procfs") == args.end())
+    {
+        struct stat statbuf;
+        if (0 != stat(systemdSysctl.c_str(), &statbuf))
+        {
+            systemdSysctl = std::string("/usr/lib/systemd/systemd-sysctl");
+            if (0 != stat(systemdSysctl.c_str(), &statbuf))
+            {
+                OsConfigLogError(log, "Failed to locate systemdSysctl command");
+                return Error("Failed to locate systemdSysctl command");
+            }
+        }
+    }
+    systemdSysctl += " --cat-config";
+    // systemd-sysclt shows all configs used by system that have sysctl's
+    if ((0 != ExecuteCommand(NULL, systemdSysctl.c_str(), false, false, 0, 0, &output, NULL, log)) || (output == NULL))
+    {
+        OsConfigLogError(log, "Failed to execute systemdSysctl command");
+        return Error("Failed to execute systemdSysctl command");
+    }
+    std::string sysctlConfigs(output);
+    FREE_MEMORY(output);
+
+    std::istringstream configurations(sysctlConfigs);
+    std::string line;
+    std::string runSysctlValue;
+    std::vector<std::string> sysctlConfigLines;
+
+    // Reverse order of output, to check sysctls from most important till less important
+    while (std::getline(configurations, line))
+    {
+        sysctlConfigLines.push_back(line);
+    }
+
+    bool found = false;
+    bool invalid = false;
+    auto lines = sysctlConfigLines.rbegin();
+    // Iterate sysctlConfigLines backwards, when last value matches expected valueRegex
+    // we are done.
+    for (; !found && lines != sysctlConfigLines.rend(); ++lines)
+    {
+        line = *lines;
+        size_t comment = line.find('#');
+        if (comment != std::string::npos)
+        {
+            line = line.substr(0, comment);
+        }
+        if (line == "")
+        {
+            continue;
+        }
+
+        // TODO(kkanas) consider using regex_match when we have all platforms with up to date support
+        auto nameValuePattern = regex("\\s*([a-zA-Z0-9_]+[\\.a-zA-Z0-9_-]+)\\s*=\\s*(.*)");
+        if (regex_search(line, nameValuePattern))
+        {
+            size_t eqPos = line.find('=');
+            if (eqPos == std::string::npos)
+            {
+                continue;
+            }
+
+            std::string name = line.substr(0, eqPos);
+            name = TrimWhitesSpace(name);
+            runSysctlValue = line.substr(eqPos + 1, line.size() - eqPos);
+            runSysctlValue = TrimWhitesSpace(runSysctlValue);
+
+            if (name == sysctlName)
+            {
+                found = true;
+                if (not regex_search(runSysctlValue, valueRegex))
+                {
+                    invalid = true;
+                }
+            }
+        }
+    }
+    // we found a match with correct value, or no match at all was found
+    if (!invalid || !found)
+    {
+        context.GetLogstream() << "sysctl: '" << sysctlName << "' compliant with value: '" << sysctlValue << "'";
+        return true;
+    }
+
+    // lines are iterated backwards so filename is before last value marked by lines
+    std::string fileName;
+    for (; lines != sysctlConfigLines.rend(); ++lines)
+    {
+        line = *lines;
+        // TODO(kkanas) consider using regex_match when we have all platforms with up to date support
+        regex fileNamePattern("^\\s*#\\s*(\\/.*\\.conf)$");
+        if (regex_search(line, fileNamePattern))
+        {
+            size_t comment = line.find('#');
+            if (comment == std::string::npos)
+            {
+                continue;
+            }
+            size_t slash = line.find_first_of('/', comment + 1);
+            if (slash == std::string::npos)
+            {
+                continue;
+            }
+            fileName = line.substr(slash, line.size() - slash + 1);
+            break;
+        }
+    }
+    context.GetLogstream() << "ERROR sysctl: '" << sysctlName << "' NOT compliant expected value: '" << sysctlValue << "' got '" << runSysctlValue
+                           << "' found in: '" << fileName << "'";
+    return false;
+}
+
+std::string TrimWhitesSpace(const std::string& str)
+{
+    auto start = std::find_if_not(str.begin(), str.end(), ::isspace);
+    auto end = std::find_if_not(str.rbegin(), str.rend(), ::isspace).base();
+    if (start < end)
+    {
+        return std::string(start, end);
+    }
+    return std::string();
+}
+} // namespace compliance
--- a/src/modules/compliance/src/lib/GenJSONSchemas.py
+++ b/src/modules/compliance/src/lib/GenJSONSchemas.py
@@ -72,6 +72,8 @@ def process_procedures(content, prefix):
             args.append(current)
         fn_name, arg_dict = parse_macro_args(args)
         if fn_name:
+            if fn_name[0].islower():
+                raise ValueError(f"Error function name '{fn_name}' starts with Lowercase in '{prefix}'")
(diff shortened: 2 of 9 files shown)


src/modules/compliance/src/lib/procedures/ensureSysctl.cpp
```diff
@@ -0,0 +1,159 @@
...
+#include <Regex.h>
+#include <algorithm>
+#include <iostream>
+#include <string>
+
+namespace compliance
+{
+
+AUDIT_FN(ensureSysctl, "sysctlName:Name of the kernel module:M", "value:Regexp value Sysctl that needst to be match",
+    "test_procfs:Prefix for the /proc/sys from where sysctl will be read", "test_systemd_sysctl:Path where systemd-sysctl tool is installed")
```
wupeka, line 15: Do we need test_systemd_sysctl with ExecuteCommand mocks in place? (note: I did that for FilesystemOptions because we hadn't had the mocks then, and plan to move to mock EC calls)

src/modules/compliance/src/lib/procedures/ensureSysctl.cpp
```diff
@@ -0,0 +1,159 @@
...
+    // TODO(kkanas): pass osconfig log correctly
+    OsConfigLogHandle a_log = NULL;
+    output = LoadStringFromFile(procSysPath.c_str(), false, a_log);
+    if (output == NULL)
+    {
+        logstream << "Fail sysctl: '" << sysctlName << "' NOT compliant. Expected value: '" << sysctlValue << "' got NULL";
+        return false;
+    }
+    std::string sysctlOutput(output);
+    free(output);
```
wupeka, line 57: Use FREE_MEMORY macro.

src/modules/compliance/src/lib/procedures/ensureSysctl.cpp
```diff
@@ -0,0 +1,159 @@
...
+    output = NULL;
+
+    if (sysctlOutput.empty())
```
wupeka, line 60: Do we need the empty string it as a special case? Won't the regex "catch" it anyways?

src/modules/compliance/src/lib/procedures/ensureSysctl.cpp
```diff
@@ -0,0 +1,159 @@
...
+        std::string sysctlConfigs(output);
+        free(output);
+        output = NULL;
+
+        std::istringstream configurations(sysctlConfigs);
+        std::string line;
+        std::string runSysctlValue;
+        std::vector<std::string> sysctlConfigLines;
+
+        // Reverse order of output, to check sysctls from most important till less importat
```
wupeka, line 97: typo: important

src/modules/compliance/src/lib/procedures/ensureSysctl.cpp
```diff
@@ -0,0 +1,159 @@
...
+            {
+                std::string name = match[1];
+                runSysctlValue = match[2];
+                if (name == sysctlName && not regex_match(runSysctlValue, match, valueRegexp))
+                {
+                    done = true;
+                }
+            }
+        }
+        if (done == false)
```
wupeka, line 130: nit: !done

src/modules/compliance/src/lib/procedures/ensureSysctl.cpp
```diff
@@ -0,0 +1,159 @@
...
+            if (regex_match(line, match, fileNamePattern))
+            {
+                fileName = match[1];
+                break;
+            }
+        }
+        logstream << "ERROR sysctl: '" << sysctlName << "' NOT compliant expected value: '" << sysctlValue << "' got '" << runSysctlValue << "'";
+        return false;
+    }
+    catch (const std::exception& e)
```
wupeka, line 152: We have a global catch with proper handling already in the Interface, so in the procedures just catch specific errors, as narrowly as possible.

src/modules/compliance/src/lib/procedures/ensureSysctl.cpp
```diff
@@ -0,0 +1,159 @@
...
+        }
+        if (done == false)
+        {
+            logstream << "sysctl: '" << sysctlName << "' compliant with value: '" << sysctlValue << "'";
+            return true;
+        }
+
+        // lines are iterated backwards so filename is before current lines iterator
+        std::string fileName;
+        for (; lines != sysctlConfigLines.rend(); ++lines)
```
wupeka, line 138: What does this loop do?

src/modules/compliance/src/lib/procedures/ensureSysctl.cpp
```diff
@@ -0,0 +1,159 @@
...
+        systemdSysctl += " --cat-config";
+        // systemd-sysclt shows all configs used by system that have sysctl's
+        if ((0 != ExecuteCommand(NULL, systemdSysctl.c_str(), false, false, 0, 0, &output, NULL, NULL)) || (output == NULL))
+        {
+            logstream << "Failed to execute systemdSysctl command";
+            return Error("Failed to execute systemdSysctl command");
+        }
+        std::string sysctlConfigs(output);
+        free(output);
+        output = NULL;
```
wupeka, line 90: FREE_MEMORY()

src/modules/compliance/src/lib/procedures/ensureSysctl.cpp
```diff
@@ -0,0 +1,159 @@
...
+        auto lines = sysctlConfigLines.rbegin();
+        for (; !done && lines != sysctlConfigLines.rend(); ++lines)
+        {
+            line = *lines;
+            size_t comment = line.find('#');
+            if (comment != std::string::npos)
+            {
+                line = line.substr(0, comment);
+            }
+            if (line == "")
```
wupeka, line 113: Isn't that redundant? We simply won't have a match for an empty string.

src/modules/compliance/src/lib/procedures/ensureSysctl.cpp
kkanas: I am not sure. I will check what can do in moduletest and if our doceker supports systemd. In worst case I can write shell script located in //lib/systemd-sysctl that reads set of configuration files

src/modules/compliance/src/lib/procedures/ensureSysctl.cpp
kkanas: I will remove it. It made sense when I read output from sysctl program - which reports the sysctl name = sysctl value.

src/modules/compliance/src/lib/procedures/ensureSysctl.cpp
kkanas: sure but why build regexp and do more work if we after removing comment line got empty string

src/modules/compliance/src/lib/procedures/ensureSysctl.cpp
kkanas: as the comment says on line 136:  lines are iterated backwards so filename is before current lines iterator.

I think this can be useful. We have sysctl that violates policy but we don't know which file caused the violation. 
As the lines are iterated backwards the lines iterator is at current sysctl that is not compliant, iterating backwards will give you filename .
The output from `/lib/systemd-sysctl` is 
```
# /etc/sysctl.d/some.file.conf
net.ipv4.ip_forward = 1
```
so if the net.ipv4.ip_forward is to be equal 0 we need to reiterated backwards to get the filename.
I can remove it or add better comment

src/modules/compliance/src/lib/procedures/ensureSysctl.cpp
kkanas: The try catch is only for invalid regular expression for sysctl value. I think I can make it more narrow if I move regexp creation to the top.

src/modules/compliance/src/lib/procedures/ensureSysctl.cpp
wupeka: OK, but where do we use the obtained fileName?

src/modules/compliance/src/lib/procedures/ensureSysctl.cpp
wupeka: In this case I'd say that programmers brain cycles are worth more than the CPU cycles :) But leaving it just as a nit.

src/modules/compliance/src/lib/procedures/ensureSysctl.cpp
kkanas: In the log line ;) which I forgot to add ;P

src/modules/compliance/src/lib/procedures/ensureSysctl.cpp
kkanas: Fixed

src/modules/compliance/src/lib/procedures/ensureSysctl.cpp
kkanas: Added

src/modules/compliance/src/lib/procedures/ensureSysctl.cpp
```diff
@@ -0,0 +1,152 @@
...
+
+    regex valueRegex;
+    try
+    {
+        valueRegex = regex(sysctlValue);
+    }
+    catch (const std::exception& e)
+    {
+        OsConfigLogError(nullptr, "Regex error: %s", e.what());
+        return false;
```
wupeka, line 71: I'd actually return an Error here, since that's what it is, and we will have proper handling of those in the interface (e.g. won't stop the whole benchmark).

src/modules/compliance/src/lib/procedures/ensureSysctl.cpp
```diff
@@ -0,0 +1,152 @@
...
+#include <Evaluator.h>
+#include <Regex.h>
+#include <algorithm>
+#include <iostream>
+#include <string>
+
+namespace compliance
+{
+
+AUDIT_FN(ensureSysctl, "sysctlName:Name of the kernel module:M", "value:Regex value Sysctl that needst to be match",
```
wupeka, line 14: 1. s/kernel module/sysctl
2. For the value:  "Regex that the value of sysctl has to match"
3. Add 'M' for the value since it is mandatory
4. Maybe put a pattern for the sysctlName? Something like "^[a-zA-Z0-9_]+(\.[a-zA-Z0-9_]+)+$" ?

wupeka (approved):

src/modules/compliance/src/lib/procedures/ensureSysctl.cpp
kkanas: I tried regex - I think better one is [a-zA-Z0-9_]+([\.a-zA-Z0-9_-] that matches e.g.  debug.exception-trace .
Still that doesn't match 
fs.binfmt_misc.python3/10
fs.binfmt_misc.llvm-14-runtime/binfmt
which in turn translate to  /proc/sys/fs/binfmt_misc/python3.10 and /proc/sys/fs/binfmt_misc/llvm-14-runtime.binfmt. 
Which I don't have support in sysctlName -> procfs path yet.

src/modules/compliance/src/lib/procedures/ensureSysctl.cpp
kkanas: I removed the test_systemd_sysctl managed to get around it using mocks for execute command

src/modules/compliance/src/lib/procedures/ensureSysctl.cpp
```diff
@@ -0,0 +1,201 @@
...
+#include <string>
+#include <sys/stat.h>
+#include <sys/types.h>
+
+namespace compliance
+{
+
+std::string TrimWhitesSpace(const std::string& str);
+
+AUDIT_FN(ensureSysctl, "sysctlName:Name of the sytsctl:M", "value:Regex that the value of sysctl has to match:M",
```
wupeka, line 18: 1. Typo "sytsctl"
2. Can we have a pattern to match the name of the sysctl?

src/modules/compliance/src/lib/procedures/ensureSysctl.cpp
```diff
@@ -0,0 +1,201 @@
...
+    auto sysctlName = std::move(it->second);
+
+    it = args.find("value");
+    if (it == args.end())
+    {
+        return Error("Missing 'value' parameter", EINVAL);
+    }
+    auto sysctlValue = std::move(it->second);
+
+    if (args.find("test_procfs") != args.end())
```
wupeka, line 39: To use the same schema it = args.find(), if it != end  procfsLocation = std::move(it->second)

wupeka (approved):

src/modules/compliance/src/lib/procedures/ensureSysctl.cpp
kkanas: I didn't use std::move(it->second); as the test to search for     if (args.find("test_procfs") == args.end())
 is used to detect we are in tests environment and don't need to check for alternate locations of `systemd-sysctl`.

src/modules/compliance/src/lib/procedures/ensureSysctl.cpp
```diff
@@ -0,0 +1,202 @@
...
+        }
+    }
+    logstream << "ERROR sysctl: '" << sysctlName << "' NOT compliant expected value: '" << sysctlValue << "' got '" << runSysctlValue << "' found in: '"
+              << fileName << "'";
+    return false;
+}
+
+std::string TrimWhitesSpace(const std::string& str)
+{
+
```
AhmedBM, line 193: nit: empty newline

AhmedBM (approved):

MariusNi (dismissed):
:shipit:

src/modules/compliance/src/lib/procedures/ensureSysctl.cpp
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    regex valueRegex;
+    try
+    {
+        valueRegex = regex(sysctlValue);
+    }
+    catch (const std::exception& e)
+    {
+        OsConfigLogError(nullptr, "Regex error: %s", e.what());
```
robwoj, line 71:
```suggestion
        OsConfigLogError(log, "Regex error: %s", e.what());
```

src/modules/compliance/src/lib/procedures/ensureSysctl.cpp
```diff
@@ -0,0 +1,202 @@
...
+    }
+    if (args.find("test_procfs") == args.end())
+    {
+        struct stat statbuf;
+        if (0 != stat(systemdSysctl.c_str(), &statbuf))
+        {
+            systemdSysctl = std::string("/usr/lib/systemd/systemd-sysctl");
+            if (0 != stat(systemdSysctl.c_str(), &statbuf))
+            {
+                return Error("Failed to execute systemdSysctl command");
```
robwoj, line 88: Add OsConfigLogError

src/modules/compliance/src/lib/procedures/ensureSysctl.cpp
```diff
@@ -0,0 +1,202 @@
...
+            }
+        }
+    }
+    systemdSysctl += " --cat-config";
+    // systemd-sysclt shows all configs used by system that have sysctl's
+    if ((0 != ExecuteCommand(NULL, systemdSysctl.c_str(), false, false, 0, 0, &output, NULL, NULL)) || (output == NULL))
+    {
+        logstream << "Failed to execute systemdSysctl command";
```
robwoj, line 96: Replace logstream with OsConfigLogError

src/modules/compliance/src/lib/procedures/ensureSysctl.cpp
```diff
@@ -0,0 +1,202 @@
...
+            systemdSysctl = std::string("/usr/lib/systemd/systemd-sysctl");
+            if (0 != stat(systemdSysctl.c_str(), &statbuf))
+            {
+                return Error("Failed to execute systemdSysctl command");
+            }
+        }
+    }
+    systemdSysctl += " --cat-config";
+    // systemd-sysclt shows all configs used by system that have sysctl's
+    if ((0 != ExecuteCommand(NULL, systemdSysctl.c_str(), false, false, 0, 0, &output, NULL, NULL)) || (output == NULL))
```
robwoj, line 94: pass log to ExecuteCommand

src/modules/compliance/src/lib/procedures/ensureSysctl.cpp
```diff
@@ -0,0 +1,202 @@
...
+        {
+            line = line.substr(0, comment);
+        }
+        if (line == "")
+        {
+            continue;
+        }
+
+        // TODO(kkanas) consider using regex_match when we have all platforms with up to date support
+        auto nameValuePattern = regex("\\s*([^\\s]*)\\s*=\\s*(.*)");
```
robwoj, line 132: keep consistent with regex compilation and exception handling (see line 65). I'm fine with both handling it specifically to improve error message and relying on global try-catch.

robwoj (approved):

src/modules/compliance/src/lib/procedures/ensureSysctl.cpp
kkanas: Done

MariusNi (approved):
:shipit:

PR #963, opened 2025-04-09
New packageInstalled procedure

wupeka:
## Description
New packageInstalled procedure:
- support for both dpkg and rpm, with autodetection and override
- support for caching of package list
- full test suite

## Checklist

### DIFF
--- a/src/modules/compliance/src/lib/procedures/packageInstalled.cpp
+++ b/src/modules/compliance/src/lib/procedures/packageInstalled.cpp
@@ -15,6 +15,7 @@
 namespace
 {
 
+using compliance::ContextInterface;
 using compliance::Error;
 using compliance::Result;
 
@@ -44,13 +45,15 @@ private:
     std::function<void()> f;
 };
 
-std::string DetectPackageManager(OsConfigLogHandle log)
+std::string DetectPackageManager(ContextInterface& context)
 {
-    if (0 == ExecuteCommand(NULL, "dpkg -l dpkg", false, false, 0, 0, NULL, NULL, log))
+    auto dpkg = context.ExecuteCommand("dpkg -l dpkg");
+    if (dpkg.HasValue())
     {
         return "dpkg";
     }
-    else if (0 == ExecuteCommand(NULL, "rpm -qa rpm", false, false, 0, 0, NULL, NULL, log))
+    auto rpm = context.ExecuteCommand("rpm -qa rpm");
+    if (rpm.HasValue())
     {
         return "rpm";
     }
@@ -111,7 +114,7 @@ Result<PackageCache> LoadPackageCache(const std::string& path)
 
 Result<int> SavePackageCache(const PackageCache& cache, const std::string& path)
 {
-    std::string tempPath = path + ".XXXXXX.tmp";
+    std::string tempPath = path + ".tmp.XXXXXX";
     char pathTemplate[PATH_MAX];
     snprintf(pathTemplate, sizeof(pathTemplate), "%s", tempPath.c_str());
     int fd = mkstemp(pathTemplate);
@@ -158,26 +161,21 @@ Result<int> SavePackageCache(const PackageCache& cache, const std::string& path)
     return 0;
 }
 
-Result<PackageCache> GetInstalledPackagesRpm(OsConfigLogHandle log)
+Result<PackageCache> GetInstalledPackagesRpm(ContextInterface& context)
 {
     PackageCache cache;
     cache.packageManager = "rpm";
     cache.lastUpdateTime = time(NULL);
 
-    const char* cmd = "rpm -qa --qf='%{NAME}\n'";
-    char* output = nullptr;
+    const std::string cmd = "rpm -qa --qf='%{NAME}\n'";
 
-    int rv = ExecuteCommand(nullptr, cmd, false, false, 0, 0, &output, nullptr, log);
-    if (rv != 0 || output == nullptr)
+    auto rpmqa = context.ExecuteCommand(cmd);
+    if (!rpmqa.HasValue())
     {
-        if (output)
-        {
-            free(output);
-        }
         return Error("Failed to execute rpm command");
     }
 
-    std::istringstream stream(output);
+    std::istringstream stream(rpmqa.Value());
     std::string packageName;
     while (std::getline(stream, packageName))
     {
@@ -186,31 +184,24 @@ Result<PackageCache> GetInstalledPackagesRpm(OsConfigLogHandle log)
             cache.packageNames.insert(packageName);
         }
     }
-
-    free(output);
     return cache;
 }
 
-Result<PackageCache> GetInstalledPackagesDpkg(OsConfigLogHandle log)
+Result<PackageCache> GetInstalledPackagesDpkg(ContextInterface& context)
 {
     PackageCache cache;
     cache.packageManager = "dpkg";
     cache.lastUpdateTime = time(NULL);
 
-    const char* cmd = "dpkg -l";
-    char* output = nullptr;
+    const std::string cmd = "dpkg -l";
 
-    int rv = ExecuteCommand(nullptr, cmd, false, false, 0, 0, &output, nullptr, log);
-    if (rv != 0 || output == nullptr)
+    auto dpkgl = context.ExecuteCommand(cmd);
+    if (!dpkgl.HasValue())
     {
-        if (output)
-        {
-            free(output);
-        }
         return Error("Failed to execute dpkg command");
     }
 
-    std::istringstream stream(output);
+    std::istringstream stream(dpkgl.Value());
     std::string line;
     bool headerSkipped = false;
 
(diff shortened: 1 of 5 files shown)


robertwoj-microsoft:
Could you add module tests?

src/modules/compliance/src/lib/procedures/packageInstalled.cpp
```diff
@@ -3,31 +3,357 @@
...
+        return Error("Invalid cache file format");
+    }
+
+    auto separatorPos = header.find('@');
+    if (std::string::npos == separatorPos)
+    {
+        return Error("Invalid cache file header format");
+    }
+
+    cache.packageManager = header.substr(15, separatorPos - pkgCacheHeader.length());
```
robertwoj-microsoft, line 69: use a const or strlen instead of 15

src/modules/compliance/src/lib/procedures/packageInstalled.cpp
```diff
@@ -3,31 +3,357 @@
...
+{
+    PackageCache cache;
+    const std::string pkgCacheHeader = "# PackageCache "; // "# PackageCache <packageManager>@<timestamp>\n"
+    cache.lastUpdateTime = 0;
+    cache.packageManager = "";
+    cache.packageNames.clear();
+    std::ifstream cacheFile(path);
+    if (!cacheFile.is_open())
+    {
+        return Error("Failed to open cache file: " + path);
```
robertwoj-microsoft, line 54: Add OsConfigLogError()

src/modules/compliance/src/lib/procedures/packageInstalled.cpp
```diff
@@ -3,31 +3,357 @@
...
+        {
+            logstream << "No package manager found";
+            return Error("No package manager found");
+        }
+    }
+
+    auto cachePathIt = args.find("test_cachePath");
+    std::string cachePath = cachePathIt != args.end() ? std::move(cachePathIt->second) : PACKAGE_CACHE_PATH;
+
+    PackageCache cache;
```
robertwoj-microsoft, line 285: Could this variable be static so it's initialized once at the first use time? We could save some disk operations in this case if there are multiple calls to packageInstalled procedure.

src/modules/compliance/src/lib/procedures/packageInstalled.cpp
```diff
@@ -3,31 +3,357 @@
...
+}
+
+} // namespace
+
 namespace compliance
 {
 
-AUDIT_FN(packageInstalled, "packageName:Package name:M")
+AUDIT_FN(packageInstalled, "packageName:Package name:M", "packageManager:Package manager, autodetected by default::^(rpm|dpkg)$",
+    "test_cachePath:Cache path")
```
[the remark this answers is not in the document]
wupeka: It'd have to be a part of MOF, and it doesn't feel right there. A way to store "global" configuration would be nice, but that's out of the scope here. Since now it's used only for testing, I'd leave it as it is .

src/modules/compliance/src/lib/procedures/packageInstalled.cpp
wupeka: Done.

src/modules/compliance/src/lib/procedures/packageInstalled.cpp
wupeka: We're logging this at a higher level, and this is not an error from the procedure's perspective.

src/modules/compliance/src/lib/procedures/packageInstalled.cpp
wupeka: Since we sometimes are reloading the library (and would thus be flushing the in mem cache) and sometimes not I'm worried about weird bugs that could happen because the library was reloaded between the specific checks, the on-disk-only approach seems more consistent, and loading a <100kb file, most likely from cache, doesn't seem to be a significant overhead. So I'm open, but not convinced.

src/common/commonutils/CommandUtils.c
```diff
@@ -85,7 +85,10 @@ int ExecuteCommand(void* context, const char* command, bool replaceEol, bool for
         }
         if (0 == strncmp(mock->expectedCommand, command, stringLen))
         {
-            *textResult = DuplicateString(mock->output);
+            if ((NULL != textResult) && (NULL != mock->output))
```
AhmedBM, line 88: this is fixed in another PR

AhmedBM (approved):

src/modules/compliance/src/lib/procedures/packageInstalled.cpp
```diff
@@ -3,31 +3,357 @@
...
+
+    std::istringstream stream(output);
+    std::string line;
+    bool headerSkipped = false;
+
+    while (std::getline(stream, line))
+    {
+        if (!headerSkipped)
+        {
+            if (line.find("+++-") == 0)
```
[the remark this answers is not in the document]
wupeka: There is one in the test (packageInstalledTest.cpp)

src/common/commonutils/CommandUtils.c
wupeka: Cherry-picked here as it was crashing.

robertwoj-microsoft (changes requested):
Please add module tests as I believe using real package managers and cache can be verified much better there.

src/modules/compliance/src/lib/procedures/packageInstalled.cpp
```diff
@@ -3,31 +3,357 @@
...
+    }
+    close(fd);
+
+    templatePath = pathTemplate;
+    std::string tempPath = templatePath + ".tmp";
+    std::ofstream tempFile(tempPath, std::ios::out | std::ios::trunc);
+    if (!tempFile.is_open())
+    {
+        return Error("Failed to open temporary file for writing: " + tempPath);
+    }
```
jepio, line 114: This usage of mkstemp seems suspicious because you end up opening a different file (so you're racing), and you never cleanup the file created by `mkstemp` (it doesn't get cleaned up automagically).
How about something more like this:
```
int fd = mkstemp(...)
if (fd == -1)
{
...
}
std::ofstream tempFile(pathTemplate, ...);
close(fd);
```

src/modules/compliance/src/lib/procedures/packageInstalled.cpp
```diff
@@ -3,31 +3,357 @@
...
+    std::string packageName;
+    while (std::getline(cacheFile, packageName))
+    {
+        if (!packageName.empty())
+        {
+            cache.packageNames.insert(packageName);
+        }
+    }
+
+    if (cacheFile.bad())
```
jepio, line 88: Some places in this file check for `bad()`, others `fail()`, others `operator bool`. Are you able to distinguish which errors are captured by which? If not then I suggest checking `if (!cacheFile)` or `if (cacheFile)` consistently, which captures any error state. My reference for this stuff is the table here: https://en.cppreference.com/w/cpp/io/basic_ios/bad

src/modules/compliance/src/lib/procedures/packageInstalled.cpp
```diff
@@ -3,31 +3,357 @@
...
+typedef struct
+{
+    time_t lastUpdateTime;
+    std::string packageManager;
+    std::set<std::string> packageNames;
+} PackageCache;
+
+static constexpr long PACKAGELIST_TTL = 3000L;        // just shy of an hours
+static constexpr long PACKAGELIST_STALE_TTL = 12600L; // 3.5 hours
+static constexpr char PACKAGE_CACHE_PATH[] = "/var/lib/osconfig/packageCache";
```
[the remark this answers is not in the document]
wupeka: Path fixed, re: existence - caching will fail if the directory doesn't exist, so isn't that "good enough"?

src/modules/compliance/src/lib/procedures/packageInstalled.cpp
wupeka: Done.

src/modules/compliance/src/lib/procedures/packageInstalled.cpp
wupeka: Replaced fail() with !, we still need bad() as we getline sets failbit and we don't want it to fail

src/modules/compliance/src/lib/procedures/packageInstalled.cpp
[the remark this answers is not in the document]
jepio: The path is still the same.

> caching will fail if the directory doesn't exist, so isn't that "good enough"?

Could be, but without additional handling you'll print at least 2 error to the log for every `packageInstalled` check.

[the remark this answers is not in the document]
AhmedBM: im not against using that path, but wouldn't using the temp dir be better (`P_tmpdir` macro)?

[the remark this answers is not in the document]
wupeka: That location was suggested by Amit

robertwoj-microsoft (approved):

PR #965, opened 2025-04-11
Add -Wshadow=local to prevent coding errors

kkanas:
Add -Wshadow=local to prevent coding error

## Description
Enable -Wshadow to utilize compiler checks better when local variables shadows some wider scope variable it usually can lead to an error.
I spend some time debugging code due the fact of pasting code sniped that I added during refactoring. 

```
int value = 0;
if (importantThing != end) {
   int value = *importatntThing;
   ExecuteStuff(value);
}
return value;
```
## Checklist

### DIFF
--- a/src/modules/compliance/src/lib/CMakeLists.txt
+++ b/src/modules/compliance/src/lib/CMakeLists.txt
@@ -2,7 +2,9 @@
 # Licensed under the MIT License.
 
 project(compliancelib)
-
+if (CMAKE_COMPILER_IS_GNUCC AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 4.9)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wshadow=local")
+endif()
 # Temporary solution, until we stabilize the list of procedures.
 file(GLOB PROCEDURES procedures/*.cpp)
 file(GLOB SCHEMAS procedures/*.schema.json)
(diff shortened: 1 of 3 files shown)


robertwoj-microsoft (approved):
Fix description

pallakatos:
could you please give an example of what it would help to protect against?

src/modules/compliance/src/lib/CMakeLists.txt
```diff
@@ -2,7 +2,9 @@
 # Licensed under the MIT License.
 
 project(compliancelib)
-
+if (CMAKE_COMPILER_IS_GNUCC AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 4.9)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wshadow=local")
```
AhmedBM, line 6: I like the idea of this, any reason why it can't be even more restrictive (just `-Wshadow`)?

AhmedBM (approved):

MariusNi (approved):
:shipit:

src/modules/compliance/src/lib/CMakeLists.txt
kkanas: With -Wshadow compiler complains about 
```
    Error(std::string message, int code)
        : code(code),
          message(std::move(message))
    {
```` 
That code from Error constructor args shadows code member. AFIK this is valid C++ and poses no problems. 
To change that we probably require some convention e.g prefix `m_` for members ... which I think is ok but right now I want to address smaller problem.

PR #968, opened 2025-04-11
Proc modules

jepio:
## Description

In the container environment that we test in `lsmod` is not always available. This may also be the case on customer machines. /proc/modules is guaranteed to be present (it is also the underlying data source for lsmod), and has the same (pretty much) layout (modulename at the start of the line). Switch to parsing `cat /proc/modules` output.

## Checklist


AhmedBM (approved):
👌

jepio:
Closing: will refactor code so that we can mock file contents properly without relying on cat.

PR #972, opened 2025-04-16
EnsureKernelModule: Use /proc/modules for data

jepio:
## Description

In the container environment that we test in lsmod is not always available. This may also be the case on customer machines. /proc/modules is guaranteed to be present (it is also the underlying data source for lsmod), and has the same (pretty much) layout (modulename at the start of the line). Switch to parsing cat /proc/modules output.

## Checklist

### DIFF
--- a/src/modules/compliance/src/lib/Result.cpp
+++ b/src/modules/compliance/src/lib/Result.cpp
@@ -1,0 +1,15 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include "Result.h"
+
+#include <ostream>
+
+namespace compliance
+{
+std::ostream& operator<<(std::ostream& os, const compliance::Error& error)
+{
+    os << "Error: " << error.message << ": code " << error.code;
+    return os;
+}
+} // namespace compliance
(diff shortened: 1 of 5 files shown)


src/modules/compliance/src/lib/Result.cpp
```diff
@@ -0,0 +1,15 @@
...
+
+#include "Result.h"
+
+#include <ostream>
+
+namespace compliance
+{
+std::ostream& operator<<(std::ostream& os, const compliance::Error& error)
+{
+    os << "Error: " << error.message << ": code " << error.code;
```
MariusNi, line 12: >os << "Error: " << error.message << ": code " << error.code; [](http://example.com/codeflow?start=4&length=60)

This gets routed to the correct log file, not to console, correct? What logging level, is error one automatically chosen for this?

Please be aware that we will have the logging level added as a configurable policy parameter, will be good to also have Compliance respect that logging level and log accordingly.

MariusNi (approved):
:shipit:

robertwoj-microsoft (approved):

src/modules/compliance/src/lib/Result.cpp
robertwoj-microsoft: This is not used for logging but constructing payload returned from MmiGet and it's correct here.

PR #987, opened 2025-04-24
EnsureSystemdUnit

kkanas:
## Description

Initial EnsureSystemdUnit to check systed managd service state

## Checklist

### DIFF
--- a/src/modules/compliance/src/lib/procedures/SystemdUnitState.cpp
+++ b/src/modules/compliance/src/lib/procedures/SystemdUnitState.cpp
@@ -1,0 +1,122 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include <CommonUtils.h>
+#include <Evaluator.h>
+#include <Regex.h>
+#include <algorithm>
+#include <iostream>
+#include <string>
+#include <sys/stat.h>
+#include <sys/types.h>
+
+using compliance::Optional;
+
+namespace compliance
+{
+
+// Documentation for dbus {ActiveState, LoadState, UnitFileState} possible values and meaning
+// https://www.freedesktop.org/wiki/Software/systemd/dbus/
+AUDIT_FN(SystemdUnitState, "unitName:Name of the systemd unit:M", "ActiveState:value of systemd ActiveState of unitName to match",
+    "LoadState:value of systemd LoadState of unitName to match", "UnitFileState:value of systemd UnitFileState of unitName to match")
+{
+    struct systemdQueryParams
+    {
+        std::string argName;
+        std::string value;
+        Optional<regex> valueRegex;
+        systemdQueryParams(std::string name)
+            : argName(name),
+              value("")
+        {
+        }
+    };
+    systemdQueryParams params[] = {systemdQueryParams("ActiveState"), systemdQueryParams("LoadState"), systemdQueryParams("UnitFileState")};
+    bool argFound = false;
+    auto log = context.GetLogHandle();
+    std::string systemCtlCmd = "systemctl show ";
+
+    auto it = args.find("unitName");
+    if (it == args.end())
+    {
+        OsConfigLogError(log, "Error: EnsureSystemdUnit: missing 'unitName' parameter ");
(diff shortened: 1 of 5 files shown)


src/modules/compliance/src/lib/procedures/ensureSystemdUnit.cpp
```diff
@@ -0,0 +1,59 @@
...
+#include <sys/stat.h>
+#include <sys/types.h>
+
+namespace compliance
+{
+
+// Documentation for dbus {ActiveState, LoadState, UnitFileState} possible values and meaning
+// https://www.freedesktop.org/wiki/Software/systemd/dbus/
+AUDIT_FN(EnsureSystemdUnit, "unitName:Name of the systemd unit", "activeState:value of systemd ActiveState of unitName to match:M",
+    "loadState:value of systemd LoadState of unitName to match:M", "unitFileState:value of systemd UnitFileState of unitName to match:M")
```
wupeka, line 19: 1. s/EnsureSytemdUnit/SystemdUnitState/ ? 
2. Can we make all of those options optional? e.g. allowing only checking activeState?

wupeka: 3. Can the allowed state be a regex?

kkanas: 1. Ack
2. My plan is to use regex as every argument (except unitName), so if you are intrested in activeState only use loadState=.* unitFileState=.* etc 
3. as per 2

wupeka: re: 2 - why? From the user perspective, those are unnecessary fields.

wupeka: I do understand (and I like) the regex, but I still don't understand why require all three fields if I only want to check one

kkanas: That for me was easy API and code direction but - we can always go for empty string as indicator that caller is not interested in particular field - but this for me looks like more difficult to read - to be sure I would have to check implementation.  If I see regex `.*` it's simpler to reason without looking into implementation. 

But I forgot we can make AUDIT_FN arguments optional.
So the new API would have every arg optional except unitName.
For args values I would still keep regex as this provides more possibility to match many states at once.


What do you think ?

wupeka: SGTM

MariusNi (approved):
:shipit:

src/modules/compliance/src/lib/procedures/SystemdUnitState.cpp
```diff
@@ -0,0 +1,119 @@
...
+    }
+    systemCtlCmd += " " + unitName;
+
+    if (argFound == false)
+    {
+        OsConfigLogError(log, "Error: EnsureSystemdUnit: none of 'activeState loadState UnitFileState' parameters are present");
+        return indicators.NonCompliant("None of 'activeState loadState UnitFileState' parameters are present");
+    }
+
+    if ((0 != ExecuteCommand(NULL, systemCtlCmd.c_str(), false, false, 0, 0, &output, NULL, log)) || (output == NULL))
```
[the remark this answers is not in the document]
kkanas: done

src/modules/compliance/src/lib/procedures/SystemdUnitState.cpp
```diff
@@ -0,0 +1,119 @@
...
+              value("")
+        {
+            valueRegex = regex("");
+        }
+    };
+    systemdQueryParams params[] = {systemdQueryParams("activeState"), systemdQueryParams("loadState"), systemdQueryParams("unitFileState")};
+    bool argFound = false;
+    char* output = NULL;
+    int paramsCount = 0;
+    std::string systemCtlCmd = "systemdctl";
```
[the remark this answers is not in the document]
kkanas: Yes it is done

wupeka (approved):

PR #991, opened 2025-04-24
Add regular expressions sub-pattern functionality fallback

robertwoj-microsoft:
## Description

We have a fallback solution for non-functional std::regex on old and unsupported distributions (RHEL-7/8 and derivatives). This PR adds sub-pattern matching drop-in replacement matching `std::smatch` and adds support for `regex_search(t, m, r)`, previously we only supported `regex_search(t, r)`.

_Note_: Naming convention (CamelCase etc.) is not held on purpose to keep in line with `std::` equivalents.

## Checklist

### DIFF
--- a/src/modules/compliance/src/lib/Regex.h
+++ b/src/modules/compliance/src/lib/Regex.h
@@ -1,68 +1,31 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+#ifndef COMPLIANCE_REGEX_H
+#define COMPLIANCE_REGEX_H
+
 // GCC <= 4.8 has broken std::regex support. This is a very
 // ugly hack that uses libc regexes to emulate std::regex
 // for the functionality that we currently need in compliance.
 // It has to be removed and forgotten after we abandon support
 // for older GCC versions.
 
-#if __cplusplus >= 201103L &&                                                                                                                          \
-    (!defined(__GLIBCXX__) || (__cplusplus >= 201402L) ||                                                                                              \
-        (defined(_GLIBCXX_REGEX_DFS_QUANTIFIERS_LIMIT) || defined(_GLIBCXX_REGEX_STATE_LIMIT) || (defined(_GLIBCXX_RELEASE) && _GLIBCXX_RELEASE > 4)))
+#if __cplusplus >= 201103L && ((__cplusplus >= 201402L) || (defined(_GLIBCXX_REGEX_DFS_QUANTIFIERS_LIMIT) || defined(_GLIBCXX_REGEX_STATE_LIMIT) ||    \
+                                                               (defined(_GLIBCXX_RELEASE) && _GLIBCXX_RELEASE > 4)))
+#define USE_REGEX_FALLBACK 0
+#else
+#define USE_REGEX_FALLBACK 1
+#endif
 
+#if USE_REGEX_FALLBACK == 0
 #include <regex>
 #define regex_search std::regex_search
+#define regex_match std::regex_match
 using regex = std::regex;
+using smatch = std::smatch;
+using regex_error = std::regex_error;
+#else // #if USE_REGEX_FALLBACK == 0
+#include <RegexFallback.h>
+#endif // #if USE_REGEX_FALLBACK == 0
 
-#else
-
-#include <memory>
-#include <regex.h>
-#include <regex>
-#include <string>
-namespace RegexLibcWrapper
-{
-class Regex
-{
-    int ConvertFlags(std::regex_constants::syntax_option_type options)
-    {
-        int flags = 0;
-        if (options & std::regex_constants::icase)
-        {
-            flags |= REG_ICASE;
-        }
-        if (options & std::regex_constants::extended)
-        {
-            flags |= REG_EXTENDED;
-        }
-        return flags;
-    }
-
-public:
-    Regex() = default;
-    Regex(const Regex&) = delete;
-    Regex(Regex&&) noexcept = default;
-    Regex& operator=(Regex&&) noexcept = default;
-    Regex(const std::string& r, std::regex_constants::syntax_option_type options = std::regex_constants::extended)
-    {
-        this->preg = std::unique_ptr<regex_t>(new (regex_t));
-        int v = regcomp(this->preg.get(), r.c_str(), ConvertFlags(options));
-        if (0 != v)
-        {
-            throw std::runtime_error("Invalid regex");
-        }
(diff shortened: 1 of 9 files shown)


src/modules/compliance/src/lib/Regex.h
```diff
@@ -7,62 +7,23 @@
...
-        (defined(_GLIBCXX_REGEX_DFS_QUANTIFIERS_LIMIT) || defined(_GLIBCXX_REGEX_STATE_LIMIT) || (defined(_GLIBCXX_RELEASE) && _GLIBCXX_RELEASE > 4)))
-
-#include <regex>
-#define regex_search std::regex_search
-using regex = std::regex;
+#ifndef COMPLIANCE_REGEX_H
+#define COMPLIANCE_REGEX_H
 
+#if __cplusplus >= 201103L && ((__cplusplus >= 201402L) || (defined(_GLIBCXX_REGEX_DFS_QUANTIFIERS_LIMIT) || defined(_GLIBCXX_REGEX_STATE_LIMIT) ||    \
+                                                               (defined(_GLIBCXX_RELEASE) && _GLIBCXX_RELEASE > 4)))
```
MariusNi, line 14: nit: this line looks a bit strange, may want  either yo not break or to align broken part 4 spaces to left

MariusNi (approved):
:shipit:

src/modules/compliance/src/lib/RegexFallback.h
```diff
@@ -0,0 +1,237 @@
...
+            }
+        }
+    }
+
+    std::string mTarget;
+    std::size_t mSize = 0;
+    std::unique_ptr<regmatch_t[]> mPmatch;
+};
+
+inline bool regexSearch(const std::string& s, MatchResults& m, const Regex& r)
```
kkanas, line 218: Should this be as is replacement if so there is std::regex_match and I can't seem to find regexSearch.

src/modules/compliance/src/lib/RegexFallback.h
```diff
@@ -0,0 +1,237 @@
...
+    bool ready() const
+    {
+        return nullptr != mPmatch;
+    }
+
+    SubMatch operator[](std::size_t i) const
+    {
+        assert(ready());
+        assert(i < mSize);
+        return SubMatch(&mTarget, &mPmatch[i]);
```
kkanas, line 170: std::submatch behavior is different when i > size, like it won't segfault.
https://en.cppreference.com/w/cpp/regex/match_results/operator_at

src/modules/compliance/src/lib/RegexFallback.h
```diff
@@ -0,0 +1,237 @@
...
+          mIndex(index)
+    {
+    }
+
+    const std::string* mTarget = nullptr;
+    const regmatch_t* mPmatch = nullptr;
+    std::size_t mIndex = 0;
+};
+
+class MatchResults
```
kkanas, line 148: std::smatch can be used in std::stream operators aka std::cout << match[0] << std::endl; 
Is that also possible with replacement? 
If the intention is not to support it maybe we could do ensure it with some compilation error or sth?

src/modules/compliance/src/lib/RegexFallback.h
```diff
@@ -0,0 +1,237 @@
...
+}
+
+inline bool regexSearch(const std::string& s, const Regex& r)
+{
+    return (0 == regexec(r.preg.get(), s.c_str(), 0, NULL, 0));
+}
+
+} // namespace RegexLibcWrapper
+
+#define regex_search RegexLibcWrapper::regexSearch
```
kkanas, line 234: What about regex_match ?

src/modules/compliance/src/lib/Regex.h
```diff
@@ -7,62 +7,23 @@
...
-    (!defined(__GLIBCXX__) || (__cplusplus >= 201402L) ||                                                                                              \
-        (defined(_GLIBCXX_REGEX_DFS_QUANTIFIERS_LIMIT) || defined(_GLIBCXX_REGEX_STATE_LIMIT) || (defined(_GLIBCXX_RELEASE) && _GLIBCXX_RELEASE > 4)))
-
-#include <regex>
-#define regex_search std::regex_search
-using regex = std::regex;
+#ifndef COMPLIANCE_REGEX_H
+#define COMPLIANCE_REGEX_H
 
+#if __cplusplus >= 201103L && ((__cplusplus >= 201402L) || (defined(_GLIBCXX_REGEX_DFS_QUANTIFIERS_LIMIT) || defined(_GLIBCXX_REGEX_STATE_LIMIT) ||    \
```
kkanas, line 13: Can we add a comment why we need this glibc version detection ?
Second - maybe we should move this to CMake?

src/modules/compliance/src/lib/RegexFallback.h
robertwoj-microsoft: Omitted intentionally as we don't use it now, if needed I can add one.

robertwoj-microsoft: As per discussion, will add that

src/modules/compliance/src/lib/Regex.h
robertwoj-microsoft: I'd leave it for a separate PR

src/modules/compliance/src/lib/RegexFallback.h
robertwoj-microsoft: OK, will fix then

src/modules/compliance/src/lib/Regex.h
robertwoj-microsoft: Moved the comment closer to the pragma.

src/modules/compliance/src/lib/RegexFallback.h
robertwoj-microsoft: Done.

src/modules/compliance/src/lib/RegexFallback.h
robertwoj-microsoft: Done.

src/modules/compliance/src/lib/RegexFallback.h
robertwoj-microsoft: Done.

src/modules/compliance/src/lib/RegexFallback.h
robertwoj-microsoft: Done.

kkanas (approved):

PR #992, opened 2025-04-24
Add clang static analysis worfklow

jepio:
## Description

Add clang static analysis to our CI. This relies on clang's internal analyzer, called through a script shipped with clang called analyze-build (there exists also a helper tool called scan-build).

This PR also fixes all of the findings. The findings were (in quantity)

* dead stores (writes to variables that are never used or read)
* memory leaks (missing frees on some paths through functions)
* destroying uninitialized objects (OsConfigResource.c)
* mismatched new[]/delete[] (mmi payloads should be free'd using the modules free function)
* use after move in tests disabled using ifdef __clang_analyzer__ (the tests intentionally test to move constructor).

If there are findings in the future, the best way is to download the artifact from the workflow, unpack and open index.html which provides a step-by-step explanation for each finding.

## Checklist

### DIFF
--- a/src/adapters/pnp/PnpAgent.c
+++ b/src/adapters/pnp/PnpAgent.c
@@ -407,7 +407,7 @@ void CloseAgent(void)
     if (NULL != g_mpiHandle)
     {
         CallMpiClose(g_mpiHandle, GetLog());
-        g_mpiHandle = NULL;
+        FREE_MEMORY(g_mpiHandle);
     }
 
     FREE_MEMORY(g_reportedProperties);
--- a/src/adapters/pnp/PnpAgent.c
+++ b/src/adapters/pnp/PnpAgent.c
@@ -406,7 +406,7 @@ void CloseAgent(void)
     if (NULL != g_mpiHandle)
     {
         CallMpiClose(g_mpiHandle, GetLog());
-        FREE_MEMORY(g_mpiHandle);
+        g_mpiHandle = NULL;
     }
 
     FREE_MEMORY(g_reportedProperties);
--- a/src/adapters/pnp/PnpAgent.c
+++ b/src/adapters/pnp/PnpAgent.c
@@ -597,6 +597,9 @@ int main(int argc, char *argv[])
     FREE_MEMORY(cpuType);
     FREE_MEMORY(cpuVendor);
     FREE_MEMORY(cpuModel);
+    FREE_MEMORY(kernelName);
+    FREE_MEMORY(kernelRelease);
+    FREE_MEMORY(kernelVersion);
     FREE_MEMORY(productName);
     FREE_MEMORY(productVendor);
     FREE_MEMORY(encodedProductInfo);
@@ -709,6 +712,7 @@ int main(int argc, char *argv[])
 done:
     OsConfigLogInfo(GetLog(), "OSConfig Agent (PID: %d) exiting with %d", pid, g_stopSignal);
 
+    FREE_MEMORY(jsonConfiguration);
     FREE_MEMORY(g_x509Certificate);
     FREE_MEMORY(g_x509PrivateKeyHandle);
     FREE_MEMORY(connectionString);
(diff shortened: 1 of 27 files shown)


src/adapters/pnp/PnpAgent.c
```diff
@@ -407,7 +406,7 @@ void CloseAgent(void)
     if (NULL != g_mpiHandle)
     {
         CallMpiClose(g_mpiHandle, GetLog());
-        g_mpiHandle = NULL;
+        FREE_MEMORY(g_mpiHandle);
```
MariusNi, line 409: >FREE_MEMORY(g_mpiHandle); [](http://example.com/codeflow?start=8&length=25)

The right place to free the memory for this is in the Modules Manager's MpiClose implementation, which is invoked by this code over the MPI REST API (the MPI Client -> the MPI Server -> the Modules Manager).

src/common/commonutils/SshUtils.c
```diff
@@ -232,7 +231,8 @@ static int IsSshConfigIncludeSupported(OsConfigLogHandle log)
 
     if (NULL != textResult)
     {
-        if (((textPrefixLength = strlen(expectedPrefix)) + 3) < (textResultLength = strlen(textResult)))
+        textPrefixLength = strlen(expectedPrefix) + 3;
+        if (textPrefixLength < strlen(textResult))
```
MariusNi, line 235: Just curious, why is this necessary/why is gets flagged by the analysis?

src/common/logging/Logging.c
```diff
@@ -53,6 +53,7 @@ void SetConsoleLoggingEnabled(bool enabledOrDisabled)
 const char* GetLoggingLevelName(LoggingLevel level)
 {
     const char* result = g_debug;
+    (void)result;
```
MariusNi, line 56: >(void)result; [](http://example.com/codeflow?start=4&length=13)

Line 90 ensures that result is always set after the initialization on line 55. Could you please summarize why is it necessary?

src/common/logging/Logging.c
```diff
@@ -196,8 +197,9 @@ char* GetFormattedTime(void)
 {
     time_t rawTime = {0};
     struct tm* timeInfo = NULL;
+    struct tm result = {0};
     time(&rawTime);
-    timeInfo = localtime(&rawTime);
+    timeInfo = localtime_r(&rawTime, &result);
```
MariusNi, line 202: >localtime_r( [](http://example.com/codeflow?start=15&length=12)

If I am not mistaken, localtime_r does not include tzset (for timezone information) which may need to be separately called (something that localtime does not need)

MariusNi (approved):
:shipit:

src/common/commonutils/SshUtils.c
jepio: The assignments to `textResultLength` and `textPrefixLength` are both not used outside the `if` condition, and that gets flagged as a dead store (a store that is never read).

I opted to remove the `textResultLength` variables as `strlen(textResult)` seems equally readable.
For `textPrefixLength` I found assigning `strlen(expectedPrefix) + 3` to the variable does add some clarity.

src/common/logging/Logging.c
jepio: This gets flagged as dead store, since the value assigned to `result` in line 55 always gets overwritten by all branches (including `default`) in the `switch` statement. But removing the initialization here would be in conflict with our requirement to always initialize variables.

I would agree that this is a false-positive, that's why I added the `(void)result` (it suppresses this warning https://clang.llvm.org/docs/analyzer/user-docs/FAQ.html#dead-store).

Other alternatives didn't work well: removing the `default` label changes nothing as there are cases for all possible enum values, variable initializer still gets overwritten.

src/adapters/pnp/PnpAgent.c
jepio: I pushed the fix here: https://github.com/Azure/azure-osconfig/pull/992/commits/6c2c541180c18ffa9b762f4b5faf3addd2b97127

jepio:
(i temporarily resolved some discussions while rebasing to avoid getting into a broken state again)

jepio:
Pushed again due to another conflict.

kkanas (approved):

PR #995, opened 2025-04-25
Revert FileRegexMatch logic when statePattern is not present

robertwoj-microsoft:
## Description

The FileRegexMatch function is used to match against one or 2 patterns. In case of a single pattern match, the logic was incorrect and it has been fixed.

These are the most notable changes:
- Merge single pattern and double patten modes into one function
- Make matchOperation and stateOperation arguments optional
- Add some debug output

Depends on:
- [ ] #991 
## Checklist

### DIFF
--- a/src/modules/compliance/src/lib/FactExistenceValidator.cpp
+++ b/src/modules/compliance/src/lib/FactExistenceValidator.cpp
@@ -1,0 +1,129 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include <FactExistenceValidator.h>
+#include <cassert>
+#include <map>
+
+namespace compliance
+{
+FactExistenceValidator::FactExistenceValidator(Behavior behavior)
+    : mBehavior(behavior)
+{
+}
+
+void FactExistenceValidator::Finish()
+{
+    if (Done())
+    {
+        return;
+    }
+
+    switch (mBehavior)
+    {
+        case Behavior::AllExist:
+            mState = Status::Compliant;
+            break;
+        case Behavior::NoneExist:
+            assert(!mHasAtLeastOneFact);
+            mState = Status::Compliant;
+            break;
+
+        case Behavior::AtLeastOneExists:
+            assert(!mHasAtLeastOneFact);
+            mState = Status::NonCompliant;
+            break;
+
+        case Behavior::AnyExist:
+            assert(!mHasAtLeastOneFact);
+            mState = Status::Compliant;
+            break;
+
+        case Behavior::OnlyOneExists:
+            mState = mHasAtLeastOneFact ? Status::Compliant : Status::NonCompliant;
+            break;
+    }
+    assert(Done());
+}
+
+void FactExistenceValidator::CriteriaMet()
+{
+    if (Done())
+    {
+        return;
+    }
+
+    switch (mBehavior)
+    {
+        case Behavior::AllExist:
+            break;
+
+        case Behavior::AnyExist:
+            mState = Status::Compliant;
+            break;
+
+        case Behavior::AtLeastOneExists:
+            mState = Status::Compliant;
+            break;
+
+        case Behavior::NoneExist:
+            mState = Status::NonCompliant;
+            break;
+
+        case Behavior::OnlyOneExists:
+            if (mHasAtLeastOneFact)
+            {
+                mState = Status::NonCompliant;
+            }
+            break;
+    }
+    mHasAtLeastOneFact = true;
+}
+
+void FactExistenceValidator::CriteriaUnmet()
+{
+    if (Done())
+    {
+        return;
+    }
+
+    switch (mBehavior)
+    {
+        case Behavior::AllExist:
+            mState = Status::NonCompliant;
+            break;
+        default:
+            break;
+    }
+}
+
+bool FactExistenceValidator::Done() const
+{
+    return mState.HasValue();
+}
+
+Status FactExistenceValidator::Result() const
+{
+    assert(mState.HasValue());
+    return mState.Value();
+}
+
+compliance::Result<FactExistenceValidator::Behavior> FactExistenceValidator::MapBehavior(const std::string& value)
+{
+    static const std::map<std::string, Behavior> sMap = {
+        {"all_exist", Behavior::AllExist},
+        {"any_exist", Behavior::AnyExist},
+        {"at_least_one_exists", Behavior::AtLeastOneExists},
+        {"none_exist", Behavior::NoneExist},
+        {"only_one_exists", Behavior::OnlyOneExists},
+    };
+
+    auto it = sMap.find(value);
+    if (it == sMap.end())
+    {
+        return Error("unsupported value: " + value, EINVAL);
+    }
+
+    return it->second;
+}
+} // namespace compliance
--- a/src/modules/compliance/src/lib/FactExistenceValidator.h
+++ b/src/modules/compliance/src/lib/FactExistenceValidator.h
@@ -1,0 +1,47 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#ifndef COMPLIANCE_FACT_EXISTENCE_VALIDATOR_H
+#define COMPLIANCE_FACT_EXISTENCE_VALIDATOR_H
+
+#include <MmiResults.h>
+#include <Optional.h>
+#include <Result.h>
+#include <string>
+
+namespace compliance
+{
+class FactExistenceValidator
+{
+public:
+    enum class Behavior
+    {
+        AllExist,
+        AnyExist,
+        AtLeastOneExists,
+        NoneExist,
+        OnlyOneExists,
+    };
+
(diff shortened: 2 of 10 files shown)


src/modules/compliance/src/lib/procedures/FileRegexMatch.cpp
```diff
@@ -8,76 +8,86 @@
...
-Result<Status> SinglePatternMatchMode(std::ifstream& input, const std::string& matchPattern, std::regex_constants::syntax_option_type syntaxOptions,
-    IndicatorsTree& indicators)
+// This function is used to check if the file contents match the given pattern.
+// It reads the file line by line and checks each line against the matchPattern.
+// If a line matches the matchPattern, it checks if the statePattern (if provided) also matches.
+// If the matchPattern does not match any line, it marks the status as Compliant.
+// When the matchPattern matches a line, it checks the statePattern:
+//   - if the statePattern is not provided, it marks the status as NonCompliant, otherwise:
+//   - if the statePattern matches all lines, it marks the status as Compliant, otherwise NonCompliant.
+Result<Status> Match(ifstream& input, const string& matchPattern, const Optional<string>& statePattern, syntax_option_type syntaxOptions,
```
kkanas, line 23: Nit maybe better function name then only Match? matchInStream ?

src/modules/compliance/src/lib/procedures/FileRegexMatch.cpp
```diff
@@ -8,76 +8,86 @@
...
+                OsConfigLogDebug(context.GetLogHandle(), "No state pattern provided, marking as non-compliant");
+                indicators.NonCompliant("pattern '" + matchPattern + "' matched line " + std::to_string(lineNumber));
+                status = Status::NonCompliant;
+                break;
             }
-
-            lineNumber++;
         }
     }
     catch (const std::exception& e)
```
kkanas, line 70: Should we try to catch for regex_error first before saying this is regex error, and then for std::exception ?
Or maybe we could change handling so the log and errors are more connected to exception caught ?

src/modules/compliance/src/lib/procedures/FileRegexMatch.cpp
```diff
@@ -8,76 +8,86 @@
...
+// If a line matches the matchPattern, it checks if the statePattern (if provided) also matches.
+// If the matchPattern does not match any line, it marks the status as Compliant.
+// When the matchPattern matches a line, it checks the statePattern:
+//   - if the statePattern is not provided, it marks the status as NonCompliant, otherwise:
+//   - if the statePattern matches all lines, it marks the status as Compliant, otherwise NonCompliant.
+Result<Status> Match(ifstream& input, const string& matchPattern, const Optional<string>& statePattern, syntax_option_type syntaxOptions,
+    IndicatorsTree& indicators, ContextInterface& context)
 {
+    auto status = Status::Compliant;
+    auto mainMatched = false;
```
kkanas, line 27: nit: IIUC mainMatch is used to indicate there was no matches at all so maybe `auto noMatch = true;` and invert if logic at the end of function

src/modules/compliance/src/lib/procedures/FileRegexMatch.cpp
robertwoj-microsoft: Currently we have this:
https://github.com/Azure/azure-osconfig/blob/d33114e0dadf5bfd8a1d7d8f7c713614a7e83eab/src/modules/compliance/src/lib/RegexFallback.h#L46
Should I change it to a fallback regex_error class to capture this?

kkanas: yes - I think we should be exception compliant as well - it will make updating code easier when / if we drop the old platform support.

jepio:
oval has the following attributes for `textfilecontent54_test`:
```
check_existence="all_exist"
check_existence="any_exist"
check_existence="at_least_one_exists"
check_existence="none_exist"
```

jepio:
> oval has the following attributes for `textfilecontent54_test`:
> 
> ```
> check_existence="all_exist"
> check_existence="any_exist"
> check_existence="at_least_one_exists"
> check_existence="none_exist"
> check_existence="only_one_exists" # this is not applicable to textfilecontents54
> ```
> 
> we'll need to review how to feed this information into procedures.

These are also applicable to some other tests, so we need to support this in the parser.

src/modules/compliance/src/lib/procedures/FileRegexMatch.cpp
robertwoj-microsoft: The exception fallback has been added to #991

robertwoj-microsoft:
I have created a `FactExistenceValidator` class which deals with the statuses we have discovered yesterday. It's put externally to `FileRegexMatch` as it has broader scope and should potentially be used by other procedures. The `FileRegexMatch` function updates the validator and finishes loop based on the validator outcome, so the pattern matching logic becomes simpler which I believe is crucial as it already is complex enough.

src/modules/compliance/src/lib/FactExistenceValidator.h
```diff
@@ -0,0 +1,47 @@
...
+
+    void Finish();
+    void CriteriaMet();
+    void CriteriaUnmet();
+    bool Done() const;
+    Status Result() const;
+    static compliance::Result<Behavior> MapBehavior(const std::string& value);
+
+private:
+    Behavior mBehavior;
```
MariusNi, line 41: >mB [](http://example.com/codeflow?start=13&length=2)

We need to add this to STYLE. We should not prefix class members, just use simple camelCased names for variables, and PascalCased names for method members.

Here, would be: 'behavior' and 'state' respectively.

MariusNi: Added a comment there. Let's discuss there and ignore here for now?

src/modules/compliance/src/lib/procedures/FileRegexMatch.schema.json
```diff
@@ -45,6 +44,11 @@
                   "type": "string",
                   "description": "Determine whether the match should be case sensitive, applies to both 'matchPattern' and 'statePattern'",
                   "pattern": "(^\\$[a-zA-Z0-9_]+:(true|false)$|(^true|false$))"
+                },
+                "behavior": {
+                  "type": "string",
+                  "description": "Determine the function behavior",
+                  "pattern": "(^\\$[a-zA-Z0-9_]+:((all_exist|any_exist|at_least_one_exists|none_exist))$|(^(all_exist|any_exist|at_least_one_exists|none_exist)$))"
```
MariusNi, line 51: >(all_exist|any_exist|at_least_one_exists|none_exist)$))" [](http://example.com/codeflow?start=107&length=56)

'all_exist' works. Shouldn't be ''any_exist' 'any_exists' and so on (see the remaining values)?

src/modules/compliance/tests/CMakeLists.txt
```diff
@@ -15,6 +15,7 @@ add_executable(compliancetests
     ComplianceTest.cpp
     EngineTest.cpp
     EvaluatorTest.cpp
+    FactExistenceValidatorTest.cpp
```
MariusNi, line 18: >FactExistenceValidatorTest.cpp [](http://example.com/codeflow?start=4&length=30)

"Validation" instead? Above "Test" it used, not "Tester"

src/modules/compliance/src/lib/procedures/FileRegexMatch.schema.json
kkanas: This maps oval definitions.... https://oval.mitre.org/language/version5.11/ovaldefinition/documentation/oval-common-schema.html#ExistenceEnumeration

src/modules/compliance/src/lib/FactExistenceValidator.cpp
```diff
@@ -0,0 +1,129 @@
...
+            break;
+
+        case Behavior::AnyExist:
+            assert(!mHasAtLeastOneFact);
+            mState = Status::Compliant;
+            break;
+
+        case Behavior::OnlyOneExists:
+            mState = mHasAtLeastOneFact ? Status::Compliant : Status::NonCompliant;
+            break;
```
MariusNi, line 44: >break; [](http://example.com/codeflow?start=12&length=6)

nit: break not necessary here. Q: useful to also add a default option just in case mBehavior would not be recognized (could be a bug too)

src/modules/compliance/src/lib/FactExistenceValidator.cpp
```diff
@@ -0,0 +1,129 @@
...
+        case Behavior::NoneExist:
+            mState = Status::NonCompliant;
+            break;
+
+        case Behavior::OnlyOneExists:
+            if (mHasAtLeastOneFact)
+            {
+                mState = Status::NonCompliant;
+            }
+            break;
```
MariusNi, line 78: >break; [](http://example.com/codeflow?start=12&length=6)

nit: same as above, will stop commenting on these

MariusNi (approved):
:shipit:

kkanas:
Closing as the work was incorporated https://github.com/Azure/azure-osconfig/pull/1024

PR #997, opened 2025-04-25
Test state pattern against the captured value

robertwoj-microsoft:
## Description

`FileRegexMatch` procedure used to test the state pattern against the full line. In general, the state pattern should match against a captured value with a given instance. **The instance is for now skipped and will be added later based on actual needs, for now it defaults to 1.** <- **WARNING**

This depends on:
- [x] #991 
- Outdated by #1024
## Checklist


src/modules/compliance/src/lib/procedures/FileRegexMatch.cpp
```diff
@@ -55,7 +56,11 @@ Result<Status> Match(const std::string& filename, const string& matchPattern, co
             mainMatched = true;
             if (stateRegex.HasValue())
             {
-                if (!regex_search(line, stateRegex.Value()))
+                assert(match.ready());
+                assert(match.size() > 0);
+                auto valueToMatch = match.size() > 1 ? match[1].str() : match[0].str();
+                OsConfigLogDebug(context.GetLogHandle(), "Value to match: %s", valueToMatch.c_str());
```
MariusNi, line 62: >Value t [](http://example.com/codeflow?start=58&length=7)

May be useful to add the function/method name. To keep things simple, we do not automatically capture that (as context we capture the source file name and line number). In a case like this logging like this can be clear for us, but not for a customer that may read our logs.

MariusNi (approved):
:shipit:

src/modules/compliance/src/lib/procedures/FileRegexMatch.cpp
MariusNi: This comment now applies to line 67 (code changed from previous commit)

kkanas:
Closing as I moved content to - #1024

PR #999, opened 2025-04-25
Updated base image for mariner-2

AhmedBM:
## Description

* Updated `mariner-2` base image to use a Gen2 VM as there seems to be disk-space issues with the previous image Gen1 image. Already tested new ES image which fixed previously failing test runs.

## Checklist

### DIFF
--- a/devops/e2e/cloudtest/mariner-2.json
+++ b/devops/e2e/cloudtest/mariner-2.json
@@ -1,4 +1,4 @@
-// Base Image: MicrosoftCBLMariner:cblmariner:cbl-mariner-2:latest
+// Base Image: MicrosoftCBLMariner:cblmariner:cbl-mariner-2-gen2/latest
 {
     "artifacts": [
         {


MariusNi (approved):
:shipit:

simonjaeger (approved):

devops/e2e/cloudtest/mariner-2.json
```diff
@@ -1,4 +1,4 @@
-// Base Image: MicrosoftCBLMariner:cblmariner:cbl-mariner-2:latest
+// Base Image: MicrosoftCBLMariner:cblmariner:cbl-mariner-2-gen2/latest
```
robertwoj-microsoft, line 1: Frankly, I don't understand this. It's basically a comment and I don't see where it is applied. If that's OK, then I'm fine with it, but it looks strange.

AhmedBM: Yes, it is just a comment. This was our way of versioning since 1ES Images do not support ARM deployments; however it would be a good idea to store them as ARM templates as you mention these definitions might be confusing. Ill have a follow-up PR to convert all of these to ARM templates

PR #1003, opened 2025-05-06
Audit procedure for ensureRootPath

wupeka:
## Description
Audit procedure for ensureRootPath

## Checklist

### DIFF
--- a/src/modules/compliance/src/lib/procedures/EnsureRootPath.cpp
+++ b/src/modules/compliance/src/lib/procedures/EnsureRootPath.cpp
@@ -64,7 +64,7 @@ AUDIT_FN(EnsureRootPath)
             auto owner = getpwuid(statbuf.st_uid);
             if (!owner || std::string(owner->pw_name) != "root")
             {
-                return indicators.NonCompliant("Directory '" + path + "' from root's PATH is owned by: " + (owner ? owner->pw_name : "unknown"));
+                return indicators.NonCompliant("Directory '" + path + "' from root's PATH is not owned by root");
             }
 
             if ((statbuf.st_mode & 0777) & ~maxPerm)
@@ -81,6 +81,6 @@ AUDIT_FN(EnsureRootPath)
         }
     }
 
-    return indicators.Compliant("Root's PATH is compliant");
+    return indicators.Compliant("Root's PATH does not contain dangerous entries.");
 }
 } // namespace compliance
--- a/src/modules/compliance/src/lib/procedures/EnsureRootPath.cpp
+++ b/src/modules/compliance/src/lib/procedures/EnsureRootPath.cpp
@@ -1,0 +1,86 @@
+#include <CommonUtils.h>
+#include <Evaluator.h>
+#include <grp.h>
+#include <iostream>
+#include <pwd.h>
+#include <sstream>
+#include <string>
+#include <sys/stat.h>
+#include <unistd.h>
+#include <vector>
+
+namespace compliance
+{
+AUDIT_FN(EnsureRootPath)
+{
+    UNUSED(args);
+    UNUSED(context);
+
+    // Directory must not be world- or group-writable
+    const int maxPerm = 0777 & ~0022;
+    auto rootEnv = context.ExecuteCommand("sudo -Hiu root env");
+    if (!rootEnv.HasValue())
+    {
+        return indicators.NonCompliant("Failed to run sudo: " + rootEnv.Error().message);
+    }
+
+    std::string rootPath;
+    std::istringstream envStream(rootEnv.Value());
+    std::string line;
+    while (std::getline(envStream, line))
+    {
+        if (line.find("PATH=") == 0)
+        {
+            rootPath = line.substr(strlen("PATH="));
+            break;
+        }
+    }
+
+    if (rootPath.empty())
+    {
+        return indicators.NonCompliant("root's PATH is empty");
+    }
+
+    if (rootPath[rootPath.length() - 1] == ':')
+    {
+        return indicators.NonCompliant("Trailing colon in root's PATH");
+    }
+    std::istringstream pathStream(rootPath);
+    std::string path;
+    while (std::getline(pathStream, path, ':'))
+    {
+        if (path.empty())
+        {
+            return indicators.NonCompliant("Empty path in root's PATH");
+        }
+        if ("." == path || ".." == path)
+        {
+            return indicators.NonCompliant("Path in root's PATH is '.' or '..'");
+        }
+
+        struct stat statbuf;
+        if (stat(path.c_str(), &statbuf) == 0 && S_ISDIR(statbuf.st_mode))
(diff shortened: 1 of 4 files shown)


src/modules/compliance/src/lib/procedures/EnsureRootPath.cpp
```diff
@@ -0,0 +1,82 @@
...
+#include <vector>
+
+namespace compliance
+{
+AUDIT_FN(EnsureRootPath)
+{
+    UNUSED(args);
+    UNUSED(context);
+
+    const int maxPerm = 0777 & ~0022;
```
kkanas, line 19: Nit: maybe comment the intent of what maxPerm checks

src/modules/compliance/src/lib/procedures/EnsureRootPath.cpp
```diff
@@ -0,0 +1,82 @@
...
+        {
+            auto owner = getpwuid(statbuf.st_uid);
+            if (!owner || std::string(owner->pw_name) != "root")
+            {
+                return indicators.NonCompliant("Directory '" + path + "' from root's path is owned by: " + (owner ? owner->pw_name : "unknown"));
+            }
+
+            if ((statbuf.st_mode & 0777) & ~maxPerm)
+            {
+                return indicators.NonCompliant("Directory '" + path + "' from root's path has too permissive access");
```
kkanas, line 71: Nit: Add directory permissions to indicator

kkanas (approved):

src/modules/compliance/src/lib/procedures/EnsureRootPath.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+namespace compliance
+{
+AUDIT_FN(EnsureRootPath)
+{
+    UNUSED(args);
+    UNUSED(context);
+
+    const int maxPerm = 0777 & ~0022;
+    auto rootEnv = context.ExecuteCommand("sudo -Hiu root env");
```
AhmedBM, line 20: what if `sudo` isn't present or we are already `root`?

wupeka: We not only need to be uid0, we also need to load all the environment etc etc. that the root loads by default. This is how CIS does it, so I wouldn't reinvent the wheel here.

src/modules/compliance/src/lib/procedures/EnsureRootPath.cpp
wupeka: Done

kkanas (approved):

kkanas (approved):

jepio (commented):
Could you please include module tests with this PR?

src/modules/compliance/src/lib/procedures/EnsureRootPath.cpp
```diff
@@ -0,0 +1,86 @@
...
+namespace compliance
+{
+AUDIT_FN(EnsureRootPath)
+{
+    UNUSED(args);
+    UNUSED(context);
+
+    // Directory must not be world- or group-writable
+    const int maxPerm = 0777 & ~0022;
+    auto rootEnv = context.ExecuteCommand("/usr/bin/sudo -Hiu root env");
```
pallakatos, line 21: Are we relying on sudo here? (it's not standard on every distro) - if we fail to run sudo -  is it automatically return with non-compliance?

wupeka: Yes, we're relying on sudo. Yes, it'll be non-compliant on systems without sudo. It's how CIS does that.

pallakatos: Ok for now let's follow CIS - but let's add some @NOTE here - because probably in future we would want to discuss this with CIS - if there is a better way - or in fact it can be any other comment tag - but it would be nice if we would start to "collect" things which we will want to open a ticket about with CIS or we don't agree with the way they do / or find an error so we can filter on those in the future or have some form of mechanism to collect them..

wupeka: Done.

src/modules/compliance/src/lib/procedures/EnsureRootPath.cpp
```diff
@@ -0,0 +1,86 @@
...
+            return indicators.NonCompliant("Path in root's PATH is '.' or '..'");
+        }
+
+        struct stat statbuf;
+        if (stat(path.c_str(), &statbuf) == 0 && S_ISDIR(statbuf.st_mode))
+        {
+            auto owner = getpwuid(statbuf.st_uid);
+            if (!owner || std::string(owner->pw_name) != "root")
+            {
+                return indicators.NonCompliant("Directory '" + path + "' from root's PATH is owned by: " + (owner ? owner->pw_name : "unknown"));
```
MariusNi, line 67: >owner->pw_name [](http://example.com/codeflow?start=116&length=14)

We no longer log in clear user account names, unless they are system accounts.

src/modules/compliance/src/lib/procedures/EnsureRootPath.cpp
```diff
@@ -0,0 +1,86 @@
...
+                return indicators.NonCompliant(oss.str());
+            }
+        }
+        else
+        {
+            return indicators.NonCompliant("Path '" + path + "' from root's PATH does not exist or is not a directory");
+        }
+    }
+
+    return indicators.Compliant("Root's PATH is compliant");
```
MariusNi, line 84: >is compliant [](http://example.com/codeflow?start=45&length=12)

Sorry, compliant with what? Can this line )(and others like it be more descriptive?

MariusNi (approved):
:shipit:

src/modules/compliance/src/lib/procedures/EnsureRootPath.cpp
```diff
@@ -0,0 +1,86 @@
...
+namespace compliance
+{
+AUDIT_FN(EnsureRootPath)
+{
+    UNUSED(args);
+    UNUSED(context);
+
+    // Directory must not be world- or group-writable
+    const int maxPerm = 0777 & ~0022;
+    auto rootEnv = context.ExecuteCommand("sudo -Hiu root env");
```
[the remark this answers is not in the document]
wupeka: This is not used to gain root privileges but to get the full root environment from the login shell.

src/modules/compliance/src/lib/procedures/EnsureRootPath.cpp
wupeka: Done

src/modules/compliance/src/lib/procedures/EnsureRootPath.cpp
wupeka: Done.

PR #1006, opened 2025-05-12
Add 'Collection' mode to EnsureFilePermission audit and remediation procedures

wupeka:
## Description
Alternative version of EnsureFilePermission that instead of filename argument takes a directory and "extension" (e.g. '*.exe')

## Checklist

### DIFF
--- a/src/modules/compliance/src/lib/procedures/EnsureFilePermissions.cpp
+++ b/src/modules/compliance/src/lib/procedures/EnsureFilePermissions.cpp
@@ -21,9 +21,8 @@ namespace
 {
 // Mask to display permissions
 const unsigned short displayMask = 0xFFF;
-} // namespace
 
-Result<Status> AuditEnsureFilePermissionsHelper(const std::string& filename, std::map<std::string, std::string> args, IndicatorsTree& indicators,
+Result<Status> AuditEnsureFilePermissionsHelper(const std::string& filename, const std::map<std::string, std::string>& args, IndicatorsTree& indicators,
     ContextInterface& context)
 {
     auto log = context.GetLogHandle();
@@ -173,11 +172,12 @@ Result<Status> AuditEnsureFilePermissionsHelper(const std::string& filename, std
     return indicators.Compliant(oss.str());
 }
 
-Result<Status> RemediateEnsureFilePermissionsHelper(const std::string& filename, std::map<std::string, std::string> args, IndicatorsTree& indicators,
-    ContextInterface& context)
+Result<Status> RemediateEnsureFilePermissionsHelper(const std::string& filename, const std::map<std::string, std::string>& args,
+    IndicatorsTree& indicators, ContextInterface& context)
 {
     auto log = context.GetLogHandle();
     struct stat statbuf;
+    std::string owner, groupName;
     if (stat(filename.c_str(), &statbuf) < 0)
     {
         const int status = errno;
@@ -197,12 +197,12 @@ Result<Status> RemediateEnsureFilePermissionsHelper(const std::string& filename,
     auto it = args.find("owner");
     if (it != args.end())
     {
-        auto owner = std::move(it->second);
+        owner = std::move(it->second);
         const struct passwd* pwd = getpwnam(owner.c_str());
         if (pwd == nullptr)
         {
             OsConfigLogDebug(log, "No user with UID %d", statbuf.st_gid);
-            return indicators.NonCompliant("No user with name " + args["owner"]);
+            return indicators.NonCompliant("No user with name " + owner);
         }
         uid = pwd->pw_uid;
         if (uid != statbuf.st_uid)
@@ -218,7 +218,7 @@ Result<Status> RemediateEnsureFilePermissionsHelper(const std::string& filename,
     it = args.find("group");
     if (it != args.end())
     {
-        auto groupName = std::move(it->second);
+        groupName = std::move(it->second);
         const struct group* grp = getgrgid(statbuf.st_gid);
         if (nullptr == grp)
         {
@@ -271,7 +271,7 @@ Result<Status> RemediateEnsureFilePermissionsHelper(const std::string& filename,
             return Error(std::string("Chown error: ") + strerror(status), status);
         }
 
-        indicators.Compliant(args["filename"] + " owner changed to " + args["owner"] + ":" + args["group"]);
+        indicators.Compliant(filename + " owner changed to " + owner + ":" + groupName);
     }
 
     bool has_permissions = false;
@@ -326,14 +326,15 @@ Result<Status> RemediateEnsureFilePermissionsHelper(const std::string& filename,
 
         std::ostringstream oss;
         oss << std::oct << new_perms;
-        indicators.Compliant(args["filename"] + " permissions changed to " + oss.str());
+        indicators.Compliant(filename + " permissions changed to " + oss.str());
     }
 
     OsConfigLogDebug(log, "File '%s' remediation succeeded", filename.c_str());
     return Status::Compliant;
 }
 
-Result<Status> EnsureFilePermissionsCollectionHelper(std::map<std::string, std::string> args, IndicatorsTree& indicators, ContextInterface& context, bool isRemediation)
+Result<Status> EnsureFilePermissionsCollectionHelper(const std::map<std::string, std::string>& args, IndicatorsTree& indicators,
+    ContextInterface& context, bool isRemediation)
 {
     auto log = context.GetLogHandle();
     auto it = args.find("directory");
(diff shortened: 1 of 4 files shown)


src/modules/compliance/src/lib/procedures/EnsureFilePermissions.cpp
```diff
@@ -17,21 +19,42 @@ namespace
 {
 // Mask to display permissions
 const unsigned short displayMask = 0xFFF;
+static std::string Regexize(const std::string& str)
```
kkanas, line 22: glob_to_regex ?

src/modules/compliance/src/lib/procedures/EnsureFilePermissions.cpp
```diff
@@ -186,7 +209,92 @@ REMEDIATE_FN(EnsureFilePermissions, "filename:Path to the file:M", "owner:Requir
...
+    }
+    auto directory = std::move(it->second);
+    it = args.find("ext");
+    if (it == args.end())
+    {
+        OsConfigLogError(log, "No file pattern provided");
+        return Error("No file pattern provided", EINVAL);
+    }
+    auto ext = std::move(it->second);
+    auto extRegexized = Regexize(ext);
```
jepio, line 234: Do we need to regexize if the patterns are compatible with `glob(3)`?

kkanas (approved):

src/modules/compliance/src/lib/procedures/EnsureFilePermissions.cpp
```diff
@@ -186,7 +209,92 @@ REMEDIATE_FN(EnsureFilePermissions, "filename:Path to the file:M", "owner:Requir
...
+
+    if (dp == NULL)
+    {
+        OsConfigLogInfo(log, "Directory '%s' does not exist", directory.c_str());
+        return indicators.Compliant("Directory '" + directory + "' does not exist");
+    }
+    auto dp_up = std::unique_ptr<DIR, int (*)(DIR*)>(dp, closedir);
+
+    bool hasFiles = false;
+    while ((entry = readdir(dp)) != NULL)
```
jepio, line 258: Does this need to be a recursive walk over all subdirectories? The original implementation(s) use(s) `find` without `maxdepth`. See also `fts(3)`.

src/modules/compliance/src/lib/procedures/EnsureFilePermissions.cpp
jepio: or `fnmatch(3)`

src/modules/compliance/src/lib/procedures/EnsureFilePermissions.cpp
jepio: fyi from copilot:
```
#include <stdio.h>
#include <fts.h>
#include <fnmatch.h>
#include <string.h>

int main(int argc, char **argv) {
    char *paths[] = {"./", NULL};  // Root directory
    FTS *ftsp;
    FTSENT *entry;

    ftsp = fts_open(paths, FTS_PHYSICAL | FTS_NOCHDIR, NULL);
    if (!ftsp) {
        perror("fts_open");
        return 1;
    }

    while ((entry = fts_read(ftsp)) != NULL) {
        if (entry->fts_info == FTS_F) {  // Only regular files
            if (fnmatch("*.conf", entry->fts_name, 0) == 0) {
                printf("Found: %s\n", entry->fts_path);
            }
        }
    }

    fts_close(ftsp);
    return 0;
}
```

src/modules/compliance/src/lib/procedures/EnsureFilePermissions.cpp
kkanas: Maybe unrelated but The oval textfilecontent54_object (aka FileRegexMatch) 
filename + pattern where pattern is regex so maybe it will be more consistent to use regex across API ?

jepio:
@wupeka: Could you please include module tests with this PR?

AhmedBM (approved):

src/modules/compliance/src/lib/procedures/EnsureFilePermissions.cpp
wupeka: Removed completely

src/modules/compliance/src/lib/procedures/EnsureFilePermissions.cpp
wupeka: Here the caller expect it to be a shell-like match.

src/modules/compliance/src/lib/procedures/EnsureFilePermissions.cpp
wupeka: Thanks for the suggestion. I moved the code to use fts + fnmatch.

src/modules/compliance/src/lib/procedures/EnsureFilePermissions.cpp
```diff
@@ -342,4 +332,121 @@ REMEDIATE_FN(EnsureFilePermissions, "filename:Path to the file:M", "owner:Requir
...
+    }
+    auto ext = std::move(it->second);
+
+    auto directorycharp = std::unique_ptr<char, void (*)(void*)>(strdup(directory.c_str()), free);
+    if (nullptr == directorycharp)
+    {
+        OsConfigLogError(log, "Failed to allocate memory for directory path");
+        return Error("Failed to allocate memory for directory path", ENOMEM);
+    }
+    char* const paths[] = {directorycharp.get(), nullptr};
```
jepio, line 360: Could this be replaced by:
```suggestion
    char* const paths[] = {&directory[0], nullptr};
```
?

src/modules/compliance/src/lib/procedures/EnsureFilePermissions.cpp
```diff
@@ -342,4 +332,121 @@ REMEDIATE_FN(EnsureFilePermissions, "filename:Path to the file:M", "owner:Requir
...
+    {
+        if (FTS_F == entry->fts_info)
+        {
+            if (0 == fnmatch(ext.c_str(), entry->fts_name, 0))
+            {
+                hasFiles = true;
+                const char* fileName = entry->fts_path;
+
+                Result<Status> result = isRemediation ? RemediateEnsureFilePermissionsHelper(fileName, args, indicators, context) :
+                                                        AuditEnsureFilePermissionsHelper(fileName, args, indicators, context);
```
jepio, line 382: This doesn't have to be acted on now, but we're copying `args` here for every loop iteration. I'm not sure if we're running this for large directories, but consider passing `args` by `const &` to the helpers? I don't think they necessarily need to modify the args.

jepio (approved):
I wasn't going to comment with these minor suggestions but since I saw the dangling `return` I figure I may as well.

src/modules/compliance/src/lib/procedures/EnsureFilePermissions.cpp
wupeka: I was way too worried about the constness. Fixed.

src/modules/compliance/tests/procedures/EnsureFilePermissionsTest.cpp
```diff
@@ -38,14 +42,19 @@ class EnsureFilePermissionsTest : public ::testing::Test
...
         system("groupadd -g 1 bin >/dev/null 2>&1");
         system("useradd -g 1 -u 1 bin >/dev/null 2>&1");
+        testDir = mkdtemp(dirTemplate);
+        ASSERT_FALSE(testDir.empty());
         indicators.Push("EnsureFilePermissions");
     }
+
     void TearDown() override
     {
+        return;
```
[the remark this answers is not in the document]
wupeka: Yep, debugging artifact.

src/modules/compliance/src/lib/procedures/EnsureFilePermissions.cpp
```diff
@@ -342,4 +332,121 @@ REMEDIATE_FN(EnsureFilePermissions, "filename:Path to the file:M", "owner:Requir
...
+    "mask:Required octal permissions of the file - mask::^[0-7]{3,4}$")
+{
+    return EnsureFilePermissionsCollectionHelper(args, indicators, context, false);
+}
+
+REMEDIATE_FN(EnsureFilePermissionsCollection, "directory:Directory path:M", "ext:File pattern:M", "owner:Required owner of the file",
+    "group:Required group of the file", "permissions:Required octal permissions of the file::^[0-7]{3,4}$",
+    "mask:Required octal permissions of the file - mask::^[0-7]{3,4}$")
+{
+    return EnsureFilePermissionsCollectionHelper(args, indicators, context, true);
```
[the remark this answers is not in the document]
wupeka: Switched to const ref instead.

src/modules/compliance/src/lib/procedures/EnsureFilePermissions.cpp
wupeka: Done.

jepio (approved):

PR #1007, opened 2025-05-12
EnsureMountPointExists audit procedure

wupeka:
## Description
EnsureMountPointExists audit procedure
## Checklist

### DIFF
--- a/src/modules/compliance/src/lib/procedures/EnsureMountPointExists.cpp
+++ b/src/modules/compliance/src/lib/procedures/EnsureMountPointExists.cpp
@@ -1,0 +1,42 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include <CommonUtils.h>
+#include <Evaluator.h>
+#include <Regex.h>
+#include <iostream>
+#include <string>
+
+namespace compliance
+{
+
+AUDIT_FN(EnsureMountPointExists, "mountPoint:Mount point to check:M")
+{
+    auto it = args.find("mountPoint");
+    if (it == args.end())
+    {
+        return Error("No mount point provided");
+    }
+    auto mountPoint = std::move(it->second);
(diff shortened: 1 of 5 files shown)


kkanas:
I think it is ok - just simple module test would give us an answer how well findmnt  is supported across all distributions

src/modules/compliance/src/lib/procedures/EnsureMountPointExists.cpp
```diff
@@ -0,0 +1,42 @@
...
+AUDIT_FN(EnsureMountPointExists, "mountPoint:Mount point to check:M")
+{
+    auto it = args.find("mountPoint");
+    if (it == args.end())
+    {
+        return Error("No mount point provided");
+    }
+    auto mountPoint = std::move(it->second);
+
+    Result<std::string> findMntOutput = context.ExecuteCommand("findmnt -knl");
```
jepio, line 22: `findmnt <mountpoint>` will give you the information for a specific mountpoint directly.

jepio (commented):
Could you please include module tests with this PR?

src/modules/compliance/src/lib/procedures/EnsureMountPointExists.cpp
wupeka: It'll fail when mountpoint is not a mountpoint, doing it this way gives clear distinction between findmnt failure and "not a mountpoint".

jepio: are you saying you expect `findmnt <mountpoint>` to fail for other reasons?

wupeka: The simplest scenario for that would be findmnt missing from the system. We currently return an error in this case

AhmedBM (approved):

PR #1024, opened 2025-05-21
FileRegexMatchTest introduce match sub pattern and behaviors

kkanas:
## Description
Rework and rebase #997 

procedure used to test the state pattern against the full line. In general, the state pattern should match against a captured value with a given instance. **The instance is for now skipped and will be added later based on actual needs, for now it defaults to 1.** <- **WARNING**

## Checklist

### DIFF
--- a/src/modules/compliance/tests/procedures/FileRegexMatchTest.cpp
+++ b/src/modules/compliance/tests/procedures/FileRegexMatchTest.cpp
@@ -59,7 +59,7 @@ TEST_F(FileRegexMatchTest, Audit_InvalidArguments_2)
 {
     mArgs["filename"] = mTempfile;
     auto result = AuditFileRegexMatch(mArgs, mIndicators, mContext);
-    ASSERT_TRUE(!result.HasValue());
+    ASSERT_FALSE(result.HasValue());
     ASSERT_EQ(result.Error().code, EINVAL);
 }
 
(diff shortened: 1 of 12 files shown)


src/modules/compliance/tests/procedures/FileRegexMatchTest.cpp
```diff
@@ -21,237 +22,495 @@ using compliance::Status;
 class FileRegexMatchTest : public ::testing::Test
 {
 protected:
-    char mTempfile[PATH_MAX] = "/tmp/fileRegexMatchTest.XXXXXX";
-    int mTempfileFd = -1;
+    char mTempdir[PATH_MAX] = "/tmp/FileRegexMatchTest.XXXXXX";
```
MariusNi, line 25: >char mTempdir[PATH_MAX] = "/tmp/FileRegexMatchTest.XXXXXX"; [](http://example.com/codeflow?start=4&length=59)

This may not work with all compilers and/or may leave extras uninitialized. Split into two?

wupeka: What extras?

kkanas: @MariusNi  I think you are referring to non static class member initialization
https://en.cppreference.com/w/cpp/language/data_members#Member_initialization
I think this is legal in C++11 which we use. 
Do you have some specific compiler in mind that does not support that ? 
I think we use that idom in another places. Lastly this are unit test and they should be platform independent.

wupeka (approved):

PR #1028, opened 2025-05-27
Rename Compliance module (et consortes) to ComplianceEngine

wupeka:
## Description
Rename Compliance module (et consortes) to ComplianceEngine

## Checklist


kkanas:
Nit fix clang-tidy
https://github.com/Azure/azure-osconfig//blob/wupeka/rename-compliance//src/tests/clang-tidy/run-clang-tidy.py#L19

AhmedBM (approved):

src/adapters/mc/complianceengine/Baseline.c
```diff
@@ -19,21 +19,21 @@ int BaselineIsValidResourceIdRuleId(const char* resourceId, const char* ruleId,
 
 void BaselineInitialize(OsConfigLogHandle log)
 {
-    ComplianceInitialize(log);
-    gCompliance = ComplianceMmiOpen(gComponentName, -1);
+    ComplianceEngineInitialize(log);
+    gComplianceEngine = ComplianceEngineMmiOpen(gComponentName, -1);
```
MariusNi, line 23: >g [](http://example.com/codeflow?start=4&length=1)

Should be "g_" (instead of "g") as per coding style discussion.

src/modules/complianceengine/src/lib/procedures/PackageInstalled.cpp
```diff
@@ -28,7 +28,7 @@ typedef struct
 
 static constexpr long PACKAGELIST_TTL = 3000L;        // just shy of an hours
```
MariusNi, line 29: >just shy of an hours [](http://example.com/codeflow?start=57&length=20)

nit: typo in comment

MariusNi (approved):
:shipit:

wupeka:
> > std::string DetectPackageManager(ContextInterface& context)
> 
> Same Q, and I will stop asking this in this PR. Why isn't this code using the package manager functions from commonutils? And if regex specific APIs are missing, to add them there for everyone to use and still use the common functions here same as elsewhere in OSConfig? That's why we made those libraries reusable.
> 
> Refers to: src/modules/complianceengine/src/lib/procedures/PackageInstalled.cpp:56 in [fcd0744](https://github.com/Azure/azure-osconfig/commit/fcd0744bcc21938f35d0e3c77607fb57e2d10da6). [](commit_id = [fcd0744](https://github.com/Azure/azure-osconfig/commit/fcd0744bcc21938f35d0e3c77607fb57e2d10da6), deletion_comment = False)

1. Re: regex - we're using C++ std::regex, the regexFallback we have is only there because the old versions of libstdc++ from Ubuntu don't support regexes completely. This will go away once we stop supporting old distros, and we'll use std::regex everywhere

2. Re: package manager routines (but that applies to other things your mentioned): our code works differently, from the package managers we use to the detection of package manager to what we consider an installed package to level of caching, and as we talked before - since we must adhere to externally-provided standards it's often easier for us to write the code from scratch (especially such an easy one) than to find some commonalities with the code in CommonUtils. We are also heavily relying gtest and gmock for our unit testing (e.g. see https://github.com/Azure/azure-osconfig/blob/dev/src/modules/compliance/tests/procedures/PackageInstalledTest.cpp ) , and that requires some layer of abstraction.

kkanas (approved):

PR #1035, opened 2025-06-01
FileRegexMatch and RegexFallback fixes

wupeka:
## Description
For FileRegexMatch:
- use simple match vs no match vs error counting that corresponds with OVAL Existence logic.
- get rid of FactExistenceValidator
- match instead of search for the filename

For RegexFallback:
 - newline, cr and tab must be directly given in glibc regex, not escaped
 - regex_match must match the whole string

## Checklist


src/modules/compliance/src/lib/RegexFallback.h
```diff
@@ -303,7 +344,21 @@ inline bool regexSearch(const std::string& s, MatchResults& m, const Regex& r)
 
 inline bool regexMatch(const std::string& s, const Regex& r)
 {
-    return (0 == regexec(r.preg.get(), s.c_str(), 0, NULL, 0));
+    regmatch_t matches[1];
+    auto result = (0 == regexec(r.preg.get(), s.c_str(), 1, matches, 0));
+    if (result)
+    {
+        if (matches[0].rm_so != 0)
```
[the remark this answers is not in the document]
wupeka: For match we have to match the whole string, so the start offset needs to be 0, end offset needs to be length of the string. Yes, -1 means there's no match, and any other positive value means that the match starts somewhere later in the string, but those are all failures.

src/modules/test/recipes/compliance/FileRegexMatch.json
```diff
@@ -36,7 +36,7 @@
     "ObjectType": "Reported",
     "ComponentName": "Compliance",
     "ObjectName": "auditTest",
-    "Payload": "PASS{ FileRegexMatch: All facts exist } == TRUE"
+    "Payload": "PASS{ FileRegexMatch: All 1 files matched the pattern } == TRUE"
```
MariusNi, line 39: >files [](http://example.com/codeflow?start=44&length=5)

nit: when there is only one file, can we say "file" instead of "files"?

kkanas (approved):

src/modules/test/recipes/compliance/FileRegexMatch.json
wupeka: While poor grammar, having consistency here helps with automated testing.

kkanas (approved):

PR #1053, opened 2025-06-17
Add optional awk transform to ExecuteCommandGrep

wupeka:
## Description
Add optional awk transform to ExecuteCommandGrep
## Checklist


jepio:
What is the goal here and where do the `awk` scripts come from/how are they authored? (this context is needed in the PR description)
Initial thoughts are:
- what can the awk filter do that the grep can't?
- we intentionally whitelist the allowed commands that can be executed, but awk would allow to bypass that - this makes it feel like a bad idea

wupeka:
@jepio exec | awk | grep is a pattern that's used in a few SCE scripts.
re: security, I added -S (run in sandbox) option to awk, which stops scripts from doing nasty stuff like executing things

robertwoj-microsoft:
> @jepio exec | awk | grep is a pattern that's used in a few SCE scripts. re: security, I added -S (run in sandbox) option to awk, which stops scripts from doing nasty stuff like executing things

Maybe add a simple test which tries to execute sth but fails due to --sandbox flag?

jepio:
> > @jepio exec | awk | grep is a pattern that's used in a few SCE scripts.
> 
> Which ones? Can you expand the PR description
> 

I would really like to understand the real-world application of this - based on the tests I can't follow why we need this.

Also: are we guaranteed to have awk installed and are we sure it's going to be gawk (as far as I can tell the only one that support `-S`).

PR #1054, opened 2025-06-17
Add a prefix to a non-critical incompliance message

robertwoj-microsoft:
## Description

We have a graceful way of handling non-critical incompliance messages in the `ComplianceEngineMmiGet` function. This functionality prevents from prematurely aborting audit runs and allows GC worker to continue evaluation if a single rule fails.

The Augmentation Engine parses the outputs reported in the reason field and used to match against a 'failed to' string. However, this is incorrect as the messages vary and rely strictly on the procedures' outputs.

This PR adds a static prefix that we can attach to instead of the 'failed to' which is too generic and can be used in compliant cases.

## Checklist

### DIFF
--- a/src/modules/complianceengine/src/lib/ComplianceEngineInterface.cpp
+++ b/src/modules/complianceengine/src/lib/ComplianceEngineInterface.cpp
@@ -114,8 +114,7 @@ int ComplianceEngineMmiGet(MMI_HANDLE clientSession, const char* componentName, 
             {
                 OsConfigLogError(engine.Log(), "ComplianceEngineMmiGet failed with a non-critical error: %s (errno: %d)",
                     result.Error().message.c_str(), result.Error().code);
-                result =
-                    ComplianceEngine::AuditResult(Status::NonCompliant, "ComplianceEngineMmiGet failed with a non-critical error: " + result.Error().message);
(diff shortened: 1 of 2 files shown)


src/modules/complianceengine/src/lib/ComplianceEngineInterface.cpp
```diff
@@ -114,7 +114,8 @@ int ComplianceEngineMmiGet(MMI_HANDLE clientSession, const char* componentName,
             {
                 OsConfigLogError(engine.Log(), "ComplianceEngineMmiGet failed with a non-critical error: %s (errno: %d)",
                     result.Error().message.c_str(), result.Error().code);
-                result = ComplianceEngine::AuditResult(Status::NonCompliant, result.Error().message);
+                result =
+                    ComplianceEngine::AuditResult(Status::NonCompliant, "ComplianceEngineMmiGet failed with a non-critical error: " + result.Error().message);
```
MariusNi, line 118: nit: this looks a bit strange, don't you think? Breaking the line this way. Make lines 117 and 118 one line?

robertwoj-microsoft: Changed the prefix, should be better now.

kkanas (approved):

PR #1106, opened 2025-08-06
Add UFW status detection

robertwoj-microsoft:
## Description

Avoid failing with errors when the service is inactive. We fail some rules in case the UFW is inactive, the output of `ufw status verbose` command becomes:
```
Status: inactive
```
This PR adds handling for this situation and the following minor changes:
- downgrades some log messages from errors which were most likely missed during code review

## Checklist

### DIFF
--- a/src/modules/complianceengine/src/lib/procedures/EnsureFirewallOpenPorts.cpp
+++ b/src/modules/complianceengine/src/lib/procedures/EnsureFirewallOpenPorts.cpp
@@ -10,6 +10,14 @@
 namespace ComplianceEngine
 {
 
+namespace
+{
+enum class UfwStatus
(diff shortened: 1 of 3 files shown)


src/modules/complianceengine/src/lib/procedures/EnsureFirewallOpenPorts.cpp
```diff
@@ -10,6 +10,14 @@
 namespace ComplianceEngine
 {
 
+namespace
+{
+enum class UfwStatus
```
wupeka, line 15: Could we have a simple bool ufwActive instead?

wupeka (approved):

PR #1119, opened 2025-09-02
ASB fix for AuditEnsurePasswordCreationRequirements rule

MariusNi:
## Description

ASB fix for AuditEnsurePasswordCreationRequirements rule. The audit of this rule was missing corner cases when /etc/pam.d/common-password exists and fails to audit, to cause whole audit to fail, ignoring possible remediation done by user in  /etc/security/pwquality.conf. 

With this the audit approach is slightly changed like the following, see comment added to CheckPasswordCreationRequirements in src\common\commonutils\PassUtils.c:

```
    // First, check if at least one of the following files exists :
    // - /etc/pam.d/common-password
    // - /etc/security/pwquality.conf
    // If neither exists, the audit fails early.
    //
    // If /etc/pam.d/common-password exists:
    // - Audit inline parameters (e.g., minlen, dcredit, ucredit, etc.).
    // - If all required parameters are present and valid, the audit passes.
    //
    // If the audit fails for common-password and /etc/security/pwquality.conf is present :
    // - Audit pwquality.conf for the required parameters.
    // - If valid, the audit passes.
    //
    // This logic ensures compatibility across distros and PAM module configurations and supports both common-password and pwquality.conf setups.
```

## Checklist

### DIFF
--- a/src/common/commonutils/PassUtils.c
+++ b/src/common/commonutils/PassUtils.c
@@ -713,20 +713,41 @@ static int CheckRequirementsForPwQualityConf(int retry, int minlen, int minclass
 
 int CheckPasswordCreationRequirements(int retry, int minlen, int minclass, int dcredit, int ucredit, int ocredit, int lcredit, char** reason, OsConfigLogHandle log)
 {
+    // First, check if at least one of the following files exists :
+    // - /etc/pam.d/common-password
+    // - /etc/security/pwquality.conf
+    // If neither exists, the audit fails early.
+    //
+    // If /etc/pam.d/common-password exists:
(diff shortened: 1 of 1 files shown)


robertschaedler3 (approved):

AhmedBM (approved):
discussed offline, difficult to test with current infrastructure

MariusNi:
> discussed offline, difficult to test with current infrastructure
We have unittests for the invoked functions, and we have the test recipe tests for each individual distro. This special corner case that this fix addresses is as you mentioned very difficult to setup. we'll need to mock or setup different PAM installations on each VM, etc.

PR #1121, opened 2025-09-04
Refactor arguments parsing

robertwoj-microsoft:
## Description

Introduce automatic binding procedure arguments to native structures. The idea is to reduce the amount of manual extraction of arguments from the args array and parsing them based on the type.

Rationale:
We have the following code almost everywhere:
```
AUDIT_FN(EnsureSysctl, "sysctlName:Name of the sysctl:M:^([a-zA-Z0-9_]+[\\.a-zA-Z0-9_-]+)$", "value:Regex that the value of sysctl has to match:M")
```
```
    auto it = args.find("sysctlName");
    if (it == args.end())
    {
        return Error("Missing 'sysctlName' parameter", EINVAL);
    }
    auto sysctlName = std::move(it->second);

    it = args.find("value");
    if (it == args.end())
    {
        return Error("Missing 'value' parameter", EINVAL);
    }
    auto sysctlValue = std::move(it->second);

```

The goal of this proposal is to have the following:
```
struct AuditEnsureSysctlParams
{
    /// description:Name of the sysctl
    /// pattern:^([a-zA-Z0-9_]+[\\.a-zA-Z0-9_-]+)$
    std::string sysctlName;

    /// description:Regex that the value of sysctl has to match
    std::string sysctlValue;

    /// description:Some optional regular expression
    Optional<regex> someOptionalRegex;
};
```

And:
```
AUDIT_FN(EnsureSysctl)
```
or, directly without a macro:
```
Result<Status> AuditEnsureSysctl(const AuditEnsureSysctlParams& params, IndicatorsTree& indicators,   ContextInterface& context)
```

Inside the implementation, instead of `args` we'd have `params`/`parameters` argument which is a constref to the `AuditEnsureSysctlParams` structure.

## Checklist


src/modules/complianceengine/src/lib/Pattern.h
```diff
@@ -0,0 +1,89 @@
...
+    Pattern(const std::string& pattern) noexcept(false);
+
+public:
+    Pattern() = default;
+    Pattern(const Pattern&) = default;
+    Pattern(Pattern&&) = default;
+    Pattern& operator=(const Pattern& other)
+    {
+        if (&other == this)
+            return *this;
```
MariusNi, line 32: nit (coding style): wrap one line block in curly braces

{
    //line
}

Here and elsewhere.

And super minor, don't worry to skip.

robertwoj-microsoft:
Closing as #1134 continues the work.

PR #1127, opened 2025-09-08
Added 1DS Telemetry implementation

AhmedBM:
## Description

This change includes integration with [Microsoft 1DS C++ SDK](https://github.com/microsoft/cpp_client_telemetry).

The integration requires at least GCC 5.5+ meaning the Machine Configuration packages which require GCC 4.8 will use a different toolchain.

* nrp-build - add ingestion key for telemetry flow (Only including NonProd for now)
* Updated Build-Containers workflow to use root context in order to include vcpkg.json from source
* Added GCC 5.5 to Ubuntu 14 (+others) - Needed for 1DS C++ SDK
* Added External_Project targets for 1DS C++ SDK in order to compile with different toolchain (gcc 5.5)
* Added VCPKG custom triplet and toolchain for gcc5
* Added telemetry binary and plumbing to fork external binary
* Added CMake plumbing for running external target tests for CTest integration

Requires #1129

## Checklist


src/CMakeLists.txt
```diff
@@ -20,17 +17,25 @@ if (NOT DEFINED CMAKE_TOOLCHAIN_FILE)
     endif()
 endif()
 
-cmake_minimum_required(VERSION 3.3)
+cmake_minimum_required(VERSION 3.16)
```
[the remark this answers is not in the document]
AhmedBM: ill also update the README, we require CMake 3.16 because of some needed `find_package / NO_DEFAULT_PATH` features and `ExternalProject_Add / GIT_SUBMODULES` functionality

src/adapters/mc/CMakeLists.txt
```diff
@@ -1,6 +1,37 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 
+if (ENABLE_TELEMETRY)
+    include(ExternalProject)
+    if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS "5.5")
+        find_program(GCC5_FOUND gcc-5)
+        if(NOT GCC5_FOUND)
```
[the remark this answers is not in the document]
AhmedBM: this part is for building the external binary, the one which requires a different toolchain. It uses `ExternalProject_Add` to use a different toolchain, in this case gcc5 when it detects that the default compiler is <5.5

src/common/asb/CMakeLists.txt
```diff
@@ -12,7 +12,13 @@ target_include_directories(asb PUBLIC
...
+target_link_libraries(asb
+    PUBLIC
+        $<$<BOOL:${ENABLE_TELEMETRY}>:telemetryjsonlib>
+        $<$<NOT:$<BOOL:${ENABLE_TELEMETRY}>>:telemetrystublib>
+    PRIVATE
+        dl
+        logging
+        commonutils
+        parsonlib
+)
```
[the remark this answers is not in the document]
AhmedBM: could you elaborate, not sure what you mean here?

src/common/telemetry/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+cmake_minimum_required(VERSION 3.16)
```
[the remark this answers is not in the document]
AhmedBM: This project has to be buildable independently without needing the rest of the project, this is necessary as `ExternalProject_Add` will spawn CMake from this level so i need to redefine some defaults just to make CMake happy

src/common/telemetry/exe/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+project(telemetrybin)
+set (CMAKE_C_STANDARD 11)
```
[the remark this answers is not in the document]
AhmedBM: its unecessary as lib holds the `gcc5` dependency and lib depends on this target so the check will be done transitively

src/common/telemetry/exe/main.cpp
```diff
@@ -0,0 +1,146 @@
...
+bool file_exists(const std::string& filepath) {
+    std::ifstream file(filepath.c_str());
+    return file.good();
+}
+
+void print_usage(const char* program_name) {
+    std::cout << "Usage: " << program_name << " [OPTIONS] <json_file_path> [teardown_time_seconds] [syslog_identifier]" << std::endl;
+    std::cout << "  json_file_path         - Path to the JSON file to process" << std::endl;
+    std::cout << "  teardown_time_seconds  - Optional: Teardown time in seconds (default: 5s)" << std::endl;
+    std::cout << "  syslog_identifier      - Optional: Custom identifier for syslog entries (default: telemetry-exe)" << std::endl;
```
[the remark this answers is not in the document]
AhmedBM: will do this in another PR

src/common/telemetry/lib/CMakeLists.txt
```diff
@@ -0,0 +1,108 @@
...
+if (DEFINED ENV{VCPKG_ROOT})
+    set(VCPKG_INSTALLED_DIR "$ENV{VCPKG_ROOT}/vcpkg_installed")
+else()
+    set(VCPKG_INSTALLED_DIR "${CMAKE_CURRENT_SOURCE_DIR}/../../../external/vcpkg/vcpkg_installed")
+endif()
+set(VCPKG_PACKAGES_DIR "${VCPKG_INSTALLED_DIR}/${VCPKG_TARGET_TRIPLET}")
+
+# Set CMAKE_PREFIX_PATH to vcpkg only
+set(CMAKE_PREFIX_PATH "${VCPKG_PACKAGES_DIR}")
+
```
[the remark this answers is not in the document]
AhmedBM: no sorry for the confusion, this `lib` is only used by the external binary target (`exe`), the gcc4 lib is `libjson`

.github/workflows/nrp-build.yml
```diff
@@ -19,3 +19,5 @@ jobs:
       artifact: policy-packages
       machine-config: true
       release: true
+    secrets:
+      onedstenant: ${{ github.ref == 'refs/heads/main' && secrets.ONEDS_TENANT_PROD || github.ref == 'refs/heads/dev' && secrets.ONEDS_TENANT_NONPROD || '' }}
```
MariusNi, line 23: >onedstenant: ${{ github.ref == 'refs/heads/main' && secrets.ONEDS_TENANT_PROD || github.ref == 'refs/heads/dev' && secrets.ONEDS_TENANT_NONPROD || '' }} [](http://example.com/codeflow?start=6&length=152)

nit: can we break this block on multiple lines?

src/adapters/mc/asb/CMakeLists.txt
```diff
@@ -24,16 +24,25 @@ set_target_properties(OsConfigResourceAsb PROPERTIES OUTPUT_NAME libOsConfigReso
 
 # Create the policy artifacts ZIP packages for the Azure Security Baseline for Linux (Get and Set)
 
+# Add telemetry binary if enabled
+if(BUILD_TELEMETRY)
+    list(APPEND STAGING_COMMANDS_GET
+        COMMAND ${CMAKE_COMMAND} -E copy_if_different "${BINARY_DIR}/exe/telemetry" ${PROJECT_BINARY_DIR}/StagingAzureLinuxBaseline/Modules/DscNativeResources/OsConfigResource/OSConfigTelemetry
+    )
```
MariusNi, line 31: >) [](http://example.com/codeflow?start=4&length=1)

Please move at the end of line 30. Same for line 34 below. Thanks!

src/common/telemetry/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
...
+project(telemetry)
+
+if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS "5.5")
+    message(FATAL_ERROR "GCC version must be 5.5 or greater to enable telemetry")
+endif()
+
```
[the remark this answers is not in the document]
AhmedBM: It should break as it will break the build. Its much more dev/user friendly to just fail the build and say why it failed.

This will not happen in our environments and in fact, it was useful at debugging some build failures i saw on some distros.

src/common/telemetry/exe/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
...
+set (CMAKE_CXX_STANDARD 11)
+
+add_executable(telemetry
+    main.cpp
+)
+
+target_link_libraries(telemetry
+    PRIVATE
+        telemetrylib
```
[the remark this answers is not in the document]
AhmedBM: no the TelemetryJsonLib is used by the MC targets (gcc 4.8), this is our external binary target which links against the 1DS telemetry libs (`telemetrylib`) which require gcc 5.5+

src/common/telemetry/exe/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
...
+)
```
MariusNi, line 15: Move this at end of preceding line, and also do same everywhere else necessary

src/common/telemetry/lib/Telemetry.hpp
```diff
@@ -0,0 +1,71 @@
...
+
+#include <Keys.h>
+#include <LogManager.hpp>
+#include <Logging.h>
+
+#include "ParameterSets.hpp"
+
+namespace Telemetry {
+
+    class TelemetryManager {
```
MariusNi, line 23: >{ [](http://example.com/codeflow?start=27&length=1)

Break { on next line, here and everywhere else same occurrs

.github/workflows/universalnrp-test.yml
```diff
@@ -190,7 +190,8 @@ jobs:
         uses: ./.github/actions/check-size
         with:
           package: '${{ steps.normalize.outputs.path }}'
-          limit: 350
+          # TODO: Adjust size limit as needed
```
MariusNi, line 193: ># TODO: Adjust size limit as needed [](http://example.com/codeflow?start=10&length=35)

How would we know when this limit would be in need to be increased.... or is too large and we need it decreased?

And speaking of this, what is this limit for and what are the units in? Can we have this described in a comment instead of the bare TODO one?

devops/docker/rhel-7-amd64/Dockerfile
```diff
@@ -1,11 +1,13 @@
 # ubi7 images are missing RPM devtools, copy from centos7
 FROM centos:centos7 as centos
 RUN rm /etc/yum.repos.d/*.repo
-COPY centos.repo /etc/yum.repos.d/centos.repo
+# Docker context is at the root of the project
```
MariusNi, line 4: Why do we need to repeat this same comment as many times? As this comment really does not seem to bring anything useful to the reader, can we simply remove it from all places it is appearing?

src/CMakeLists.txt
```diff
@@ -151,6 +171,21 @@ if (BUILD_MODULES)
     add_subdirectory(modules)
 endif()
 
+# Include External projects with tests
+if (BUILD_TELEMETRY)
```
MariusNi, line 175: Remove this comment as well -- it does not stat anything useful, is clear that the telemetry project gets built.

src/CMakeLists.txt
```diff
@@ -151,6 +171,21 @@ if (BUILD_MODULES)
...
+    ExternalProject_Get_Property(telemetry BINARY_DIR)
+    set(TELEMETRY_TESTS_DIR ${BINARY_DIR}/lib/tests)
+    configure_file(
+        ${CMAKE_SOURCE_DIR}/cmake/ExternalTests.cmake.in
+        ${CMAKE_BINARY_DIR}/ExternalTests.cmake
+        @ONLY
+    )
```
MariusNi, line 182: >) [](http://example.com/codeflow?start=4&length=1)

This ) here and the one on line 186 are still in the opposite style of the project's style.

Do:

foo)

Instead:

foo
)

src/adapters/mc/CMakeLists.txt
```diff
@@ -1,6 +1,38 @@
...
 
+if (BUILD_TELEMETRY)
+    include(ExternalProject)
+    if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS "5.5")
+        find_program(GCC5_FOUND gcc-5)
+        if(NOT GCC5_FOUND)
+            message(FATAL_ERROR "gcc-5 not found in PATH, but required for 1DS SDK with GCC < 5.5")
+        endif()
+
+        find_program(GPP5_FOUND g++-5)
```
MariusNi, line 12: nit: remove blank line 11

src/common/telemetry/lib/Telemetry.hpp
```diff
@@ -0,0 +1,73 @@
...
+
+        // Delete copy constructor and assignment operator to enforce singleton
+        TelemetryManager(const TelemetryManager&) = delete;
+        TelemetryManager& operator=(const TelemetryManager&) = delete;
+
+        // Initialize telemetry system
+        bool Initialize(bool enableDebug = false, int teardownTime = CONFIG_DEFAULT_TEARDOWN_TIME);
+
+        // Check if telemetry is initialized
+        bool IsInitialized() const;
```
MariusNi, line 40: > bool IsInitialized() const; [](http://example.com/codeflow?start=7&length=28)

This one can be a private function and be called by ProcessJsonFile. 

The client does not need to see GetInstance, Initialize, and Shutdown. All these can be automatically called internally by ProcessJsonFile.

src/common/telemetry/lib/Telemetry.cpp
```diff
@@ -0,0 +1,369 @@
...
+
+bool TelemetryManager::IsInitialized() const
+{
+    std::lock_guard<std::mutex> lock(m_mutex);
+    return m_initialized;
+}
+
+void TelemetryManager::EventWrite(Microsoft::Applications::Events::EventProperties event)
+{
+    if (!m_initialized || !m_logger)
```
MariusNi, line 93: > if (!m_initialized || !m_logger) [](http://example.com/codeflow?start=3&length=33)

Add inner pharens

src/common/telemetry/lib/Telemetry.cpp
```diff
@@ -0,0 +1,369 @@
...
+        LogManager::SetTransmitProfile(TransmitProfile::TransmitProfile_RealTime);
+
+        m_initialized = true;
+        return true;
+    }
+    catch (const std::exception& e)
+    {
+        if (m_logfileHandle)
+        {
+            OsConfigLogError(m_logfileHandle, "Failed to initialize telemetry: %s", e.what());
```
MariusNi, line 79: >m_logfileHandle [](http://example.com/codeflow?start=29&length=15)

This is not a file handle. May want to change name to something like 'm_log', is all that we need to know.

src/common/telemetry/lib/Telemetry.cpp
```diff
@@ -0,0 +1,369 @@
...
+        {
+            OsConfigLogError(m_logfileHandle, "Unexpected parameter '%s' for event '%s'", jsonKey.c_str(), eventName.c_str());
+            return false;
+        }
+    }
+
+    return true;
+}
+
+void TelemetryManager::ProcessJsonLine(const std::string& jsonLine)
```
MariusNi, line 226: >TelemetryManager::ProcessJsonLine [](http://example.com/codeflow?start=5&length=33)

Why is this not validating that the TelemetryManager is initialized?

Please make those members private and invoke them from inside this function, at beginning and at the end,

.github/workflows/universalnrp-test.yml
AhmedBM: I will remove the TODO but ive already checked the release package size and the largest, being the complianceengine is ~4800KB so ive left 5000KB as the ceiling.

src/vcpkg.json
```diff
@@ -1,8 +1,15 @@
 {
+  "name": "azure-osconfig",
+  "version": "1.0.0",
   "dependencies": [
```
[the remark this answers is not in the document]
AhmedBM: Can you elaborate, not sure what your suggesting here?

src/common/telemetry/lib/tests/TelemetryTests.cpp
```diff
@@ -0,0 +1,103 @@
...
+// Licensed under the MIT License.
+
+#include <chrono>
+#include <fstream>
+#include <gtest/gtest.h>
+#include <thread>
+
+#include <Telemetry.hpp>
+
+class TelemetryTest : public ::testing::Test
```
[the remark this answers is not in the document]
AhmedBM: i have a few functions doing so, they have the pattern `Telemetry*Json*`

src/common/telemetry/exe/main.cpp
```diff
@@ -0,0 +1,142 @@
...
+        }
+        catch (const std::exception& e)
+        {
+            OsConfigLogError(g_log, "Error: Invalid teardown time argument '%s'. Must be a valid integer.", argv[arg_index]);
+            return _exit(g_log, 1);
+        }
+    }
+
+    // Initialize telemetry
+    auto& telemetry = Telemetry::TelemetryManager::GetInstance();
```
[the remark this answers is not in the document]
AhmedBM: yes it is

MariusNi (approved):
:shipit:

src/common/telemetry/lib/Telemetry.cpp
```diff
@@ -0,0 +1,363 @@
...
+}
+
+bool TelemetryManager::Initialize(bool enableDebug, int teardownTime)
+{
+    if (m_initialized)
+    {
+        return true;
+    }
+
+        OsConfigLogInfo(m_log, "Initializing telemetry...");
```
MariusNi, line 49: >    OsConfigLogInfo(m_log, "Initializing telemetry..."); [](http://example.com/codeflow?start=4&length=56)

(nt) nit of misalignment here on line 49

src/common/telemetry/lib/Telemetry.cpp
```diff
@@ -0,0 +1,363 @@
...
+        return false;
+    }
+}
+
+void TelemetryManager::EventWrite(Microsoft::Applications::Events::EventProperties event)
+{
+    if ((!m_initialized) || (!m_logger))
+    {
+        return;
+    }
```
MariusNi, line 85: This would be an error -- can be log?

src/common/telemetry/exe/main.cpp
```diff
@@ -0,0 +1,189 @@
...
+        : m_log(OpenLog(logPath, rollingFile))
+    {
+    }
+
+    ~LogHandle()
+    {
+        if (m_log != nullptr)
+        {
+            CloseLog(&m_log);
+            m_log = nullptr;
```
adteran, line 29: line 29 can be removed.

src/common/telemetry/exe/main.cpp
```diff
@@ -0,0 +1,189 @@
...
+#include <Telemetry.hpp>
+
+class LogHandle
+{
+private:
+    OsConfigLogHandle m_log;
+
+public:
+    explicit LogHandle(const char* logPath, const char* rollingFile = nullptr)
+        : m_log(OpenLog(logPath, rollingFile))
```
adteran, line 20: Can OpenLog fail?

src/common/telemetry/exe/main.cpp
```diff
@@ -0,0 +1,189 @@
...
+        }
+        else
+        {
+            OsConfigLogError(log.get(), "Error: Failed to initialize telemetry.");
+            return 1;
+        }
+
+        return 0;
+    }
+    catch (...)
```
adteran, line 185: This will catch and hide exceptions, how will you diagnose exceptions caught here in the wild?

src/common/telemetry/exe/main.cpp
```diff
@@ -0,0 +1,189 @@
...
+        }
+
+        bool verbose = false;
+        std::string filepath;
+        int teardown_time = -1; // Use -1 to indicate default should be used
+        int arg_index = 1;
+
+        // Parse verbose flag
+        try
+        {
```
adteran, line 70: Readability, avoid long functions like this. Move these to helper functions to parse the parameters

src/common/telemetry/exe/main.cpp
```diff
@@ -0,0 +1,189 @@
...
+            if (argc > 1 && (std::string(argv[1]) == "-v" || std::string(argv[1]) == "--verbose"))
```
adteran, line 71: So you can only enable verbose as the first param? Forcing the arguments to a specific position seems as an arbitrary limitation. 
 This is re-inventing the wheel, is there a lib you can use here to parse arguments in a more robust way?

Also was this part of the signed off design? Who will run the binary with -v?

src/common/telemetry/exe/main.cpp
```diff
@@ -0,0 +1,189 @@
...
+            OsConfigLogInfo(log.get(), "%s", init_message.c_str());
+        }
+        catch (const std::exception& e)
+        {
+            OsConfigLogError(log.get(), "Error: Failed to create initialization message: %s", e.what());
+            return 1;
+        }
+
+        bool init_success = true;
+
```
adteran, line 147: I don't follow, are the var and if needed (146, 148)? you return early on failure.

src/common/telemetry/exe/main.cpp
```diff
@@ -0,0 +1,189 @@
...
+        {
+            try
+            {
+                OsConfigLogInfo(log.get(), "Telemetry initialized successfully!");
+
+                Telemetry::TelemetryManager::SetupConfiguration(verbose, teardown_time);
+                Telemetry::TelemetryManager::ProcessJsonFile(filepath);
+                OsConfigLogInfo(log.get(), "Processed telemetry JSON file: %s", filepath.c_str());
+
+                auto start_time = std::chrono::high_resolution_clock::now();
```
adteran, line 158: I don't understand why we are measuring shutdown instead of the json processing in lines 154 to 156...

src/common/telemetry/exe/main.cpp
```diff
@@ -0,0 +1,189 @@
...
+        bool init_success = true;
+
+        if (init_success)
+        {
+            try
+            {
+                OsConfigLogInfo(log.get(), "Telemetry initialized successfully!");
+
+                Telemetry::TelemetryManager::SetupConfiguration(verbose, teardown_time);
+                Telemetry::TelemetryManager::ProcessJsonFile(filepath);
```
adteran, line 155: where is this implemented?

src/common/telemetry/lib/tests/TelemetryTests.cpp
```diff
@@ -0,0 +1,81 @@
...
+{
+    CreateTestJsonFile("");
+    EXPECT_TRUE(Telemetry::TelemetryManager::ProcessJsonFile(m_testJsonFile)); // Empty file should be processed successfully
+}
+
+// Test processing a file with valid JSON lines
+TEST_F(TelemetryTest, ProcessValidJsonFile)
+{
+    // Create a test file with a valid event
+    std::string validJson = R"({"EventName": "TestEvent", "TestParam": "value"})";
```
adteran, line 51: consider adding samples of real events.

src/common/telemetry/lib/Telemetry.cpp
```diff
@@ -0,0 +1,365 @@
...
+
+void TelemetryManager::SetupConfiguration(bool enableDebug, int teardownTime)
+{
+    auto& logConfig = LogManager::GetLogConfiguration();
+
+    logConfig[CFG_BOOL_ENABLE_TRACE] = enableDebug;
+    logConfig[CFG_INT_TRACE_LEVEL_MIN] = 0;
+    logConfig[CFG_INT_MAX_TEARDOWN_TIME] = teardownTime;
+
+    m_logger = LogManager::Initialize(API_KEY);
```
adteran, line 39: move to the initlist in the constructor?

src/common/telemetry/lib/Telemetry.cpp
```diff
@@ -0,0 +1,365 @@
...
+        return true;
+    }
+
+    OsConfigLogInfo(m_log, "Initializing telemetry...");
+
+    try
+    {
+        if (NULL == m_log)
+        {
+            m_log = OpenLog("/var/log/osconfig_telemetry.log", NULL);
```
adteran, line 55: can OpenLog fail? what happens if it does?

src/common/telemetry/lib/Telemetry.cpp
```diff
@@ -0,0 +1,365 @@
...
+
+    m_logger = LogManager::Initialize(API_KEY);
+}
+
+bool TelemetryManager::Initialize(bool enableDebug, int teardownTime)
+{
+    if (m_initialized)
+    {
+        return true;
+    }
```
adteran, line 47: make this private and you probably want to init once in the constructor. 
avoid checks like this.

src/common/telemetry/lib/Telemetry.cpp
```diff
@@ -0,0 +1,365 @@
...
+    try
+    {
+        if (NULL == m_log)
+        {
+            m_log = OpenLog("/var/log/osconfig_telemetry.log", NULL);
+        }
+
+        SetupConfiguration(enableDebug, teardownTime);
+
+        if (!m_logger)
```
adteran, line 60: Brittle, assumes external ordering of calls. 
Also, consider comment in line 47

src/common/telemetry/lib/Telemetry.cpp
```diff
@@ -0,0 +1,365 @@
...
+        {
+            return false;
+        }
+
+        LogManager::SetTransmitProfile(TransmitProfile::TransmitProfile_RealTime);
```
adteran, line 65: can this fail?

src/common/telemetry/lib/Telemetry.cpp
```diff
@@ -0,0 +1,365 @@
...
+        m_initialized = true;
+        OsConfigLogInfo(m_log, "Telemetry initialized successfully.");
+        return true;
+    }
+    catch (const std::exception& e)
+    {
+        if (m_log)
```
adteran, line 72: this seems wrong to me as you can throw and leave the members vars half set. Avoid this pattern either you construct and initialize of not.

src/common/telemetry/lib/Telemetry.cpp
```diff
@@ -0,0 +1,365 @@
...
+        {
+            OsConfigLogError(m_log, "Failed to initialize telemetry: %s", e.what());
+        }
+        return false;
+    }
+}
+
+void TelemetryManager::EventWrite(Microsoft::Applications::Events::EventProperties event)
+{
+    if ((!m_initialized) || (!m_logger))
```
adteran, line 82: we should always be init if we were able to construct. This is a silent fail on error condition... The caller will not know that something is wrong with initialization.

src/common/telemetry/lib/Telemetry.cpp
```diff
@@ -0,0 +1,365 @@
...
+    {
+        OsConfigLogError(m_log, "Telemetry not initialized");
+        return;
+    }
+
+    m_logger->LogEvent(event);
```
adteran, line 88: can this call fail?

src/common/telemetry/lib/Telemetry.cpp
```diff
@@ -0,0 +1,365 @@
...
+
+}
+
+void TelemetryManager::Shutdown()
+{
+    if (!m_initialized)
```
adteran, line 94: Call shutdown from the destructor, these checks can potentially be removed.

src/common/telemetry/lib/Telemetry.cpp
```diff
@@ -0,0 +1,365 @@
...
+    {
+        OsConfigLogError(m_log, "Telemetry not initialized");
+        return;
+    }
+
+    OsConfigLogInfo(m_log, "Shutting down telemetry...");
+
+    try
```
adteran, line 102: Seems like this should be refactored to the upload method.

src/common/telemetry/lib/Telemetry.cpp
```diff
@@ -0,0 +1,365 @@
...
+        OsConfigLogInfo(m_log, "Telemetry shutdown complete.");
+        CloseLog(&m_log);
+    }
+    catch (const std::exception& e)
+    {
+        if (m_log)
+        {
+            OsConfigLogError(m_log, "Exception during shutdown: %s", e.what());
+        }
+    }
```
adteran, line 117: avoid low level try / catch, you may want to have this wrapper at the main level and have the internal functions throw exceptional errors if encountered.

src/common/telemetry/lib/Telemetry.cpp
```diff
@@ -0,0 +1,365 @@
...
+        if (!Initialize())
+        {
+            OsConfigLogError(m_log, "Failed to initialize TelemetryManager");
+            return false;
+        }
+    }
+
+    OsConfigLogInfo(m_log, "Processing JSON file: %s", filePath.c_str());
+
+    FILE* file = fopen(filePath.c_str(), "r");
```
adteran, line 133: use raii, either a class, smart pointer construct or an unwinder this may leak on throw.

src/common/telemetry/lib/Telemetry.cpp
```diff
@@ -0,0 +1,365 @@
...
+    if (!file)
+    {
+        OsConfigLogError(m_log, "Failed to open file: %s", filePath.c_str());
+        return false;
+    }
+
+    char* line = nullptr;
```
adteran, line 140: same as 133

src/common/telemetry/lib/Telemetry.cpp
```diff
@@ -0,0 +1,365 @@
...
+    {
+        OsConfigLogError(m_log, "Error processing JSON file '%s': %s", filePath.c_str(), e.what());
+        success = false;
+    }
+
+    OsConfigLogInfo(m_log, "Completed processing JSON file: %s", filePath.c_str());
+
+    // Cleanup
+    if (line)
+    {
```
adteran, line 176: remove use raii.

src/common/telemetry/lib/Telemetry.cpp
```diff
@@ -0,0 +1,365 @@
...
+            return false;
+        }
+    }
+
+    return true;
+}
+
+void TelemetryManager::ProcessJsonLine(const std::string& jsonLine)
+{
+    if (!m_logger)
```
adteran, line 224: see other comments on removing these checks.

src/common/telemetry/lib/Telemetry.cpp
```diff
@@ -0,0 +1,365 @@
...
+    {
+        return;
+    }
+
+    JSON_Value* jsonValue = nullptr;
```
adteran, line 229: define on first use. no need to set to null and then assign below.

src/common/telemetry/lib/Telemetry.cpp
```diff
@@ -0,0 +1,365 @@
...
+    JSON_Object* jsonObject = nullptr;
+
+    try
+    {
+        // Parse the JSON string
+        jsonValue = json_parse_string(jsonLine.c_str());
+        if (!jsonValue)
```
adteran, line 236: compare to nullptr not !

src/common/telemetry/lib/Telemetry.cpp
```diff
@@ -0,0 +1,365 @@
...
+        {
+            OsConfigLogError(m_log, "Failed to parse JSON line: %s", jsonLine.c_str());
+            return;
+        }
+
+        jsonObject = json_value_get_object(jsonValue);
+        if (!jsonObject)
+        {
+            OsConfigLogError(m_log, "JSON line is not an object: %s", jsonLine.c_str());
+            json_value_free(jsonValue);
```
adteran, line 246: use raii, create an object to track and release when exiting scope.

src/common/telemetry/lib/Telemetry.cpp
```diff
@@ -0,0 +1,365 @@
...
+            return;
+        }
+
+        // Extract event name - required field
+        const char* eventName = json_object_get_string(jsonObject, "EventName");
```
adteran, line 251: who has ownership of this string, is this a leak?

src/common/telemetry/lib/Telemetry.cpp
```diff
@@ -0,0 +1,365 @@
...
+        {
+            OsConfigLogError(m_log, "Unexpected parameter '%s' for event '%s'", jsonKey.c_str(), eventName.c_str());
+            return false;
+        }
+    }
+
+    return true;
+}
+
+void TelemetryManager::ProcessJsonLine(const std::string& jsonLine)
```
adteran, line 222: Readability, refactor into helper functions

src/common/telemetry/lib/Telemetry.cpp
```diff
@@ -0,0 +1,365 @@
...
+            json_value_free(jsonValue);
+            return;
+        }
+
+        // Collect all JSON keys for validation
+        std::set<std::string> jsonKeys;
+        size_t count = json_object_get_count(jsonObject);
+        for (size_t i = 0; i < count; i++)
+        {
+            const char* key = json_object_get_name(jsonObject, i);
```
adteran, line 264: raii, who owns this pointer?

src/common/telemetry/lib/Telemetry.cpp
```diff
@@ -0,0 +1,365 @@
...
+            {
+                jsonKeys.insert(std::string(key));
+            }
+        }
+
+        // Validate parameters against the event's parameter set
+        if (!ValidateEventParameters(std::string(eventName), jsonKeys))
+        {
+            OsConfigLogError(m_log, "Parameter validation failed for event '%s': %s", eventName, jsonLine.c_str());
+            json_value_free(jsonValue);
```
adteran, line 275: this should not be needed, use raii.

src/common/telemetry/lib/Telemetry.cpp
```diff
@@ -0,0 +1,365 @@
...
+        // Log the event with all properties
+        m_logger->LogEvent(event);
+    }
+    catch (const std::exception& e)
+    {
+        OsConfigLogError(m_log, "Exception processing JSON line '%s': %s", jsonLine.c_str(), e.what());
+    }
+
+    // Cleanup
+    if (jsonValue)
```
adteran, line 359: same here use raii...

adteran (changes requested):

src/adapters/mc/complianceengine/CMakeLists.txt
```diff
@@ -23,13 +23,20 @@ target_link_libraries(OsConfigResourceComplianceEngine
         parsonlib
         complianceenginelib)
 
+# Add telemetry binary if enabled
+if(BUILD_TELEMETRY)
```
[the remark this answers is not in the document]
AhmedBM: There will be a follow-up PR with telemetry on the ComplienceEngine functions as well, similar to what we currently

src/common/telemetry/exe/main.cpp
AhmedBM: Refactored much of this, simplified the flow and used getopt for cli parsing

src/common/telemetry/exe/main.cpp
AhmedBM: i changed much of the parsing logic to simplify it. The Binary is only run with `-v` when compiling in Debug. This produces verbose logging from the 1DS library which we can use to debug issues with the client.

src/common/telemetry/lib/Telemetry.cpp
AhmedBM: refactored

src/common/telemetry/exe/main.cpp
```diff
@@ -0,0 +1,190 @@
...
+        return false;
+    }
+    args.filepath = non_option_args[0];
+
+    // Parse optional teardown time argument (second non-option argument)
+    if (non_option_args.size() > 1)
+    {
+        try
+        {
+            args.teardown_time = std::stoi(non_option_args[1]);
```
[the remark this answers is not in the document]
AhmedBM: Well today i learned! I've added it

src/common/telemetry/lib/Telemetry.cpp
```diff
@@ -0,0 +1,321 @@
...
+    auto& logConfig = LogManager::GetLogConfiguration();
+
+    logConfig[CFG_BOOL_ENABLE_TRACE] = enableDebug;
+    logConfig[CFG_INT_TRACE_LEVEL_MIN] = 0;
+    logConfig[CFG_INT_MAX_TEARDOWN_TIME] = teardownTime;
+
+    m_logger = LogManager::Initialize(API_KEY);
+}
+
+bool TelemetryManager::Initialize(bool enableDebug, int teardownTime)
```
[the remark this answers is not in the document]
AhmedBM: Ive refactored the lib/bin, ive removed the c-style constructs in favor of c++

src/common/telemetry/lib/Telemetry.cpp
```diff
@@ -0,0 +1,229 @@
...
+#include <nlohmann/json.hpp>
+#include <sstream>
+#include <thread>
+
+using namespace MAT;
+
+namespace Telemetry
+{
+
+TelemetryManager::TelemetryManager(bool enableDebug, int teardownTime, OsConfigLogHandle logHandle)
```
adteran: Make the default constructor private so it can't be invoked outside of the class.

src/common/telemetry/lib/Telemetry.cpp
```diff
@@ -0,0 +1,229 @@
...
+    : m_log(logHandle)
+    , m_logManager(nullptr)
+    , m_logger(nullptr)
+{
+#if defined(DEBUG) || defined(_DEBUG) || !defined(NDEBUG)
```
adteran: how is this used? Why is it needed? Could you point to the tests supporting both positive and negative setting of DEBUG and the other variables?

src/common/telemetry/lib/Telemetry.cpp
```diff
@@ -0,0 +1,229 @@
...
+    {
+        OsConfigLogInfo(m_log, "Telemetry initialized successfully.");
+
+        m_logger = m_logManager->GetLogger(API_KEY, "logger_direct");
+        if (!m_logger)
+        {
+            OsConfigLogError(m_log, "Failed to get logger instance");
+        }
+    }
+    else
```
adteran: Do we need to throw here? CreateLogManager failed, the object is probably on a bad state. Throw here.

src/common/telemetry/lib/Telemetry.cpp
```diff
@@ -0,0 +1,229 @@
...
+    OsConfigLogDebug(m_log, "Processing JSON line: %s", jsonLine.c_str());
+
+    nlohmann::json jsonObject;
+    try
+    {
+        jsonObject = nlohmann::json::parse(jsonLine);
+    }
+    catch(const nlohmann::json::parse_error& e)
+    {
+        // Ignore invalid JSON lines
```
adteran: log or telemetry here. why would we have invaild lines? Wouldn't we need to investigate and fix?

src/common/telemetry/Telemetry.h
```diff
@@ -95,12 +95,12 @@ static inline int64_t TsToUs(struct timespec ts)
...
-            OSConfigTelemetryCleanup();                                                                                                                \
             if (NULL != command)                                                                                                                       \
             {                                                                                                                                          \
                 ExecuteCommand(NULL, command, false, false, 0, TELEMETRY_TIMEOUT_SECONDS, NULL, NULL, NULL);                                           \
             }                                                                                                                                          \
             FREE_MEMORY(command);                                                                                                                      \
+            OSConfigTelemetryCleanup();                                                                                                                \
```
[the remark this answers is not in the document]
AhmedBM: This is never called from the Telemetry API, its called internally when from the Telemetry APIs `OSConfigProcessTelemetryFile` macro which is called when unloading ASB (for MC case).
Its job is to signal that we have completed so that we can process the telemetry and cleanup resources, the functions are NULL-safe.

src/common/telemetry/lib/Telemetry.cpp
AhmedBM: It is used, and there are also tests supporting the file argument (TelemetryBinTest::VerboseFlagWorks)

src/common/telemetry/bin/main.cpp
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    std::string init_message = "Initializing telemetry with verbose=" + std::string(args.verbose ? "true" : "false");
+    if (args.teardown_time != TELEMETRY_TIMEOUT_SECONDS) // Default teardown time
+    {
+        init_message += " and teardown_time=" + std::to_string(args.teardown_time) + "s";
+    }
+    OsConfigLogInfo(g_log, "%s", init_message.c_str());
+
+    try
```
adteran: Extend to all main, the try should start at line 32

src/common/telemetry/bin/main.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+        telemetryManager.ProcessJsonFile(args.filepath);
+        OsConfigLogInfo(g_log, "Processed telemetry JSON file: %s", args.filepath.c_str());
+
+        if (std::remove(args.filepath.c_str()) != 0)
+        {
+            OsConfigLogError(g_log, "Warning: Failed to delete JSON file: %s", args.filepath.c_str());
+        }
+    }
+    catch (const std::exception& e)
```
adteran: Are there any other exception types you need to catch?

src/common/telemetry/bin/tests/TelemetryBinTests.cpp
```diff
@@ -0,0 +1,319 @@
...
+        }
+    }
+
+    std::string CreateTestJsonFile(const std::string& content)
+    {
+        std::string filePath = m_testDir + "/test_events.json";
+        std::ofstream file(filePath);
+        EXPECT_TRUE(file.is_open());
+        file << content;
+        file.close();
```
adteran: missing guard. add guard to close on exit

src/common/telemetry/bin/tests/TelemetryBinTests.cpp
```diff
@@ -0,0 +1,319 @@
...
+    std::vector<std::string> argList = {"telemetrybin"};
+    char** argv = CreateArgv(argList);
+
+    CommandLineArgs args;
+    bool result = ParseCommandLineArgs(argList.size(), argv, args, nullptr);
+
+    EXPECT_FALSE(result);
+    EXPECT_TRUE(args.filepath.empty());
+
+    FreeArgv(argv, argList.size());
```
adteran: missing guard, add guard (freeargv) at line 86

adteran: same for the other functions in this file.

src/common/telemetry/lib/tests/TelemetryTests.cpp
```diff
@@ -0,0 +1,103 @@
...
+    // Helper function to create a test JSON file
+    bool CreateTestJsonFile(const std::string& content)
+    {
+        std::ofstream file(m_testJsonFile);
+        if (!file.is_open())
+        {
+            return false;
+        }
+        file << content;
+        file.close();
```
adteran: calls to close are not needed and can be removed. Correct?

MariusNi (approved):
LGTM

One suggestion for the future:

Try to replace the '...' in catch() with exception objects so you can extract the error code and log and also, do not risk to hide potential crashes.

PR #1129, opened 2025-09-10
Added missing vcpkg build dependencies for telemetry - openssl, curl

AhmedBM:
## Description

The new telemetry requires OpenSSL/cURL which depends on some build tooling being present:

* Updated Git to v2.14.0 in CentOS 7, Oracle Linux 7, and RHEL 7 (vcpkg dependency)
* Added Perl/Perl-IPC
* Added other misc missing build tooling (gexttext/pkgconfig/zlib) and cleanup

## Checklist

- [X] I ran [pre-commit](https://pre-commit.com/) on my changes prior to this PR

### DIFF
--- a/devops/docker/almalinux-9-amd64/Dockerfile
+++ b/devops/docker/almalinux-9-amd64/Dockerfile
@@ -11,6 +11,8 @@ RUN dnf install -y \
     libtool \
     make \
     openssl-devel \
+    perl \
+    perl-IPC-Cmd \
     rpm-build \
     tar \
     unzip \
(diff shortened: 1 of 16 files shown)


AhmedBM:
> If the telemetry will be separate binary - shouldn't we have separate build process for it, Separate cmake, separate Dockerfiles etc. The reason for docker files is that once we update them we might be tempted to use them in main azure-osconfig WDYT?

It is a separate build process, its done using `ExternalProject_Add` with a custom toolchain (gcc 5.5) for the case where we build MC packages and the same toolchain when more modern compilers are used. This changeset contains all necessary changes for all our build environments.

main/dev share the same container environments

PR #1159, opened 2025-10-07
Minimal double free guard for FREE_MEMORY macro

MariusNi:
## Description

Minimal double free guard for FREE_MEMORY macro. The only kept piece from PR #1150 as we want to keep the rule to check against NULL for freeing memory while we check against NULL for successful memory allocation. 

## Checklist

### DIFF
--- a/src/common/commonutils/CommonUtils.h
+++ b/src/common/commonutils/CommonUtils.h
@@ -16,9 +16,10 @@
 #define UNUSED(a) (void)(a)
 
 #define FREE_MEMORY(a) {\
-    if (NULL != a) {\
-        free(a);\
-        a = NULL;\
+    void* b = (a);\
+    (a) = NULL;\
+    if (NULL != b) {\
+        free(b);\
     }\
 }\
 


simonjaeger (approved):

robertwoj-microsoft (changes requested):
1. I don't see how this prevents from double free. Previous version already nullified the argument.
2. If it really prevents from double free, we __MUST__ log this at least, not hide the problems and delay crashes - this is much worse than just letting the program crash as W already explained.
3. I think you've forgotten to add tests? Please add a test that confirms the solution works.

MariusNi:
> 1. I don't see how this prevents from double free. Previous version already nullified the argument.
> 2. If it really prevents from double free, we **MUST** log this at least, not hide the problems and delay crashes - this is much worse than just letting the program crash as W already explained.
> 3. I think you've forgotten to add tests? Please add a test that confirms the solution works.

Actually, this does not hide anything, neither fixes any current bug (the double free leading to GC worker crash was fixed in another PR), it is just healthy practice how to free a pointer, which can become useful if the code is ever used in MT environment. The change is no-op really. Thanks for your feedback though

MariusNi:
Not really necessary right now so closing

PR #1254, opened 2026-03-11
devops: cmake version up && use vcpkg to install build deps

kkanas:
## Description

azure-osconfig uses vcpkg to manage dependencies. 
The devops containers are used to build & test project on various distros.
It used to be that all pre-requisites are installed in every container, but now with package build vcpkg is downloading prerequisites from network. This causes random build failures when mirror rejects package downloads. 

Optimize this by adding bootstrap vcpkg and install in each container so that containers have all build tools up to date.

## Checklist


devops/docker/almalinux-9-amd64/Dockerfile
```diff
@@ -24,3 +24,12 @@ WORKDIR /git
 # CMake
 RUN git clone https://github.com/Kitware/CMake --recursive -b ${CMAKE_VERSION}
 RUN cd CMake && ./bootstrap && make -j$(nproc) && make install && hash -r && rm -rf /git/CMake
+
+COPY .  /src
+WORKDIR /src
+RUN ls -al
```
robertwoj-microsoft: I guess these could be dropped now

devops/docker/almalinux-9-amd64/Dockerfile
```diff
@@ -1,7 +1,7 @@
 # TODO: Add mcr.io retagged images
 FROM almalinux:9
 
-ARG CMAKE_VERSION=v3.21.7
+ARG CMAKE_VERSION=v3.30.1
```
robertwoj-microsoft, line 4: Was the cmake upgrade required?

devops/docker/almalinux-9-amd64/Dockerfile
```diff
@@ -24,3 +24,12 @@ WORKDIR /git
 # CMake
 RUN git clone https://github.com/Kitware/CMake --recursive -b ${CMAKE_VERSION}
 RUN cd CMake && ./bootstrap && make -j$(nproc) && make install && hash -r && rm -rf /git/CMake
+
+COPY .  /src
+WORKDIR /src
+RUN ls -al
+RUN ls -al external || true
+RUN ls -al external/vcpkg || true
+RUN cd external/vcpkg && sed -e "s/\"2\\.7\\.4\"/\"1\\.0\\.0\"/g" -i scripts/vcpkg-tools.json && ./bootstrap-vcpkg.sh
```
robertwoj-microsoft: Why is this sed used? Maybe add a comment

devops/docker/ubuntu-14.04-amd64/Dockerfile
robertwoj-microsoft, line 57: No bootstrap here?

devops/docker/almalinux-9-amd64/Dockerfile
kkanas: Yes - the vcpkg always gets that version. 
If we preinstall this we will reduce build times (no download & build) and reduce failures due to mirror rate limiting github.

devops/docker/ubuntu-14.04-amd64/Dockerfile
kkanas: It's intended. 
1. It doesn't work (I forgot the error)
2. We don't have up to date rebuild of ubunt14 in .github/test-matrix.json

@simonjaeger  what is your opinion on Ubuntu14 as this is mostly used by your team in release process

devops/docker/almalinux-9-amd64/Dockerfile
kkanas: I delete sed ... if possible if not I will add comment - I think that was necesary only on old distro that not had git in version 2.7.4 or on some github runners

devops/docker/almalinux-9-amd64/Dockerfile
```diff
@@ -24,3 +24,9 @@ WORKDIR /git
 # CMake
 RUN git clone https://github.com/Kitware/CMake --recursive -b ${CMAKE_VERSION}
 RUN cd CMake && ./bootstrap && make -j$(nproc) && make install && hash -r && rm -rf /git/CMake
+
+COPY .  /src
```
simonjaeger, line 30: Will this have any other unintended consequences (now that we copy the sources into the container)? Are you relying on these sources later or should this be cleaned up?

devops/docker/ubuntu-14.04-amd64/Dockerfile
simonjaeger: I'm fine with this. We don't have disruptions on this image and even if we did, those are OK to rerun.

devops/docker/almalinux-9-amd64/Dockerfile
simonjaeger: Just trying to understand since I guess these sources are not going to be relevant once snapped into the image?

devops/docker/ubuntu-14.04-amd64/Dockerfile
kkanas: Ok closing then.

devops/docker/almalinux-9-amd64/Dockerfile
kkanas: Re:
1 Will this have any other unintended consequences (now that we copy the sources into the container)? 
Should not have ... but not to make precedence(or too eays to grab src from build container) I removed src from the final container
Re 2: 
 Just trying to understand since I guess these sources are not going to be relevant once snapped into the image?
 No the sources should not be there.
 If there was a better, simpler way of extracting vcpkg than I would use it, unfortunately it needs the external stuff -so it's just easier to pull everything

robertwoj-microsoft (approved):

simonjaeger (approved):

devops/docker/almalinux-9-amd64/Dockerfile
simonjaeger: Got it - thanks for the context, and for making the change (seems cleaner) - appreciate it.

PR #1258, opened 2026-03-18
SystemdParameter: block parameter do not need `[]`, add  passOnNotFound to make audit succeed if no value is found 

kkanas:
## Description
SystemdParameter: block parameter do not need `[]`, add passOnNotFound to make audit succeed if no value is found (aka default is ok but we have no way of checking it) 

## Checklist


wupeka (approved):

danielszot (approved):
Would be best to add also at least one test for `params.passOnNotFound = false` set explicitly to catch when the default value will be changed (on purpose or by mistake).

kkanas:
> Would be best to add also at least one test for `params.passOnNotFound = false` set explicitly to catch when the default value will be changed (on purpose or by mistake).

I believe that the current test are sufficient to test value of passOnNotFound, there are that test for missing values and giving Compliant/Non Compliant. There are test that Check if value is there and flag passOnNotFound. 
Unless you have more specific test in mind then I can add it .

PR #1260, opened 2026-03-20
AuditEnsureAccountsWithoutShellAreLockedParams: add additional optional parameter to exclude users with not valid shells

kkanas:
## Description

AuditEnsureAccountsWithoutShellAreLockedParams: add additional optional parameter to exclude users with not valid shells

Azure Linux3 in addition  to excluding UID_MIN also excludes users with not valid shell from failing compliance

## Checklist

### DIFF
--- a/src/modules/complianceengine/src/lib/procedures/EnsureAccountsWithoutShellAreLocked.cpp
+++ b/src/modules/complianceengine/src/lib/procedures/EnsureAccountsWithoutShellAreLocked.cpp
@@ -27,6 +27,7 @@ Result<Status> AuditEnsureAccountsWithoutShellAreLocked(const AuditEnsureAccount
     }
 
     const auto validShells = ListValidShells(context);
+
     if (!validShells.HasValue())
     {
         OsConfigLogError(context.GetLogHandle(), "Failed to get valid shells: %s", validShells.Error().message.c_str());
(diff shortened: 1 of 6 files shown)


src/modules/complianceengine/src/lib/procedures/EnsureAccountsWithoutShellAreLocked.h
```diff
@@ -16,6 +16,8 @@ struct AuditEnsureAccountsWithoutShellAreLockedParams
     Optional<Separated<std::string, ','>> excludeUsers;
     /// Parse /etc/login.defs and skip users with uid below UID_MIN
     Optional<bool> skip_below_uid_min = false;
+    /// Parse /etc/shells and if user does not have valid shell skip it
+    Optional<bool> skip_not_valid_shells = false;
```
robertwoj-microsoft:
```suggestion
    Optional<bool> skip_invalid_shells = false;
```

src/modules/complianceengine/src/lib/procedures/EnsureAccountsWithoutShellAreLocked.cpp
```diff
@@ -75,7 +75,14 @@ Result<Status> AuditEnsureAccountsWithoutShellAreLocked(const AuditEnsureAccount
         {
             continue;
         }
+
         bool shouldSkip = false;
+        if (params.skip_not_valid_shells.HasValue() && params.skip_not_valid_shells.Value())
```
robertwoj-microsoft: If you assign a default value at the structure level, you'll always have HasValue() as true, so this conditional can be simplified, potentially with an assertion:

```suggestion
        assert(params.skip_not_valid_shells.HasValue());
        if (params.skip_not_valid_shells.Value()))
```

But up to you this is obviously correct as is.

robertwoj-microsoft (approved):

wupeka (approved):

PR #1267, opened 2026-04-16
Introduce a NotApplicable state in compliance engine

robertwoj-microsoft:
## Description

Introduce a new state to Compliance Engine.

### Rationale

We've started with Machine Configuration as our main direction which only accepts the `Compliant` or `NonCompliant` states. However, there are many rules that could benefit from N/A state. So far we've followed the NRP interface from the bottom, but the idea of this PR is to use such state, but only switch from N/A to C at the NRP interface level.

### Most notable changes
- A new label to the `Status` enum
- A new set of emojis for NRP interface phrase formatting - `NestedListFormatter`
- The logic follows the 3VL (`Compliant` - True, `NonCompliant` - False, `NotApplicable` - Undefined)
- Added a sinlge use-case in the `EnsureSysctl` procedure - if running inside containers some `procfs` files may not exist

## Checklist

### DIFF
--- a/src/modules/complianceengine/src/lib/ComplianceEngineInterface.cpp
+++ b/src/modules/complianceengine/src/lib/ComplianceEngineInterface.cpp
@@ -186,7 +186,7 @@ int ComplianceEngineMmiGet(MMI_HANDLE clientSession, const char* componentName, 
         }
 
         auto payloadString = result.Value().payload;
-        if (result.Value().status == Status::Compliant)
+        if ((result.Value().status == Status::Compliant) || (result.Value().status == Status::NotApplicable))
         {
             payloadString = "PASS" + payloadString;
         }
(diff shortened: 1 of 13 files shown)


wupeka (approved):

src/modules/complianceengine/src/lib/ComplianceEngineInterface.cpp
```diff
@@ -186,7 +186,7 @@ int ComplianceEngineMmiGet(MMI_HANDLE clientSession, const char* componentName,
         }
 
         auto payloadString = result.Value().payload;
-        if (result.Value().status == Status::Compliant)
+        if ((result.Value().status == Status::Compliant) || (result.Value().status == Status::NotApplicable))
```
kkanas, line 189: Could we decide on some Policy context should we match Status::NotApplicable to Succes or Failure ? 
This could be exposed to user so that it can be decided on policy creation (e.g. whit custom rule) ? 
The default  for mapping to success is ok.

src/modules/complianceengine/src/lib/Evaluator.h
```diff
@@ -41,7 +41,15 @@ class PayloadFormatter
...
 class NestedListFormatter : public PayloadFormatter
 {
-    static void FormatNode(const IndicatorsTree::Node& node, std::ostringstream& result, int depth, bool ignored);
+    enum class Ignored
+    {
+        No,
+        Yes
+    };
+    static const std::map<std::pair<Status, NestedListFormatter::Ignored>, const char*> sEmojiMap;
+    static constexpr std::size_t cMaxNodeIndicators = 5;
```
kkanas: Can we add comment what is this value do ?

kkanas (approved):

src/modules/complianceengine/src/lib/Evaluator.h
robertwoj-microsoft: Added

kkanas: Thanks

kkanas (approved):

PR #1276, opened 2026-05-28
assessor: handle escaped quotes in MOF GetValue

robertwoj-microsoft:
## Description

The MOF format escapes inner double quotes as " inside quoted string values. The previous GetValue() implementation searched for the next '"' literally, which truncated values like:

    DesiredObjectValue = "{"mountPoint":"/tmp"}";

to just '{\' before the JSON parser ever saw them. The compliance engine then fell back to a key=value parser, which rejected '{' as a key character and emitted "Invalid key: only alphanumeric and underscore characters are allowed" once per rule with structured parameters.

Walk the string instead, honouring " and \ escape sequences, so a JSON-encoded DesiredObjectValue arrives at UpdateUserParameters() in its original unescaped form.

**Note**: I have some additional work prepared which adds tests for the assessor tool including fuzzer, so this change will be tested in following PRs

## Checklist

### DIFF
--- a/src/modules/complianceengine/src/assessor/Mof.cpp
+++ b/src/modules/complianceengine/src/assessor/Mof.cpp
@@ -20,12 +20,32 @@ string GetValue(const std::string& line)
     {
         return std::string();
     }
-    const auto end = line.find('"', start + 1);
-    if (end == std::string::npos)
+    // Walk from the opening quote, honouring \" and \\ escape sequences so
+    // that MOF string values like "{\"key\":\"val\"}" are returned
+    // unescaped as {"key":"val"} instead of being truncated at the first
+    // inner quote.
+    std::string result;
+    size_t pos = start + 1;
+    while (pos < line.size())
     {
-        return std::string();
+        if (line[pos] == '\\' && pos + 1 < line.size())
+        {
+            const char next = line[pos + 1];
+            if (next == '"' || next == '\\')
+            {
+                result += next;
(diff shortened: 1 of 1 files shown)


danielszot (approved):

src/modules/complianceengine/src/assessor/Mof.cpp
```diff
@@ -20,12 +20,32 @@ string GetValue(const std::string& line)
     {
         return std::string();
     }
-    const auto end = line.find('"', start + 1);
-    if (end == std::string::npos)
+    // Walk from the opening quote, honouring \" and \\ escape sequences so
```
kkanas, line 23: Can we have unit tests and handle nested quotes 
aka 
```
"k\\\"e\\\"y"
```
(best if we can handle unlimited nested quotes or bail out if 3 nesting levels )

kkanas (approved):

src/modules/complianceengine/src/assessor/Mof.cpp
robertwoj-microsoft: I will add a follow up PR to address this as discussed in private. The code is correct as is, but it's permissive, so it may be surprising sometimes. But it is done this was as the MOF parsing will not be needed soon.

The follow-up will add more strict MOF file parsing, but tied specifically to our form of MOF (what we generate by AE) and tied to that, so we have better coverage until MOF is substituted with better format.

PR #1281, opened 2026-06-02
Create distinct context classess for assessor and GC module

robertwoj-microsoft:
## Description

Drop `/var/lib/GuestConfig` static path that is directly tied to the GC module. Since we're adding the assessor local tool scanning capability which may work on a system without an installed GC modules, this path must be abstracted out of the source.

### New `GetStatePath()` interface
A new abstract method is added to dispatch specific values to implementing classes.

### New context classes
New classess are added: `GuestConfigurationContext` and `AssessorContext` to enclose different behavior in different types. They differ only in the state path values.

## Checklist

### DIFF
--- a/src/modules/complianceengine/src/lib/AssessorContext.h
+++ b/src/modules/complianceengine/src/lib/AssessorContext.h
@@ -7,20 +7,42 @@
 #include "CommonContext.h"
 #include "Logging.h"
 
+#include <ftw.h>
+#include <stdexcept>
+#include <unistd.h>
+
 namespace ComplianceEngine
 {
-
-namespace
-{
-constexpr char assessorStatePath[] = "/tmp/compliance-engine";
-} // namespace
 
 class AssessorContext : public CommonContext
 {
 public:
     AssessorContext(OsConfigLogHandle log)
-        : CommonContext(log, assessorStatePath)
+        : CommonContext(log, CreateTempDir())
     {
+    }
+    AssessorContext(const AssessorContext&) = delete;
+    AssessorContext& operator=(const AssessorContext&) = delete;
+    AssessorContext(AssessorContext&&) = delete;
+    AssessorContext& operator=(AssessorContext&&) = delete;
+
+    ~AssessorContext() override
+    {
+        nftw(
+            GetStatePath().c_str(),
+            [](const char* fpath, const struct stat*, int typeflag, struct FTW*) -> int { return (typeflag == FTW_DP) ? rmdir(fpath) : unlink(fpath); },
+            64, FTW_DEPTH | FTW_PHYS);
+    }
+
+private:
+    static std::string CreateTempDir()
+    {
+        char tmpl[] = "/tmp/compliance-engine-assessor.XXXXXX";
+        if (mkdtemp(tmpl) == nullptr)
+        {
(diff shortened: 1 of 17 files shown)


wupeka (approved):

kkanas (approved):

src/modules/complianceengine/src/lib/AssessorContext.h
```diff
@@ -0,0 +1,28 @@
...
+
+#include "CommonContext.h"
+#include "Logging.h"
+
+namespace ComplianceEngine
+{
+
+namespace
+{
+constexpr char assessorStatePath[] = "/tmp/compliance-engine";
```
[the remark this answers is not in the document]
robertwoj-microsoft: Fixed with `mkdtemp` + unit tests.

src/modules/complianceengine/src/lib/AssessorContext.h
```diff
@@ -0,0 +1,28 @@
...
+} // namespace
```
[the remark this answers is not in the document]
robertwoj-microsoft: Moved to the class scope

robertwoj-microsoft:
@copilot please review this branch again

src/modules/complianceengine/src/lib/AssessorContext.h
```diff
@@ -0,0 +1,61 @@
...
+    AssessorContext(const AssessorContext&) = delete;
+    AssessorContext& operator=(const AssessorContext&) = delete;
+    AssessorContext(AssessorContext&&) = delete;
+    AssessorContext& operator=(AssessorContext&&) = delete;
+
+    ~AssessorContext() override
+    {
+        nftw(
+            GetStatePath().c_str(),
+            [](const char* fpath, const struct stat*, int typeflag, struct FTW*) -> int
```
kkanas: I like it - but after the change  who owns the mStatePath the AssessorContext or the CommonContext
You destroy it in  ~AssessorContext ... but the mPath is in CommonContext. I have fealing that later on this will create subtle errors.
WDYT ?

kkanas (approved):

src/modules/complianceengine/src/lib/AssessorContext.h
robertwoj-microsoft: The string is valid as it's contained in the base class which is destroyed after `AssessorContext`. We're taking `c_str` which is a pointer to a temporary (`GetStatePath` returns a copy), but lifetime is extended to the full expression, which makes the const char* valid.

Virtual dispatch is not a concern here as well as we're not overriding `GetStatePath`.

However, IF we modified the `GetStatePath` signature to return `const string&`, then we could potentially end up with a dangling pointer returned from `c_str`, so I made an explicit local copy of the state path string.

robertwoj-microsoft:
@copilot please review again

danielszot (approved):

src/modules/complianceengine/src/lib/AssessorContext.h
kkanas: With the design as is I was more afraid of more using GetStatePath() but they GetStatePath is destroyed in Child class not in Parent

kkanas (approved):

PR #1284, opened 2026-06-09
assessor: harden process and input handling

robertwoj-microsoft:
## Description

assessor: harden input, log-file, and MOF parsing against hostile input

The assessor runs as root, so a malformed or attacker-influenced input
MOF, log-file path, or argument must not let it follow symlinks, clobber
sensitive files, or exhaust memory. Extract the input-integrity logic
into a testable InputSecurity library and tighten the surrounding paths.

Input file (--input):
- Open with O_RDONLY|O_NOFOLLOW|O_CLOEXEC and verify ownership/mode via
  fstat() on the held fd (no lstat-then-open TOCTOU window).
- Reject non-regular files (S_ISREG): a root-owned FIFO or device could
  otherwise block the read forever or stream unbounded data. Streaming
  inputs must use stdin instead.
- Reject path traversal ("..") and a group/world-writable or non-root
  parent directory (prevents a rename-swap onto a hostile file).
- Cap total bytes read at 8 MiB (kMaxInputBytes) for both --input and
  stdin to bound memory use.

Log file (--log-file):
- Validate the path before OpenLog() with RefuseUnsafeLogFile: reject
  traversal, a writable/non-root parent directory, and (if the target
  already exists) symlinks, non-regular files, non-root ownership, and
  group/world-writable modes. The shared logging code opens the path
  with a symlink-following append and chmod's it as root, so this
  prevents redirecting root's writes onto a sensitive file.
- A residual check-to-use window remains because OpenLog() is path-only
  and TrimLog() re-opens the path on rotation; the required root-owned,
  non-writable parent closes it in practice. Documented in the
  Main.cpp threat model and the RefuseUnsafeLogFile declaration.

MOF parser:
- Bound a single line at 4 MiB (kMaxLineLength) and total entries at
  100000 (kMaxMofEntries) to prevent unbounded growth on hostile input.

CLI parsing:
- Reject empty values for --log-file, --section, and --format.
- Fix the "optinos" typo in the help text.

Add InputSecurity unit tests (Result<int> API, FIFO refusal, and the
RefuseUnsafeLogFile cases); root-only checks skip cleanly as non-root.

## Checklist

### DIFF
--- a/src/modules/complianceengine/src/assessor/InputSecurity.cpp
+++ b/src/modules/complianceengine/src/assessor/InputSecurity.cpp
@@ -84,20 +84,22 @@ Result<int> OpenVerifiedInput(const std::string& path, OsConfigLogHandle logHand
         if (errno == ELOOP)
         {
             OsConfigLogError(logHandle, "Refusing to open input file '%s': path is a symlink.", path.c_str());
-            return Error("path is a symlink");
+            return Error("path is a symlink", ELOOP);
         }
-        const std::string msg = std::strerror(errno);
+        const int savedErrno = errno;
+        const std::string msg = std::strerror(savedErrno);
         OsConfigLogError(logHandle, "Failed to open input file '%s': %s", path.c_str(), msg.c_str());
-        return Error(msg);
+        return Error(msg, savedErrno);
     }
 
     struct stat st;
     if (::fstat(fd, &st) != 0)
     {
-        const std::string msg = std::strerror(errno);
+        const int savedErrno = errno;
+        const std::string msg = std::strerror(savedErrno);
         OsConfigLogError(logHandle, "Failed to stat input file '%s': %s", path.c_str(), msg.c_str());
         ::close(fd);
-        return Error(msg);
+        return Error(msg, savedErrno);
     }
     if (!S_ISREG(st.st_mode))
     {
@@ -108,7 +110,7 @@ Result<int> OpenVerifiedInput(const std::string& path, OsConfigLogHandle logHand
         // which deliberately bypasses these on-disk integrity checks.
         OsConfigLogError(logHandle, "Refusing to read input file '%s': not a regular file.", path.c_str());
         ::close(fd);
-        return Error("not a regular file");
+        return Error("not a regular file", ENOTSUP);
     }
 
     // O_NONBLOCK was set solely to prevent open() from blocking on a FIFO.
@@ -119,24 +121,25 @@ Result<int> OpenVerifiedInput(const std::string& path, OsConfigLogHandle logHand
         const int flags = ::fcntl(fd, F_GETFL);
         if (flags < 0 || ::fcntl(fd, F_SETFL, flags & ~O_NONBLOCK) < 0)
         {
-            const std::string msg = std::strerror(errno);
+            const int savedErrno = errno;
+            const std::string msg = std::strerror(savedErrno);
             OsConfigLogError(logHandle, "Failed to clear O_NONBLOCK on input file '%s': %s", path.c_str(), msg.c_str());
             ::close(fd);
-            return Error(msg);
+            return Error(msg, savedErrno);
         }
     }
     if (st.st_uid != 0)
     {
         OsConfigLogError(logHandle, "Refusing to read input file '%s': not owned by root (uid %u).", path.c_str(), static_cast<unsigned>(st.st_uid));
         ::close(fd);
-        return Error("not owned by root");
+        return Error("not owned by root", EPERM);
     }
     if (st.st_mode & (S_IWGRP | S_IWOTH))
     {
         OsConfigLogError(logHandle, "Refusing to read input file '%s': writable by group or others (mode %04o).", path.c_str(),
             static_cast<unsigned>(st.st_mode & 07777));
         ::close(fd);
-        return Error("writable by group or others");
+        return Error("writable by group or others", EPERM);
     }
     return fd;
 }
--- a/src/modules/complianceengine/src/assessor/InputSecurity.cpp
+++ b/src/modules/complianceengine/src/assessor/InputSecurity.cpp
@@ -1,0 +1,174 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include <InputSecurity.hpp>
+#include <Logging.h>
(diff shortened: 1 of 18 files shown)


robertwoj-microsoft:
@copilot This PR aims to reduce potential attack surface by adding a set of safeguards to the compliance-engine-assessor binary which is designed to run as root. The input file is a MOF file which contains defintions of functions which may affect the working system, especially if the remediation action is requested. The MOF file parsing harderning is intentionally left for later follow-up PRs. Try to investigate if you see more potential issues than covered in this PR.

src/modules/complianceengine/src/assessor/Main.cpp
```diff
@@ -129,28 +243,95 @@ int main(int argc, char* argv[])
...
-    istream& inputStream = options.input.empty() ? std::cin : file;
+    istream& inputStream = options.input.empty() ? std::cin : fileStream;
     string line;
     auto status = Status::Compliant;
     bool hasError = false;
+    // Bytes consumed from the outer scan loop. The --input path is already
+    // bounded by kMaxInputBytes above; this counter applies the same ceiling
+    // to the stdin path without buffering all of stdin (per the planned MOF
+    // streaming rework). Per-entry line-length and entry-count limits are
+    // enforced inside Resource::ParseSingleEntry.
```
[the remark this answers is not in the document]
robertwoj-microsoft: Fixed.

src/modules/complianceengine/src/assessor/Main.cpp
```diff
@@ -1,17 +1,99 @@
...
+//    1. Parent directory (stat): must be root-owned and not writable by
+//       group or others. A writable directory enables a rename-swap attack:
+//       an attacker can unlink the validated file and place a hostile one
+//       before the process reads it.
+//
+//    2. open(O_RDONLY|O_NOFOLLOW|O_CLOEXEC): the kernel refuses symlinks in
+//       the final path component atomically (ELOOP), eliminating the
+//       lstat-then-open TOCTOU window. Note: symlinks in intermediate path
+//       components are not checked; the operator is trusted to supply a
+//       straightforward path.
```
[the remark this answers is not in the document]
robertwoj-microsoft: Description updated.

src/modules/complianceengine/src/assessor/InputSecurity.cpp
```diff
@@ -0,0 +1,186 @@
...
+    //   - require a root-owned, non-group/world-writable parent directory
+    //     (prevents a rename-swap onto a hostile target),
+    //   - if the path already exists, reject symlinks, non-root ownership, or
+    //     group/world-writable modes. A non-existent path is fine because it
+    //     will be created inside the parent directory we just validated.
+    if (RefusePathTraversal(path, logHandle))
+    {
+        return true;
+    }
+    if (RefuseWritableParentDir(path, logHandle))
```
kkanas, line 162: aren't this sussectable to toctou  ? First you are checking parrent dir and then later on OpenLog is called and it's later finally openning the log file ?

kkanas: Ok I see the comment. 
I think it would be better to crate second OpenLogFd where you pass fd to OpenLog and you can prevent future TOCTOU without aditional complexity of parent directory checking, also if we would run as priviledged user (not root) that logic in RefuseWritableParentDir would have to change.

src/modules/complianceengine/src/assessor/Main.cpp
```diff
@@ -129,29 +246,97 @@ int main(int argc, char* argv[])
...
+                    tooLarge = true;
+                    break;
+                }
+                content.append(buf, static_cast<size_t>(n));
+            }
+            else if (n == 0)
+            {
+                break; // EOF
+            }
+            else if (errno != EINTR)
```
kkanas, line 286: what about other signals e.g. EWOULDBLOCK (as fd is opened as nonblock) is error too ?

kkanas: what about EAGAIN ? I believe that this could happen in unusual situation that when you are reading too fast and the device is not ready with data ?

src/modules/complianceengine/src/assessor/Main.cpp
```diff
@@ -129,29 +246,97 @@ int main(int argc, char* argv[])
...
+    // read inside Resource::ParseSingleEntry). The --input path is already
+    // bounded by kMaxInputBytes above; this shared counter applies the same
+    // ceiling to the stdin path without buffering all of stdin (per the
+    // planned MOF streaming rework). Per-entry line-length and entry-count
+    // limits are also enforced inside Resource::ParseSingleEntry.
+    size_t bytesConsumed = 0;
+    size_t entryCount = 0;
     while (std::getline(inputStream, line))
     {
+        bytesConsumed += line.size() + 1;
```
kkanas, line 322: Why `+1` is it to count `\n` ? if so can you add comment for why it's there

src/modules/complianceengine/src/assessor/InputSecurity.cpp
robertwoj-microsoft: I deliberately preserved it as is to avoid changes to the existing logging interface. Plan is to do the same logic for the logging handlers but on the kompli side where we can safely change that without affecting ASB.

robertwoj-microsoft: As discussed, I'll make a follow-up PR with logging hardening on the common code.

src/modules/complianceengine/src/assessor/Main.cpp
robertwoj-microsoft: Fixed. O_NONBLOCK is now dropped after opening - this flag is only needed for `open()`.

src/modules/complianceengine/src/assessor/Main.cpp
robertwoj-microsoft: Added the comment.

kkanas (dismissed):

wupeka (dismissed):

robertwoj-microsoft:
@copilot resolve the merge conflicts in this pull request

kkanas (approved):

wupeka (approved):

PR #1285, opened 2026-06-10
MOF parser hardening

robertwoj-microsoft:
## Description

This PR hardens `compliance-engine-assessor` MOF parsing by replacing the previous loose parser with a strict, streaming `MofResourceRange` / `MofResourceIterator` parser.

The new parser targets the exact 15-field `OsConfigResource` shape emitted by the Compliance Augmentation Engine, keeps input processing bounded, and encapsulates file-input hardening behind `MofResourceRange::Make()`.

### Changes

- Added streaming `MofResourceRange` / `MofResourceIterator` parser.
- Updated `Main.cpp` to use a range-based `for` loop over parsed MOF entries.
- Kept parser failures in-band via `Result<Resource>`.
- Enforced strict MOF validation:
  - required 15-field resource shape
  - no unknown or duplicate fields
  - validated constant fields (`ComponentName`, `ConfigurationName`, `ExpectedObjectValue`)
  - consistent rule name across `procedure*`, `init*`, `audit*`, and `remediate*` object names
- Added input limits:
  - 8 MiB total input
  - 4 MiB max line length
  - 100,000 max entries
- Moved file-input hardening into `MofResourceRange::Make(path, log)`:
  - path traversal rejection
  - writable-parent rejection
  - `O_NOFOLLOW | O_NONBLOCK | O_CLOEXEC`
  - `fstat` validation
  - regular-file, root-owner, and permission checks
- Treated empty `DesiredObjectValue` as absent payload and retained `{}` fallback for audit/remediation setup.
- Added MOF parser unit tests and expanded malformed-input coverage.
- Added optional libFuzzer target and seed corpus for parser crash testing.
- Fixed older compiler compatibility by removing an invalid explicit `noexcept` from the end-iterator constructor.

### Validation

- `compliance-engine-assessor-tests` pass locally.
- `Mof.cpp` coverage:
  - line coverage: `93.5%`
  - function coverage: `100.0%`
- Fuzzer target builds with `-DBUILD_FUZZER=ON` and seed corpus smoke test runs without crashes.

### Security Notes

The parser now rejects malformed, oversized, unexpected, or structurally unfamiliar MOF input. File-backed input is validated before streaming, while stdin remains supported with the same parser size and structure limits.

## Checklist

### DIFF
--- a/src/modules/complianceengine/src/assessor/Main.cpp
+++ b/src/modules/complianceengine/src/assessor/Main.cpp
@@ -11,8 +11,11 @@
 // are treated as operator-supplied (trusted to be benign in intent, but not
 // to be free of bugs or accidental hostile content).
 //
-//  - The input MOF parser is intentionally loose. Add fuzz coverage when
-//    extending it.
+//  - The input MOF parser is strict and streaming (see MofResourceRange): it
+//    validates the fixed field set the augmentation engine emits, rejects
+//    unknown fields, and bounds line length, total size, and entry count. It
+//    owns the input file and encapsulates the integrity checks below. A fuzzer
+//    target exercises it; extend the corpus when changing the format.
 //
 //  - Input file integrity (when --input is used; stdin bypasses all checks):
 //
@@ -35,17 +38,17 @@
 //       they cannot block the read or stream unbounded data), root-owned, and
 //       not group/world-writable.
 //
-//    4. The file is read into memory immediately after opening and the fd is
-//       closed. The fd keeps the inode reachable across the read even if the
-//       directory entry is concurrently renamed or unlinked. The total bytes
-//       read are capped (kMaxInputBytes) to bound memory use.
+//    4. The verified fd is wrapped in a stream and read incrementally (never
+//       slurped whole). The fd keeps the inode reachable across the read even
+//       if the directory entry is concurrently renamed or unlinked. The total
+//       bytes read, per-line length, and entry count are all bounded inside the
+//       streaming parser to keep memory use bounded.
 //
 //  - stdin (--input not supplied): all file integrity checks are bypassed.
 //    Streaming inputs (pipes, process substitution) must use stdin. The bytes
-//    consumed are still capped (kMaxInputBytes), and per-entry line length and
-//    total entry count are bounded by the MOF parser / scan loop.
-//    Callers in automated pipelines should always use --input with a
-//    root-owned, non-world-writable file.
+//    consumed, per-line length, and total entry count are still bounded inside
+//    the streaming MOF parser. Callers in automated pipelines should always use
+//    --input with a root-owned, non-world-writable file.
 //
 //  - umask is tightened to at least S_IRWXG|S_IRWXO (preserving any stricter
 //    inherited mask). The log file when --log-file is supplied is the primary
@@ -77,22 +80,22 @@
 //    engine spawns. Sanitizing the environment is the engine's
 //    responsibility, not the assessor's.
 
+#include "CliOptions.hpp"
+#include "CompactListFormatter.hpp"
+#include "DebugFormatter.hpp"
+#include "InputSecurity.hpp"
+#include "JsonFormatter.hpp"
+#include "Mof.hpp"
+#include "NestedListFormatter.hpp"
+
 #include <AssessorContext.h>
-#include <CliOptions.hpp>
-#include <CompactListFormatter.hpp>
-#include <DebugFormatter.hpp>
 #include <Engine.h>
-#include <InputSecurity.hpp>
-#include <JsonFormatter.hpp>
 #include <Logging.h>
-#include <Mof.hpp>
-#include <NestedListFormatter.hpp>
 #include <Optional.h>
 #include <cerrno>
 #include <cstring>
 #include <iostream>
 #include <memory>
-#include <sstream>
 #include <string>
 #include <sys/stat.h>
 #include <sys/types.h>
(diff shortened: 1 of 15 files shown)


src/modules/complianceengine/src/assessor/Mof.hpp
```diff
@@ -2,34 +2,151 @@
...
+    // into an istream (and the istream itself). For stdin/test inputs these are
+    // null and mStream references the caller's stream. Declared before
+    // mOwnedStream so the istream is destroyed before the streambuf it uses.
+    std::unique_ptr<std::streambuf> mOwnedBuf;
+    std::unique_ptr<std::istream> mOwnedStream;
+    size_t mBytesConsumed = 0;
+    size_t mEntryCount = 0;
 };
 } // namespace MOF
 } // namespace ComplianceEngine
```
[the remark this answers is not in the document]
robertwoj-microsoft: Fixed.

src/modules/complianceengine/src/assessor/Mof.hpp
```diff
@@ -2,34 +2,151 @@
 #define COMPLIANCE_ENGINE_ASSESOR_MOF_HPP
```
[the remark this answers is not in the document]
robertwoj-microsoft: Fixed.

robertwoj-microsoft:
@copilot I've modified some header inclusions too excessively and I'd like to minimize the size of this PR. Look at the summary you've generated and try to find the sole inclusion changes (the ones where the only summary is  'Switches to local include form for assessor headers.'). Try to determine which changes are really needed to keep inclusion correct.

src/modules/complianceengine/src/assessor/Main.cpp
```diff
@@ -246,104 +233,31 @@ int main(int argc, char* argv[])
...
 
-    std::istringstream fileStream;
-    if (!options.input.empty())
+    // Open the input as a strictly-validated, streaming MOF range. For --input
+    // the range encapsulates the full input-hardening posture (path-traversal
+    // rejection, root-owned non-writable parent directory, O_NOFOLLOW open, and
+    // regular-file/ownership/mode checks) and owns the file; for stdin it
+    // streams without those on-disk checks. Size, line, and entry caps are
+    // enforced inside the range.
+    auto rangeResult = options.input.empty() ? MofResourceRange::Make(std::cin, logHandle.get()) : MofResourceRange::Make(options.input, logHandle.get());
```
kkanas, line 242: I have doubts this will work very portable - the std::cin is still backed by STDIO not whole file so the getline() can work strangely in corner cases of reading from std::cin not from istringstream. 
The seekg, tellg can fail AFAIK. 
Up to you - maybe it doesn't matter that much as the std::cin is for interactive use. I leave it for your judgment.

src/modules/complianceengine/src/assessor/Mof.cpp
```diff
@@ -45,166 +92,427 @@ string GetValue(const std::string& line)
...
+    //   Optional<string>(line) — one complete line (newline consumed but not included)
+    //
+    // A non-empty partial line at EOF (no trailing newline) is returned as an
+    // error: the augmentation engine always terminates every block with '};'
+    // followed by a newline, so a line without a terminator means the input was
+    // truncated.
+    const auto readLine = [this]() -> Result<Optional<string>> {
+        string line;
+        std::istream& stream = *mStream;
+        while (true)
```
kkanas, line 322: what about stream errors should we use while(stream.good()) or sth like that ?

src/modules/complianceengine/src/assessor/Mof.cpp
```diff
@@ -45,166 +92,427 @@ string GetValue(const std::string& line)
...
-        if (line.find("ResourceID") != string::npos)
+        if (!read.Value().HasValue())
         {
-            resourceID = GetValue(line);
-            continue;
+            return Optional<Resource>(); // Clean EOF — no more entries.
         }
-
-        if (line.find("PayloadKey") != string::npos)
+        if (!TrimWhiteSpaces(read.Value().Value()).empty())
```
kkanas: How does it work IIUC std::string TrimWhiteSpaces(const std::string& str)   
so what does !std::string() and !std:string("somme trimmed string does")
also in line 375 you are using TrimWhiteSpaces again
    const string header = TrimWhiteSpaces(headerLine.Value());

kkanas (dismissed):

src/modules/complianceengine/src/assessor/Mof.cpp
danielszot: when the stream is bad it will be also EOF but just the error message will be wrong. We can do like:

```
if (ch == std::istream::traits_type::eof())
{
    if (stream.bad())
        return Error("I/O error reading MOF input", EIO);
    if (line.empty())
        return Optional<string>();   // clean EOF
    return Error("Truncated MOF input: last line has no newline terminator", EIO);
}
```

Then the message will be correct :)

Also we only check "\n", it's linux so should be rather ok but maybe just in case it;s worth consider existence of "\r"

danielszot (dismissed):

src/modules/complianceengine/src/assessor/Main.cpp
robertwoj-microsoft: The existing code is a forward-only parser. There's no seek or std::getline that would be problematic with stdin. I think it's correct as is.

src/modules/complianceengine/src/assessor/Mof.cpp
robertwoj-microsoft: `TrimWhiteSpaces` returns std::string and it's not directly convertible to bool. In this line the negation deals with .empt() which is bool - so the check is correct, we're testing for the string emptiness.

Fixed the double call to `TrimWhiteSpaces`. It was nonsense to call it twice.

src/modules/complianceengine/src/assessor/Mof.cpp
robertwoj-microsoft: Good catch, fixed and added tests.

kkanas (approved):

wupeka (approved):

wupeka (approved):

danielszot (approved):

PR #1286, opened 2026-06-10
Add Lua logging interface

robertwoj-microsoft:
### Motivation

Lua compliance scripts had `print` available, which writes directly to stdout and can clobber output. This change removes `print` from the Lua sandbox and replaces it with `ce.log.*`, routing script messages through the existing `OsConfigLog*` infrastructure (respects log levels, file and console sinks).

### Changes

| File | Change |
|------|--------|
| `LuaProcedures.cpp` | Added `LuaLogImpl(lua_State*, LoggingLevel)` + four wrappers (`LuaLogInfo/Warning/Error/Debug`). Registered `ce.log` subtable with `info`, `warning`, `error`, `debug` in `RegisterLuaProcedures()`. |
| `LuaEvaluator.cpp` | Removed `"print"` from `safeGlobals` in `SecureLuaEnvironment()`. |
| `LuaProcedures.h` | Extended doc comment on `RegisterLuaProcedures` to document `ce.log.*`. |
| `MockContext.h` | `GetLogHandle()` now returns a settable `mLogHandle` member (default `nullptr`) instead of hardcoded `nullptr`. Added `SetLogHandle()`. |
| `LuaEvaluatorTest.cpp` | 8 new tests: callability of all four levels, `print` blocked, invalid-usage cases (no arg / non-string / extra args / empty string), and an e2e test that writes to a real log file and asserts `[Lua] <message>` substrings. |

### Usage

```lua
ce.log.info("Checking sshd configuration")
ce.log.warning("Deprecated option found")
ce.log.error("Required file missing")
ce.log.debug("Raw value: " .. tostring(value))
```

### Security

The argument validation in `LuaLogImpl` applies three checks in deliberate order:

1. **Presence** — `lua_isnoneornil` rejects both a missing argument and an explicit `nil`.
2. **Arity** — `!lua_isnone(L, 2)` rejects any second argument before the type check, so `ce.log.info(false, "extra")` gives *"expected a single argument"* rather than a misleading type error.
3. **Type** — `lua_isstring` accepts strings and numbers (Lua coerces numbers to strings) but rejects booleans, tables, userdata, and functions.

After all three checks pass, the message is used only as the `%s` argument to `OsConfigLog`:

```cpp
OsConfigLog(view->ctx.GetLogHandle(), level, "[Lua] %s", message);
```

The Lua-supplied string is never the format string — format-string injection is structurally impossible regardless of message content. This follows the same validation pattern as the existing `LuaIndicatorsAddIndicator`.

## Checklist

### DIFF
--- a/src/modules/complianceengine/src/lib/LuaProcedures.h
+++ b/src/modules/complianceengine/src/lib/LuaProcedures.h
@@ -49,10 +49,10 @@ struct LuaCallContext
 //   - Snapshot semantics: iterator holds shared_ptr<const FSCache>; unaffected by background refresh.
 //   - Arguments outside unsigned 32-bit range produce Lua error.
 //
-//   ce.log.info(message)    -- log at informational level via OsConfigLogInfo
-//   ce.log.warning(message) -- log at warning level via OsConfigLogWarning
-//   ce.log.error(message)   -- log at error level via OsConfigLogError
-//   ce.log.debug(message)   -- log at debug level via OsConfigLogDebug
+//   ce.log.info(message)    -- log at informational level
+//   ce.log.warning(message) -- log at warning level
+//   ce.log.error(message)   -- log at error level
+//   ce.log.debug(message)   -- log at debug level
 //     message: string (required, non-empty) text to emit; prefixed with "[Lua] " in the log output.
 // Behavior:
 //   - Routes to the OsConfigLog* macros using the log handle from the current call context.
--- a/src/modules/complianceengine/src/lib/LuaProcedures.h
+++ b/src/modules/complianceengine/src/lib/LuaProcedures.h
@@ -48,6 +48,16 @@ struct LuaCallContext
 // Notes:
 //   - Snapshot semantics: iterator holds shared_ptr<const FSCache>; unaffected by background refresh.
 //   - Arguments outside unsigned 32-bit range produce Lua error.
+//
+//   ce.log.info(message)    -- log at informational level via OsConfigLogInfo
+//   ce.log.warning(message) -- log at warning level via OsConfigLogWarning
+//   ce.log.error(message)   -- log at error level via OsConfigLogError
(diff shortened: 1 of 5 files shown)


src/modules/complianceengine/tests/LuaEvaluatorTest.cpp
```diff
@@ -4,6 +4,7 @@
 #include "LuaEvaluator.h"
 
 #include "Indicators.h"
+#include "Logging.h"
 #include "MockContext.h"
 #include "Result.h"
```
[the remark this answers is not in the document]
robertwoj-microsoft: Fixed.

kkanas (approved):

src/modules/complianceengine/src/lib/LuaProcedures.h
```diff
@@ -48,6 +48,16 @@ struct LuaCallContext
...
 //   - Arguments outside unsigned 32-bit range produce Lua error.
+//
+//   ce.log.info(message)    -- log at informational level
+//   ce.log.warning(message) -- log at warning level
+//   ce.log.error(message)   -- log at error level
+//   ce.log.debug(message)   -- log at debug level
+//     message: string (required, non-empty) text to emit; prefixed with "[Lua] " in the log output.
+// Behavior:
+//   - Routes to the OsConfigLog* macros using the log handle from the current call context.
+//   - The message string is always a %s argument, never the format string (prevents format-string injection).
```
wupeka, line 59: Just for completness please add a test that'd log "%d", "%s", "%n", "%%" = those should be passed verbatim

wupeka (approved):

src/modules/complianceengine/src/lib/LuaProcedures.h
robertwoj-microsoft: Done.

kkanas (approved):

PR #1290, opened 2026-07-09
Add distribution check to the assessor

robertwoj-microsoft:
## Description

Add distribution check to the compliance engine assessor. This fills a gap in behavior between NRP and CLI interfaces.

## Checklist


src/modules/complianceengine/src/assessor/Main.cpp
```diff
@@ -258,6 +274,26 @@ int main(int argc, char* argv[])
...
+        // running those rules would report spurious results.
+        if (!mofEntry.benchmarkInfo.Match(engine.GetDistributionInfo().Value()))
+        {
+            OsConfigLogInfo(logHandle.get(), "Skipping entry %s: benchmark is not applicable for the current distribution", mofEntry.resourceID.c_str());
+            OsConfigLogInfo(logHandle.get(), "Current system identification: %s", std::to_string(engine.GetDistributionInfo().Value()).c_str());
+            auto overridden = engine.GetDistributionInfo().Value();
+            overridden.distribution = mofEntry.benchmarkInfo.distribution;
+            overridden.version = mofEntry.benchmarkInfo.SanitizedVersion();
+            OsConfigLogInfo(logHandle.get(), "To override this detection, place the following line inside the '%s' file: %s",
+                DistributionInfo::cDefaultOverrideFilePath, std::to_string(overridden).c_str());
```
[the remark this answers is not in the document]
robertwoj-microsoft: Fixed that and added a test. Worth to note we don't automatically generated escaped characters, but it's possible that a user does that.

src/modules/complianceengine/src/assessor/Main.cpp
```diff
@@ -258,6 +274,27 @@ int main(int argc, char* argv[])
         }
 
         const auto& mofEntry = entryResult.Value();
+
+        // Skip rules that do not target the detected distribution/version. This
```
danielszot: not rules but whole benchmark

.pre-commit-config.yaml
```diff
@@ -7,10 +7,14 @@ repos:
     - id: end-of-file-fixer
       exclude: |
          (?x)^(
-            src/tests/fuzzer/seed_corpus/.*|
+            .*fuzzer/seed_corpus/.*|
```
danielszot: this will exclude also xxxxxxfuzzer or x-fuzzer, not onlu "fuzzer" dir in every found parent dir. IDK if thats expected?

danielszot (approved):
PTAL at comments.

kkanas (approved):

src/modules/complianceengine/src/assessor/Main.cpp
robertwoj-microsoft: the idea behind skipping rules not benchmark was to support the MOF files produced by MC - they're cumulative and provide rules for a wide range of distros.
Another reason is that this won't prevent from running a subset of rules and failing at some point - if the first 100 rules are matching, they will be executed and the tool will fail at the end.
Currently the tool returns 1 if there is at least one rule skipped due to that.

.pre-commit-config.yaml
robertwoj-microsoft: There are 2 directories with seed corpus now, previous pattern was not enough, but I'll add the slash.

robertwoj-microsoft: Fixed

src/modules/complianceengine/src/assessor/Main.cpp
danielszot: im ok with that but it differs form nrp right?

danielszot (approved):

src/modules/complianceengine/src/assessor/Main.cpp
robertwoj-microsoft: it's implementation defined - gc_worker decides whether it bails out or not, we return a different error code:
```
    if (0 != BaselineIsCorrectDistribution(g_payloadKey, GetLog()))
    {
        OsConfigLogInfo(GetLog(), "[OsConfigResource.Get] BaselineIsCorrectDistribution(%s) failed", g_payloadKey);
        miResult = MI_RESULT_NOT_SUPPORTED;
        goto Exit;
    }
```

I'll switch to bailing out.

robertwoj-microsoft: Fixed.

danielszot (approved):

PR #1292, opened 2026-07-15
Add new `render` subcommand to the assessor

robertwoj-microsoft:
## Description

Refactor the way of returning results from the assessor.

### Motivation

AKS builds require split of L1/L2 rules. Currently their pipeline runs 2 consecutive scans with L1 and L2 benchmarks. However, it's a wrong approach as L1 is a subset of L2 rules and consecutive runs may yield different results as the scans are executed on a live system.

Integration in different environments may need different output formats. Currently, the assessor is able to create a single output format using the built-in formatters. But if a consumer wanted to have 2 different outputs, then we'd need to run the scan twice and we'd like to avoid it to have the results atomic.

### Single output format

The assessor now returns a single output format - JSON. This output will be used to parse and format differently with a new `render` subcommand.

### JUnit output

A new output has been provided - JUnit. It will be useful to report compliance status within pipelines.

### JSON schema

A new JSON schema file has been created to give the users a way of validation of the outputs and help agents to understand the output format structure.

### Planned follow-up

The final shape of the result JSON will be updated once the AE provides an alternative to MOF input format. This is planned to increase capabilities of the assessor:
- querying rules by tags (L1/L2 in the motivating example),
- parametrization - supported parameters will be reported with the actual values. Currently only default values are supported.

## Checklist


src/modules/complianceengine/src/assessor/tests/JsonFormatterTest.cpp
```diff
@@ -0,0 +1,285 @@
...
+{
+    JsonFormatter formatter;
+    ASSERT_FALSE(formatter.Begin(Action::Audit).HasValue());
+    auto result = formatter.Finish(Status::Compliant);
+    ASSERT_TRUE(result.HasValue());
+
+    ParsedJson doc(result.Value());
+    ASSERT_NE(doc.object, nullptr);
+    EXPECT_EQ(json_object_has_value(doc.object, "host"), 0) << "host must be absent when SetHostInfo was not called";
+}
```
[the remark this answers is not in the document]
robertwoj-microsoft: Fixed.

src/modules/complianceengine/src/assessor/assessor-result.schema.json
```diff
@@ -0,0 +1,112 @@
...
+        },
+        "ruleName": {
+          "type": "string",
+          "description": "PascalCase engine rule name."
+        },
+        "status": {
+          "type": "string",
+          "enum": ["Compliant", "NonCompliant", "NotApplicable"],
+          "description": "Per-rule status. NotApplicable rules are neutral and do not affect the overall status."
+        },
```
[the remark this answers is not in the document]
robertwoj-microsoft: Fixed.

src/modules/complianceengine/src/assessor/assessor-result.schema.json
```diff
@@ -0,0 +1,112 @@
...
+    },
+    "durationMs": {
+      "type": "number",
+      "description": "Wall-clock duration of the run in milliseconds."
+    },
+    "status": {
+      "type": "string",
+      "enum": ["Compliant", "NonCompliant", "NotApplicable"],
+      "description": "Overall status, aggregated across rules with the engine's allOf three-valued logic: NonCompliant if any rule is NonCompliant; otherwise NotApplicable if any applicable rule is NotApplicable; otherwise Compliant."
+    },
```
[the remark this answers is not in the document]
robertwoj-microsoft: Fixed.

src/modules/complianceengine/src/assessor/JUnitRenderer.cpp
```diff
@@ -0,0 +1,212 @@
...
+    {
+        const JSON_Object* rule = json_array_get_object(rules, i);
+        if (nullptr == rule)
+        {
+            return Error("Canonical result JSON 'rules' entry is not an object", EINVAL);
+        }
+        const string section = SafeString(json_object_get_string(rule, "section"));
+        const string ruleName = SafeString(json_object_get_string(rule, "ruleName"));
+        const string status = SafeString(json_object_get_string(rule, "status"));
+
```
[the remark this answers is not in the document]
robertwoj-microsoft: Fixed.

src/modules/complianceengine/src/assessor/TextRenderers.cpp
```diff
@@ -0,0 +1,153 @@
...
+    std::ostringstream out;
+    const size_t count = json_object_get_count(parameters);
+    for (size_t i = 0; i < count; ++i)
+    {
+        if (i != 0)
+        {
+            out << ", ";
+        }
+        out << SafeString(json_object_get_name(parameters, i)) << "=" << SafeString(json_value_get_string(json_object_get_value_at(parameters, i)));
+    }
```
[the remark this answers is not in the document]
robertwoj-microsoft: Fixed.

danielszot (approved):

src/modules/complianceengine/src/assessor/BenchmarkFormatter.hpp
```diff
@@ -6,12 +6,23 @@
...
 namespace ComplianceEngine
 {
 namespace BenchmarkFormatters
 {
+// Provenance of the host the scan actually ran on. The benchmark definitions
+// are architecture-agnostic, so the arch/distribution a result was produced on
+// is recorded in the result (not the definition) for multi-arch traceability.
+struct HostInfo
+{
+    std::string arch;                // uname(2) machine, e.g. "x86_64" / "aarch64"
```
kkanas: Nit we could use enum class && to_string()

kkanas (approved):

src/modules/complianceengine/src/assessor/JUnitRenderer.cpp
```diff
@@ -0,0 +1,218 @@
...
+            label = StringOrEmpty(json_object_get_string(node, "procedure"));
+        }
+        const string status = StringOrEmpty(json_object_get_string(node, "status"));
+        body << string(depth * 2, ' ') << "  - " << label;
+        if (!status.empty())
+        {
+            body << " [" << status << "]";
+        }
+        body << "\n";
+        AppendIndicators(json_object_get_array(node, "indicators"), depth + 1, body);
```
kkanas, line 96: Can this result unbouned recursion ? 
If so maybe limit the recursion depth ?

src/modules/complianceengine/src/assessor/BenchmarkFormatter.hpp
robertwoj-microsoft: Removed this struct entirely.

src/modules/complianceengine/src/assessor/JUnitRenderer.cpp
robertwoj-microsoft: Added limits. Note that the Engine class is not guarded against it, we'd need to add that as a follow-up.

src/modules/complianceengine/src/assessor/assessor-result.schema.json
```diff
@@ -0,0 +1,112 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://raw.githubusercontent.com/Azure/azure-osconfig/main/src/modules/complianceengine/src/assessor/assessor-result.schema.json",
```
pallakatos: Is it dynamically generated? Otherwise - we should not hardcode this ideally

src/modules/complianceengine/src/assessor/CliOptions.cpp
```diff
@@ -13,21 +13,37 @@ using std::string;
...
+    std::cout << "\t-f, --format\tPresentation format. Allowed values: {junit, nested-list, compact-list, debug}. Default: junit.\n";
+    std::cout << "\t    --suite-name\tName for the JUnit <testsuite>. Default: compliance.\n";
+    std::cout << "\tfilename\tRead the canonical result JSON from this file. Optional: if skipped or '-', reads standard input.\n";
 }
 
+// Long-only option identifiers (no short equivalent). Values start above the
+// ASCII range so they never collide with a short-option character.
+enum
+{
+    kSuiteNameOpt = 256
```
pallakatos, line 44: do we expect this value to change in future?

pallakatos (approved):
generally looks good - some hardcoded values if we could prevent that would be great - like the ID of the schema or the long option value

src/modules/complianceengine/src/assessor/CliOptions.cpp
robertwoj-microsoft: No, it's just a unique identifier to match on, can be anything above the ASCII limit, but still better than using raw value imho.

bketelsen (approved):
lgtm

src/modules/complianceengine/src/assessor/assessor-result.schema.json
robertwoj-microsoft: Dropped

robertschaedler3 (approved):

PR #1294, opened 2026-07-24
Fix invalid handling of comma-separated mach modes

robertwoj-microsoft:
## Description

Fix improper handling of comma-separated mach modes.

### Root cause

```
Match User *@*,<guid>
```
Contents of the pattern were passed directly to the `sshd` command:  `sshd -T -C user=*@*,<guid>`. The semantics is that the patterns are comma-separated and consittute an alternative of matches. The `sshd` command though, expects only a single parameter: either `*@*` or `<guid>`.

### The fix

Now the code splits the pattern by the comma character and tries `sshd -T -C user=<param>` command for each parsed parameter. Compliance is determined by combining all the patterns to not leave room for missed settings. For example, if we checked only the first pattern, we could miss settings for the users that have GUID as username in this case.

### Additional changes

As now for each parameter we check the same setting, I added the parameter context to the indicators to distinguish them. In the original IcM scenario we could have 2 indicators with the same message.

## Checklist


MariusNi (approved):
:shipit:

danielszot (approved):

